PR #2, opened 2017-05-09
Fix handling of *const *const ptr

weiznich:
Translating `*const *const my_type` ptr to c/c++ resulted in invalid code(`const const my_type**`)


rlhunt:
This seems reasonable to me. I will probably be doing more fixes to this code in the future because it doesn't handle many things correctly (like returning a function ptr from function), but this will help for now.

Thanks!

PR #27, opened 2017-07-19
Also generate headers for cdylib crates

jechas01:


rlhunt:
Thanks! I'll be pushing a new version to cargo sometime tonight.

PR #28, opened 2017-07-19
Support isize

jechas01:


rlhunt:
This won't work on every platform, but there isn't a cross platform solution so it's better to have something. Thanks!

emilio:
Wouldn't `intptr_t` work cross-platform?

rlhunt:
Yes, I believe it would. Don't know why I didn't think of that. I'll push a fix sometime soon.

PR #34, opened 2017-08-02
Allow to translate doc comments

weiznich:
This adds an option to translate rust doc comments into documentation
comments in the generated c/c++ header. 

For example the following code is translated into the following header:

```rust
/// This is the color enum
#[repr(u8)]
pub enum Color {
    /// Red color
    Red = 1,
    Blue = 2,
}

/// This is the color struct
#[repr(C)]
pub struct ColorStruct {
     /// A field storing the color of the struct
     my_color: Color,
}

/// A function taking the color struct
#[no_mangle]
extern "C" fn take_color_struct(a: ColorStruct) {}
```

```C++
// This is the color enum
enum class Color: uint8_t {
    // Red color
    Red = 1;
    Blue = 2;
};

// This is the color struct
struct ColorStruct {
     // A field storing the color of the struct
     Color my_color;
};

// A function taking the color struct
void take_color_struct(ColorStruct a);

```

### DIFF
--- a/src/bindgen/cdecl.rs
+++ b/src/bindgen/cdecl.rs
@@ -105,8 +105,7 @@ impl CDecl {
         }
     }
 
-    fn write<F: Write>(&self, out: &mut SourceWriter<F>, ident: Option<&str>, doc: &Documentation) {
-        doc.write(out);
+    fn write<F: Write>(&self, out: &mut SourceWriter<F>, ident: Option<&str>) {
 
         // Write the type-specifier and type-qualifier first
         if self.type_qualifers.len() != 0 {
@@ -191,7 +190,7 @@ impl CDecl {
                             // This is gross, but needed to convert &Option<String> to Option<&str>
                             let arg_ident = arg_ident.as_ref().map(|x| x.as_ref());
 
-                            arg_ty.write(out, arg_ident, &Documentation::none());
+                            arg_ty.write(out, arg_ident);
                         }
                         out.pop_tab();
                     } else {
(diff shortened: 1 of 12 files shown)


src/bindgen/cdecl.rs
```diff
@@ -213,14 +215,18 @@ impl CDecl {
...
+    if documentation {
+        &CDecl::from_func(f, layout_vertical).write(out, Some(&f.name), &f.doc_comment);
+    } else {
+        &CDecl::from_func(f, layout_vertical).write(out, Some(&f.name), &Documentation::none());
+    }
 }
 
-pub fn write_field<F: Write>(out: &mut SourceWriter<F>, t: &Type, ident: &str) {
-    &CDecl::from_type(t).write(out, Some(ident));
+pub fn write_field<F: Write>(out: &mut SourceWriter<F>, t: &Type, ident: &str, doc: &Documentation) {
```
eqrion: Can we instead have the callers print the documentation before calling `write_field`? I'd like to keep these modules separate.

src/bindgen/cdecl.rs
```diff
@@ -213,14 +215,18 @@ impl CDecl {
     }
 }
 
-pub fn write_func<F: Write>(out: &mut SourceWriter<F>, f: &Function, layout_vertical: bool) {
-    &CDecl::from_func(f, layout_vertical).write(out, Some(&f.name));
+pub fn write_func<F: Write>(out: &mut SourceWriter<F>, f: &Function, layout_vertical: bool, documentation: bool) {
```
eqrion: Ditto here.

eqrion:
👍 I like this change and the code looks good (the existing code isn't great, cbindgen could use some clean up). I'll merge it when the two issues are resolved.

weiznich:
@eqrion Thanks for the review. I've fixed the two issues.

eqrion:
Thanks!

PR #35, opened 2017-08-03
Add a option the prefix enum values by the enum name

weiznich:
This is mostly usefull for generated c headers because enums values
will live in the global namespace there

With this option enabled for the following code
```rust
#[repr(u8)]
pub enum Color {
    Red = 1,
    Blue = 2,
}
```
the following C Header is generated
```C
enum Color {
    Color_Red = 1,
    Color_Blue = 2,
};

typedef uint8_t Color;
```

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -90,8 +90,8 @@ impl Enum {
     }
 
     pub fn rename_fields(&mut self, config: &Config) {
-        if config.enumeration.prefix_with_enum_name ||
-           self.annotations.bool("prefix_with_enum_name").unwrap_or(false)
+        if config.enumeration.prefix_with_name ||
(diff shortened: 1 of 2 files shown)


src/bindgen/ir/enumeration.rs
```diff
@@ -90,6 +90,15 @@ impl Enum {
     }
 
     pub fn rename_fields(&mut self, config: &Config) {
+        if config.enumeration.prefix_with_enum_name ||
+           self.annotations.bool("prefix_with_enum_name").unwrap_or(false)
```
eqrion: Could you change the annotation to `prefix-with-name`? The other annotations use dashes and the enum part can be inferred, I think. The config value should be `prefix_with_name` also if you do this.

weiznich: Fixed

PR #36, opened 2017-08-03
Generate member functions

weiznich:
This pull request tries to make the generated c++ header more idiomatic.
In detail it tries to generate small member functions wrapping the exported functions for this type. 

For example for the following rust code
```rust
#[repr(C)]
struct MyStruct {
    some_value: i32,
}


#[no_mangle]
extern "C" fn do_something(a: *const MyStruct) {}
```

The following c++ header is generated
```C++
struct MyStruct;

void do_something(MyStruct* a);

struct MyStruct {
    uint32_t some_value;

    void do_something() {
        return ::do_something(this);
    }
};
```

In addition to generate more idiomatic code this allows to enforce some invariants of the underlying rust code. For example one could now safely assume that the pointer a for the function `do_something` is not null.

Future improvements for this feature may be to generate also constructor and destructor wrappers. 
I'm not sure about the syntax of this yet.

### DIFF
--- a/src/bindgen/ir/function.rs
+++ b/src/bindgen/ir/function.rs
@@ -14,6 +14,7 @@ use bindgen::library::*;
 use bindgen::rename::*;
 use bindgen::utilities::*;
 use bindgen::writer::*;
+use bindgen::mangle;
 
 #[derive(Debug, Clone)]
 pub struct Function {
@@ -50,7 +51,16 @@ impl Function {
         if let Some(&(_, ref ty)) = self.args.get(0) {
             match *ty {
                 Type::ConstPtr(ref t) | Type::Ptr(ref t) => {
-                    out.entry((**t).clone())
+                    let t = match **t {
+                        Type::Path(ref t, ref g) if g.is_empty() => {
+                            Type::Path(t.to_owned(), Vec::new())
+                        }
+                        Type::Path(ref p, ref g) => {
+                            Type::Path(mangle::mangle_path(p, g), Vec::new())
+                        }
+                        _ => return
+                    };
+                    out.entry(t)
                         .or_insert_with(Vec::new)
                         .push(self.clone())
                 }
@@ -97,6 +107,13 @@ impl Function {
         self.ret.mangle_paths(monomorphs);
         for &mut (_, ref mut ty) in &mut self.args {
             ty.mangle_paths(monomorphs);
+        }
+    }
+
+    pub fn as_member(self) -> Self{
+        Function {
+            args: self.args[1..].to_owned(),
+            ..self
         }
     }
 
--- a/src/bindgen/ir/function.rs
+++ b/src/bindgen/ir/function.rs
@@ -16,6 +16,12 @@ use bindgen::utilities::*;
 use bindgen::writer::*;
 use bindgen::mangle;
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum FunctionWriteMode {
+    Global,
+    MemberFunction,
+}
+
 #[derive(Debug, Clone)]
 pub struct Function {
     pub name: String,
@@ -52,9 +58,6 @@ impl Function {
             match *ty {
                 Type::ConstPtr(ref t) | Type::Ptr(ref t) => {
                     let t = match **t {
-                        Type::Path(ref t, ref g) if g.is_empty() => {
-                            Type::Path(t.to_owned(), Vec::new())
-                        }
                         Type::Path(ref p, ref g) => {
                             Type::Path(mangle::mangle_path(p, g), Vec::new())
                         }
@@ -110,15 +113,16 @@ impl Function {
         }
     }
 
-    pub fn as_member(self) -> Self{
-        Function {
-            args: self.args[1..].to_owned(),
-            ..self
-        }
-    }
-
-    pub fn write_formated<F: Write>(&self, config: &Config, out: &mut SourceWriter<F>, add_semicolon: bool) {
-        fn write_1<W: Write>(func: &Function, config: &Config, out: &mut SourceWriter<W>, add_semicolon: bool) {
+    pub fn write_formated<F: Write>(&self,
+                                    config: &Config,
+                                    out: &mut SourceWriter<F>,
+                                    mode: FunctionWriteMode)
+    {
+        fn write_1<W: Write>(func: &Function,
+                             config: &Config,
+                             out: &mut SourceWriter<W>,
+                             mode: FunctionWriteMode)
+        {
             let prefix = config.function.prefix(&func.annotations);
             let postfix = config.function.postfix(&func.annotations);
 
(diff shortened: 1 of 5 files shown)


src/bindgen/library.rs
```diff
@@ -633,8 +634,10 @@ impl Library {
         // Gather only the items that we need for this
         // `extern "c"` interface
         let mut deps = DependencyList::new();
+        let mut oop = MemberFunctions::new();
```
eqrion: What does `oop` stand for?

src/bindgen/ir/structure.rs
```diff
@@ -230,6 +255,33 @@ impl Source for Struct {
...
+        out.write(name);
+    }
+    out.write(");");
+}
+
+fn format_function_call_2<W: Write>(f: &Function, out: &mut SourceWriter<W>) {
+    out.write("return ::");
+    out.write(&f.name);
+    out.write("(");
+    let align_lenght = out.line_length_for_align();
```
eqrion: 'align_length'

src/bindgen/library.rs
```diff
@@ -873,6 +914,16 @@ impl GeneratedBindings {
         }
 
         if self.config.language == Language::Cxx {
+            out.new_line();
```
eqrion: Why two new lines?

src/bindgen/ir/function.rs
```diff
@@ -46,6 +46,19 @@ impl Function {
         })
     }
 
+    pub fn add_member_function(&self, out: &mut MemberFunctions) {
+        if let Some(&(_, ref ty)) = self.args.get(0) {
+            match *ty {
+                Type::ConstPtr(ref t) | Type::Ptr(ref t) => {
```
eqrion: We need to make these member functions const correct.

```rust
fn do_something(foo: *const Foo)
```
Should generate
```c++
struct Foo {
  void do_something() const {
    ::do_something(this)
  }
}
```

src/bindgen/ir/structure.rs
```diff
@@ -230,6 +255,33 @@ impl Source for Struct {
     }
 }
 
+fn format_function_call_1<W: Write>(f: &Function, out: &mut SourceWriter<W>) {
+    out.write("return ::");
```
eqrion: It'd be nice to omit the `return` if the return type is `void`, but from my brief testing it isn't a compile warning/error.

src/bindgen/library.rs
```diff
@@ -708,6 +711,42 @@ impl Library {
...
+                        let ty = Type::Path(s.name.clone(), Vec::new());
+                        if let Some(functions) = oop.remove(&ty) {
+                            let opaque = OpaqueItem {
+                                name:s.name.clone(),
+                                generic_params: s.generic_params.clone(),
+                                annotations: s.annotations.clone(),
+                                documentation: s.documentation.clone(),
+                            };
+                            result.items.push(PathValue::OpaqueItem(opaque));
+                            s.functions = functions.into_iter()
```
eqrion: Can you add a comment mentioning essentially that you're stripping out the `this` argument?

src/bindgen/library.rs
```diff
@@ -758,6 +797,7 @@ pub struct GeneratedBindings {
     monomorphs: Monomorphs,
     items: Vec<PathValue>,
     functions: Vec<Function>,
+    full_objects: Vec<Struct>,
```
eqrion: Can you rename this to `member_function_structs`? I don't mind the longer name as long as it is clear.

src/bindgen/ir/structure.rs
```diff
@@ -24,6 +24,7 @@ pub struct Struct {
     pub generic_params: Vec<String>,
     pub documentation: Documentation,
     pub functions: Vec<Function>,
+    pub destructor: Option<Function>,
```
eqrion: I'm worried this could be a potential footgun. In most cases if you implement any of the copy constructor, copy assignment, or destructor, you need to implement them all. [See the rule of three](http://en.cppreference.com/w/cpp/language/rule_of_three).

For example
```rust
struct WrapperVec {
  data: Vec<u8>,
}
fn destructor(this: *WrapperVec) {
  mem::drop(*this)
}
```
```c++
{
  WrapperVec a = ...
  WrapperVec b = a; // default copy constuctor does a shallow copy of the vec ptr's
} // destructor for a and b runs and the vec is dropped twice leading to a double free
```
This is a tricky problem. If we were to disable the copy constructor and copy assignment and only allow moves, I think we'd be in a better position. That would disadvantage rust structs that truly are POD. We could maybe generate `clone()` methods then?

I think whatever the solution is, it needs to be complete so we can't just generate destructors.

src/bindgen/library.rs
weiznich: Object orientated programming (It is a bad name, should better be something like member_functions)

src/bindgen/library.rs
weiznich: To separate those implementations visually from the functions before.

One new line is needed to start a new line after a function definition, the second is inserted to have an empty line between functions and classes.

src/bindgen/ir/structure.rs
weiznich: Seems like my c++ got a bit rusty :smile:  
What about the following idea:
If there is a destructor the exported struct is likely no POD and does not implement `Copy` on rust side. Therefore the copy constructor and copy assignment operator should be disabled. ( A future version may allow them by defining a clone function(like the destructor) that is internally called)
For me this seems like the solution that matches the underlying behavior of the rust code.

> I think whatever the solution is, it needs to be complete so we can't just generate destructor.

I've pushed the current implementation to see what others think about the idea. I've developed this while trying to generate a usable api for our internal project. Internally all my functions to free some allocated memory are implemented to reset the struct in a non allocating state, meaning free all memory and then set all pointers to null pointers. So the copy constructor should not harm this implementation. (But this is defined by my implementation, not by the generated header, so this is a not so great solution)

weiznich:
I've fixed all comments.

eqrion:
I'm sorry about the inconvenience, but this will probably need a rebase with the changes that have just landed.

weiznich:
> I'm sorry about the inconvenience, but this will probably need a rebase with the changes that have just landed.

As written in #42 I think it is not possible to implement this with the current dependency management in cbindgen. So we will need to improve this first.

Also I've found some additional problems related to the c++(gcc, clang) calling conventions. 
For example if I have the following rust code:

```rust
#[repr(c)]
struct Foo {
    a: i32
}

#[no_mangle]
pub extern "C" fn make_foo(a: i32) -> Foo {
    Foo {a }
}

#[no_mangle]
pub extern "C" fn some_member(foo: *mut Foo) {}
```
The straight forward C++ header will look like this:

```C++
struct Foo;

extern "C" void some_member(Foo* foo);

extern "C" struct Foo {
    int32_t a;
    
    void some_member() {
        ::some_member(this);
    }
};

extern "C" Foo make_foo(int32_t a);
```

When you try to use this code from some c++ programm, you will notice that something strange happens when you call make_foo. (The value passed will not be equal the value received on rust side).   Gcc happily accepts this code :disappointed_relieved: while clang emitts the following warning:
```
'make_foo' has C-linkage specified, but returns user-defined type 'Foo' which is incompatible with C 
```

To solve this we need to generate some more wrapping code:
```C++
struct Foo;

extern "C" void some_member(Foo* foo);

namespace intern {
    extern "C" struct Foo_intern {
        int32_t a;
    };
}
struct Foo: public intern::Foo_intern {
    Foo(intern::Foo_intern&& other) {
        this->a = other.a;
    }

    void some_member() {
        ::some_member(this);
    }
};

namespace intern {
    extern "C" Foo_intern make_foo(int32_t a);
}

inline Foo make_foo(int32_t a) {
    return intern::make_foo(a);
}
```

See [this commit](https://github.com/weiznich/cbindgen/commit/dd39e60b2a48a7473253b6ec8414b08842846b6b) for an implementation.

The big question now is how to move forward on this?
Do you want such a feature in cbindgen?
If yes how do we want to merge back all those changes? (If not, I'm also fine continue to use my own fork)

eqrion:
> The big question now is how to move forward on this?
> Do you want such a feature in cbindgen?
> If yes how do we want to merge back all those changes? (If not, I'm also fine continue to use my own fork)

So in the future I would like a feature like this. My goal with cbindgen is to make using a rust library from C/C++ safer and more ergonomic (in that order). Right now I don't have a whole lot of time for the library so I've been focusing on smaller things. This feature seems like it'll require a lot of careful thought and will be a significant investment for the codebase so I'm hesitant to merge anything right now.

I appreciate the work you've been doing, and would like to see what comes of it.

eqrion:
As there isn't a clear answer to these C++ and Rust interop/safety questions, I'm going to close this for now.

weiznich:
> As there isn't a clear answer to these C++ and Rust interop/safety questions, I'm going to close this for now.

As far as I remember all raised questions seems to be solved in my [fork](https://github.com/weiznich/cbindgen/commits/tmp/merged_all). The "biggest" issue was that thing with the calling conventions, that can be solved by generating some wrapper code as shown in [this](https://github.com/eqrion/cbindgen/pull/36#issuecomment-322764277) comment.

PR #37, opened 2017-08-04
Various bugfixes for specialisation of generic items

weiznich:


weiznich:
> 2. Vec<Vec<u8>> => Vec_u8_u8

There is already a test case with the type `Foo<Bar<f32>>`. The tests assert that this is equal to `"Foo_Bar_f32___"`. So which one is the right solution?

eqrion:
> There is already a test case with the type Foo<Bar<f32>>. The tests assert that this is equal to "Foo_Bar_f32___". So which one is the right solution?

```
Foo<Bar<f32>> => Foo_Bar_f32___
```
This is what we do today and is okay.
```
Foo<Bar<f32>> => Foo_Bar_f32
```
This would be nice.

The important thing is that `Vec<u8> => Vec_u8`, because that is the common case.

weiznich:
Thanks for the explanation:+1: (It is fixed now)

eqrion:
Thanks!

PR #39, opened 2017-08-08
Fix order of structs in case of using multiple monomorphs of the same type

weiznich:
Fixed Example:

```rust
struct List<T> {
     members: *mut T,
     count: usize
}

struct A;

struct B;

extern "C" fn foo(a: List<A>) {}
extern "C" fn bar(b: List<B>) {}
```

Before this commit the following code was generated:

```C++
struct A {};

struct List_A {
    A* members;
    size_t count;
};

struct List_B {
    B* members;
    size_t count;
}

struct B {};
```

After this commit the order of the generated structs are fixed:
```C++
struct A {};

struct B {};

struct List_A {
    A* members;
    size_t count;
};

struct List_B {
    B* members;
    size_t count;
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,6 +4,7 @@ version = "0.1.20"
 dependencies = [
  "clap 2.24.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


weiznich:
This won't work in more complex cases :disappointed: 
I will think about a better solution and open a new pull request with a fixed version (It seems like we must also track the monomorphed dependencies on it's own…)

PR #40, opened 2017-08-08
Generated specilizations should also be wrapped into the requested na…

weiznich:
…mespaces


weiznich:
Master already contains this fix, so I close this.

Issue #12, opened 2017-05-15
Maintain a more stable ordering of types


jrmuizel:
Anecdotally, it seems like if you change when I a type gets used it can end up changing order in the generated header which causes needless code churn. It might be nice to avoid this somehow.

staktrace:
This is actually quite annoying for when my WR update cron job, resulting in frequent merge conflicts. Is there something we can do to make the ordering stable? The SideOffsets2D_f32 keeps flip-flopping win the u32 variant.

eqrion:
There are definitely improvements we can make for the ordering. Unfortunately I haven't had time to improve cbindgen recently.

The issue with SideOffsets2D_f32 and SideOffsets2D_u32 seems like it could be from some uses of HashMap, which is non deterministic across changes.

I pushed a new release that uses BTreeMap instead, which might help your case. Let me know if it's still bad.

staktrace:
Initial results look good. Thanks! I'm ok with closing this issue for now and reopening it later if it's still a problem.

eqrion:
Okay cool. Let's keep this issue open for now, there are some other changes I'd like to make in the future.

eqrion:
I'm going to close this now as I think we do okay here. If there's something specific that comes up, lets open a new issue.

Issue #11, opened 2017-05-11
Give more control over which types are included


rlhunt:
Currently we will recursively export all of the referenced types from each of the exported `extern "C"` functions. This works well and keeps the bindings minimal. 

In the future, it could be useful to support a whitelist or a blacklist, along with a way to specify that a type should get a binding even if it's not referenced from an `extern "C"` function.

rlhunt:
This was improved enough for now.

PR #42, opened 2017-08-10
Fix order of structs

weiznich:
Improved version of #39 
Fixed Example:

```rust
struct List<T> {
     members: *mut T,
     count: usize
}

struct A;

struct B;

extern "C" fn foo(a: List<A>) {}
extern "C" fn bar(b: List<B>) {}
```

Before this commit the following code was generated:

```C++
struct A {};

struct List_A {
    A* members;
    size_t count;
};

struct List_B {
    B* members;
    size_t count;
}

struct B {};
```

After this commit the order of the generated structs are fixed:
```C++
struct A {};

struct B {};

struct List_A {
    A* members;
    size_t count;
};

struct List_B {
    B* members;
    size_t count;
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,6 +4,7 @@ version = "0.1.20"
 dependencies = [
  "clap 2.24.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "petgraph 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 0.9.15 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive 0.9.15 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 0.9.10 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -53,6 +54,11 @@ version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "fixedbitset"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "itoa"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -80,6 +86,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 name = "num-traits"
 version = "0.1.37"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "ordermap"
+version = "0.2.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "petgraph"
+version = "0.4.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "fixedbitset 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "ordermap 0.2.10 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
 name = "quote"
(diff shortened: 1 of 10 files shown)


eqrion:
A major refactoring of cbindgen just landed and I think this issue has been resolved. Can you test and let me know if this is still an issue?

weiznich:
> A major refactoring of cbindgen just landed and I think this issue has been resolved. Can you test and let me know if this is still an issue?

The Example from the pull request seems to be fixed now, but this does not resolve the general underlying issue. For example the following snippet will generate an invalid header:
```rust
#[repr(C)]
pub struct Foo {
    a: i32,
    b: *const Bar,
}

#[repr(C)]
pub struct Bar {
    a: *mut Foo,
}

#[no_mangle]
pub extern "C" fn foo(f: Foo) {}
```
generates the following header:
```C++

extern "C" {

struct Bar {
  Foo *a;
};

struct Foo {
  int32_t a;
  const Bar *b;
};

void foo(Foo f);

} // extern "C"
```

As soon as #36 lands the problem does get even more complex, because then we could not even output first the structs and then the generated extern functions, because to generate the member functions we will need to have access to each of the argument types and the inlined function. Each argument type could be a complete class on each own.

Let me try to outline the underlying problem: Each item could depend on a set on other items. At the time of collecting all function arguments one could not determine on which position an item should be printed. The current implementation tries to do the collection and the ordering in one step.

A possible solution for this problem seems to be to collect everything into a graph. Each node represents a specific item (e.g. a Struct or a function) and each edge represents the kind of the dependency between two items (Is it a pointer or a normal use?). For the example above the following graph is generated:
full

After that one could traverse the graph and find nodes without any dependency. This node could be added to the output and then be removed from the graph. This step is repeated till the graph is empty or there is no node without dependency left. (For example this is the case for the example above, because there is a circular dependency.). Now one will need to resolve the cycle. Because the graph contains information about which kind of dependency is used between two items it is now possible to search for pointeronly dependencies. To resolve those dependencies an opaque item could be injected and the corresponding edge can be removed. This will break the cycle and one could repeat the normal algorithm to determine the next output element. 
After breaking the cycle the graph for the example above looks like:
removed_cycle
Now one could remove each node step by step, because they do not depend on each other anymore:
removed_1_item.
This results in the following generated code:
```c++
struct Foo;

struct Bar {
  Foo *a;
};

struct Foo {
  int32_t a;
  const Bar *b;
};

void foo(Foo f);
```

If implemented all this on the [following branch](https://github.com/weiznich/cbindgen/tree/tmp/merged_all) in my repo to see how it interacts with the member function feature. (Unfortunately it is all smashed together in one branch and not rebased on master :cry: )

eqrion:
Yes, I've been considering creating an explicit graph for tracking dependencies for some time. Mostly for #12 and #11. I hadn't really considered keeping track of two types of edges (ptr and normal), but that would definitely allow us to generate forward declarations for structs when needed.

I think that would definitely be worthwhile, I think it could be implemented by rewriting the 'Dependencies' class to create an explicit graph, then creating an ordering of definitions and forward declarations using it.

I'm open to a PR for this, or I will implement it in a couple of weeks when I get time.

weiznich:
> I'm open to a PR for this, or I will implement it in a couple of weeks when I get time.

I will try to rework my branch to have a pull request for this, but this could take some days(Let's see when there is some time…)

I think implementing the dependency graph should be done before starting to implement member functions, because it is the much bigger change (in terms of complexity and code size). 

After that implementing member function should be easy to add. The "only" pain point there is to figure out the exact format of the generated header. (Although this should also be mostly done. At least for our use case it is now working (Calling a destructor, calling member functions, returning complex types…). )

eqrion:
I've filed #43 as an issue for this problem. Would you mind closoing this PR and opening up a clean one once you have an implementation?

djpiper28:
I am having errors from this and I am going to fix it and call it exam revision. (if I don't fail at fixing it lmao)

PR #44, opened 2017-08-17
Use a graph the represend the dependencies between exported items

### LINKED ISSUE #43: Generate forward declarations for structs with cyclical pointer references
Example

```rust
struct Foo {
    data: *const Bar,
}
struct Bar {
    data: *const Foo,
}
```

Possible output

```c
struct Bar;
struct Foo {
    const Bar* data;
}
struct Bar {
    const Foo* data;
}
```

This would require the dependency ordering algorithm to understand when a reference needs a declaration and when it needs a definition. If in the previous example, the data members were not pointers then it wouldn't work.

weiznich:
See #42 for details.
Fixes #43


weiznich:
@eqrion I've rebased this pull request again. I would be quite happy if we could work on merging this soon, because it changes a relative central part of cbindgen. Therefore each rebase is quite challenging.

eqrion:
@weiznich Okay, thank you for rebasing. Unfortunately, I'm away for the next week on vacation. I tried to review this before hand but was unable to.

I'll review this before making any future changes once I get back, so you won't have to rebase.

weiznich:
> This commit changes the output headers to allow functions, structs, enums, and template specializations to be intermixed. Whereas before we'd put all the functions at the bottom and all the enums at the top and all the template specializations after the extern "C" block.
>
> We should maintain the existing behavior, because the header files generated with this change are quite confusing. I also don't see any reason we can't maintain this behavior.

In the end it is required that they are intermixed because one of the items could depend on other items. (At least as soon as you start to implement the member functions feature). If this is really required it should be possible to retain a similar output order at least as long as there are no dependencies.

> Additionally, this commit changes the implementation of monomorphs and of specialization's and I don't see a need for it. I may have missed something.

As far as I understood the code and the output monomorphs are used for generating the template instantiations(`struct<T> Foo; -> `struct Foo_f32`), whereas specializations are used for generating typedefs for specialized items(`type Foo<T> = Bar<T>;`), right? 
If does not misunderstood the output the specializations thing did generate the struct again and not the a plain `typedef Foo_f32 = Bar_f32;` as it would be more correct. I did change this to use the typedef item to generate the according typedef.
I have monomorphs complete "removed" because they are not needed anymore. Monomorphing generic types is happening now as soon as a type is inserted in the graph. To generate the template specialization the "changed" specializations item. (To be honest it only took the name from the old one.)
This "new" item stores informations about what struct got specialized with which parameters.

I think it is quite hard to split this in a separate pull request because most parts of this are related to the item graph. Without the graph it would be quite hard to remove the monomorph item. Maybe the specializations item could be removed on it's one and be readded in this pull request.

eqrion:
> In the end it is required that they are intermixed because one of the items could depend on other items. (At least as soon as you start to implement the member functions feature). If this is really required it should be possible to retain a similar output order at least as long as there are no dependencies.

Yes, but member functions hasn't been implemented yet so I'd prefer not to have a bunch of code churn before it's needed. This change can be done later in the future if needed.

> I have monomorphs complete "removed" because they are not needed anymore. Monomorphing generic types is happening now as soon as a type is inserted in the graph. To generate the template specialization the "changed" specializations item. (To be honest it only took the name from the old one.)
This "new" item stores informations about what struct got specialized with which parameters.

Yes, that was what I was referring to, you changed the implementation here and I don't think it was needed for this PR. The dependency graph is generated in `Library` after monomorphs are created so it should be completely unaware of what monomorphs are.

weiznich:
> Yes, but member functions hasn't been implemented yet so I'd prefer not to have a bunch of code churn before it's needed. This change can be done later in the future if needed.

I did not implemented a sorting thing, because I did take this code straight from my member functions implementation. This will be a follow up pull request as soon as this one is merged.
If really required one could add such an feature in the mean time, but I think this is to complicated for an feature that would be used only a few weeks. (Feel free to disagree with that ;) )

> Yes, that was what I was referring to, you changed the implementation here and I don't think it was needed for this PR. The dependency graph is generated in Library after monomorphs are created so it should be completely unaware of what monomorphs are.

In theory your are right, this is not needed, but this simplifies many things. If we generate all monomorphs first and build then the dependency graph ofter those monomorphed items we will need to visit all items at least twice while building the graph. If we generate the monomorphs while building the dependency graph we only need to visit each item once (when we are adding it to the graph). This also simplifies the handling of the generated template specializations. 
As with the point above. In theory it should be possible to implement this without this change, but it will make the things much more complicated.


I'm currently abroad so I've not really any time to work on this for the next two weeks or so.

weiznich:
I've pushed an update that tries to reproduce the old order of items in the generated header files.
Notable differences are:
* There is no global `extern "C" ` block, instead each item is labeled individually. This is much easier to handle in case there are cyclic dependencies
* In case there are cyclic dependencies the order will be somewhat not like the old order, because there may items in between that are needed to break the cycle

weiznich:
Any news on this? (If not I'm in favour of closing this and using my own fork 😞)

eqrion:
Hey, I'm sorry but I don't think I'll be able to merge this soon. I've had to make other changes for new features to support webrender. It may be best for you to use your own fork for this, for now.

weiznich:
> Hey, I'm sorry but I don't think I'll be able to merge this soon. I've had to make other changes for new features to support webrender. It may be best for you to use your own fork for this, for now.

I think it should be possible to rebase this on the recent changes without any bigger issues. I'm willing to do this if there is a clear path to merge this. Ideally we would first talk about remaining issues in this implementation and finally rebase the then hopefully fixed implementation to master. If you could spare some time to work on this I would be happy to resolve the remaining issues. Otherwise I completely understand that the support for webrenderer required features has much bigger priority and will continue using my own fork.

eqrion:
My apologies for not coming back to this soon enough. As the code has changed a lot since this branch, and I think there are some issues remaining with the desired output for this feature, I'd like to close this and move any discussion to #47.

weiznich:
> My apologies for not coming back to this soon enough. As the code has changed a lot since this branch

No worries. Just ping me as soon as there is some spare time from your side to work again on this. 

>  I think there are some issues remaining with the desired output for this feature, I'd like to close this and move any discussion to #47.

As far as I remember the output was similar to the original one. The only changes are:
* `extern "C"` is not a "global" block anymore, but attached to each item separately (This allows future changes to reorder items more freely)
* If there is some kind of loop in the dependency graph those items will appear in the end of the output

PR #64, opened 2017-10-08
Also generate headers for rlib crates

Mrmaxmeier:
See also: 594cb21 and #27 
This covers all crate types except for `bin` and `proc-macro`.


eqrion:
Thanks!

PR #73, opened 2017-10-24
Support extern crate ... as ...; imports

jechas01:
If a crate is imported as `extern dependency as some_other_name;`, cbindgen fails to find it with a `can't find version for some_other_name` error message. The syn `ExternCrate` node comes with an optional alias, so use that when looking for a dependency crate if it exists.


eqrion:
Thanks!

PR #74, opened 2017-10-24
Use typedef for C enums as for structs

jechas01:
When declared as `#[repr(C)]`, enums are currently declared as
```C
enum Foo {
   ...
};
```
but show up in function arguments as
```C
void bar(Foo arg);
```
which isn't valid because it's missing the `enum` specifier.

Rather than insert it there, I followed the convention for struct declarations and made `#[repr(C)]` enums get declared as
```C
typedef enum {
    ...
} Foo;
```

Enums that are declared as `#[repr(<size>)]` and were previously typedef'd should remain unchanged.

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -162,7 +162,11 @@ impl Source for Enum {
         };
 
         if config.language == Language::C {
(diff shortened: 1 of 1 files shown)


eqrion:
Thanks!

PR #75, opened 2017-10-24
Look for -/_ replaced crates

jechas01:
Since cargo allows '-' in crate names, but rust source changes it to '_', cbindgen wasn't finding crates that were named like 'some-crate', since it was looking for 'some_crate' in the Cargo.lock. This does a find/replace on the crate name and searches for both versions of the name.


eqrion:
Awesome thanks!

PR #77, opened 2017-10-27
Support repr(usize) and repr(isize) enum.

JerryShih:
There some enums use repr(usize) and repr(isize) in rust std. Try to use intptr_t and uintptr_t[1] for these enums.

[1]
http://www.cplusplus.com/reference/cstdint/


JerryShih:
@eqrion r?

eqrion:
Thanks!

PR #80, opened 2017-11-09
Improve error message for generic types with underscores.

rillian:
mangle_name converts `Foo<Bar>` into `Foo_Bar`, and asserts
that the parent type name doesn't already contain underscores.

Report the offending type in the assert message so it's easier
to track down the problem, and add a unit test for this requirement.


eqrion:
Thanks!

Issue #72, opened 2017-10-19
How to specify `expand` for the current crate?


kirbysayshi:
I'm using cbindgen via a `build.rs` script within a root project. Let's call it `ffi_playground`. I have many macros within `ffi_playground` that help with making the ffi, including generating structs and functions.

My `cbindgen.toml` looks like this:

```
[parse]
parse_deps = false
```

With this configuration, my macros within `ffi_playground` (aka my root lib) are not expanded and my header is missing symbols.

If I attempt to force `ffi_playground` to be expanded:

```
[parse]
parse_deps = false
expand = ["ffi_playground"]
```

Now my build never finishes! I get very little output, and watching activity monitor implies an infinite loop somewhere (lots of rustc processes). I have to CTRL+C cargo.

Curiously, if I remove `build.rs`, and instead use the `cbindgen` CLI (still using cbindgen.toml), then with the first config I get the same result (incomplete header + warnings about missing symbols), and with the second config I get a _complete header_ but the process takes ~30 seconds (which is a loooong time!).

Am I doing something wrong? My assumption was that macros within the root crate would be expanded automatically, but that seems incorrect with the behavior I'm seeing.

kirbysayshi:
While I don't have my own code as an example, this publicly available code displays the same issue: https://github.com/getsentry/symbolic/tree/master/cabi.

To reproduce:
```
$ git clone git@github.com:getsentry/symbolic.git
$ cd symbolic/cabi
$ cargo build # this will be normal, there is no build.rs
$ time make
RUSTUP_TOOLCHAIN=nightly cbindgen -c cbindgen.toml . -o include/symbolic.h
WARN: Skip symbolic-cabi::LAST_ERROR - (Cannot have a non primitive const definition.)
WARN: Skip symbolic-cabi::LAST_PANIC - (Tuples are not supported types.)

real	1m17.777s
user	2m56.657s
sys	0m15.213s
```

If you add a `build.rs` file with the contents from the README from cbindgen, and add cbindgen as a `[build-dependency]`, a subsequent `cargo build` will never finish.

eqrion:
No I don't think you're doing anything wrong. Specifying `expand = ["ffi_playground"]` should do the trick.

> Curiously, if I remove build.rs, and instead use the cbindgen CLI (still using cbindgen.toml), then with the first config I get the same result (incomplete header + warnings about missing symbols), and with the second config I get a complete header but the process takes ~30 seconds (which is a loooong time!).

[...]

PR #81, opened 2017-11-09
Allow CARGO_TARGET_DIR override in cargo_expand

josefalcon:
Addresses part of #72 

Permits defining an override for the `CARGO_TARGET_DIR` variable when running cargo-expand. I used a separate variable `CARGO_EXPAND_TARGET_DIR` so as not to overload `CARGO_TARGET_DIR` in a build script, for example.


eqrion:
Looks good, thanks!

PR #84, opened 2017-11-15
Add write_fmt to SourceWriter

RReverser:
Allows to use write! macro on SourceWriter and so avoid allocation and construction of temporary Strings for dynamically constructed output.

.write method can still be used for static strings as a convenience.


eqrion:
This is a great change. Thanks!

PR #89, opened 2017-11-15
Don't prefix enum variants with name in C++

RReverser:
In C++, output already uses enum classes so enum variants are only accessible via enum name `E::A` and there is no point in manually prefixing them again to make it `E::E_A`.

C uses global scope for all enum variants, so transformation continues to work as it did.

I realise this change might be somewhat controversial, but it seems reasonable to me and would like to hear concerns if any.

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -123,8 +123,9 @@ impl Item for Enum {
     }
 
     fn rename_for_config(&mut self, config: &Config) {
-        if config.enumeration.prefix_with_name ||
-           self.annotations.bool("prefix-with-name").unwrap_or(false)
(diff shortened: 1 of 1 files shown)


eqrion:
Personally I don't know of any reason someone would want a prefixed enum variant in an enum class, but it is disabled by default so no one should have it prefixed unless they opt in.

So my only concern is if there is someone who desires this behavior, this would eliminate that choice for them. But again, I don't know who would want that.

RReverser:
> So my only concern is if there is someone who desires this behavior, this would eliminate that choice for them. But again, I don't know who would want that.

Yeah, same. I would go as far as claim that in C++ they are already prefixed with enum name by default using language facilities, so this option just allows to opt-in to the same behaviour for C.

PR #90, opened 2017-11-17
Simplify opaque struct output in C

RReverser:
There is no need for a separate forward struct declaration, just like in case with regular structures.


eqrion:
This is a good change, thanks!

PR #112, opened 2018-01-05
Include tag names on C structs

Fulkerson:
Much C code uses tag names instead of type names on structs. E.g:

    struct account x;

vs

    account x;

When implementing legacy C interfaces in Rust it is nice be able to generate header files that are compatible with this practice.

This pull request introduces three style, tag only, typedef only and both tag and typedef.

Tag only:

    struct balance {
      int dollars;
      int cents;
    };

    struct account {
        struct balance;
    };

Type only:

    typedef struct {
      int dollars;
      int cents;
    } balance;

    typedef struct {
      balance;
    } account;

Both only:

    typedef struct balance {
      int dollars;
      int cents;
    } balance;

    typedef struct account {
      balance;
    } account;

The default is Typedef only and no changes has been made to the generated headers (although it seems some other change has resulted in some new types). The test script has been modified to also build the tests with tag only and both. These files end up in tests/expectations/{tag, both}

### DIFF
--- a/src/bindgen/ir/item.rs
+++ b/src/bindgen/ir/item.rs
@@ -6,6 +6,7 @@ use std::collections::BTreeMap;
 use std::mem;
 
 use bindgen::config::Config;
+use bindgen::declarationtyperesolver::DeclarationTypeResolver;
 use bindgen::dependencies::Dependencies;
 use bindgen::ir::{AnnotationSet, Cfg, Constant, Enum, OpaqueItem, Static, Struct, Type, Typedef,
                   Union};
@@ -21,6 +22,8 @@ pub trait Item {
 
     fn container(&self) -> ItemContainer;
 
+    fn collect_declaration_types(&self, _resolver: &mut DeclarationTypeResolver) { unimplemented!() }
+    fn resolve_declaration_types(&mut self, _resolver: &DeclarationTypeResolver) { unimplemented!() }
     fn rename_for_config(&mut self, _config: &Config) {}
     fn add_dependencies(&self, _library: &Library, _out: &mut Dependencies) {}
     fn instantiate_monomorph(
--- a/src/bindgen/library.rs
+++ b/src/bindgen/library.rs
@@ -7,6 +7,7 @@ use std::mem;
 
 use bindgen::bindings::Bindings;
 use bindgen::config::{Config, Language};
+use bindgen::declarationtyperesolver::DeclarationTypeResolver;
 use bindgen::dependencies::Dependencies;
 use bindgen::error::Error;
 use bindgen::ir::{Constant, Enum, Function, Item, ItemContainer, ItemMap};
@@ -60,6 +61,8 @@ impl Library {
 
         if self.config.language == Language::C {
             self.instantiate_monomorphs();
+
+            self.resolve_declaration_types();
         }
 
         let mut dependencies = Dependencies::new();
@@ -263,6 +266,49 @@ impl Library {
 
         for item in &mut self.functions {
             item.rename_for_config(&self.config);
+        }
+    }
+
+    fn resolve_declaration_types(&mut self) {
+        if self.config.style.generate_typedef() {
+            return;
+        }
+
+        let mut resolver = DeclarationTypeResolver::new();
+
+        self.structs.for_all_items(|x| {
+            x.collect_declaration_types(&mut resolver);
+        });
+
+        self.opaque_items.for_all_items(|x| {
+            x.collect_declaration_types(&mut resolver);
+        });
+
+        self.enums.for_all_items(|x| {
+            x.collect_declaration_types(&mut resolver);
+        });
+
+        self.unions.for_all_items(|x| {
+            x.collect_declaration_types(&mut resolver);
+        });
+
+        self.enums
+            .for_all_items_mut(|x| x.resolve_declaration_types(&resolver));
+
+        self.structs
+            .for_all_items_mut(|x| x.resolve_declaration_types(&resolver));
+
+        self.unions
+            .for_all_items_mut(|x| x.resolve_declaration_types(&resolver));
+
+        self.typedefs
+            .for_all_items_mut(|x| x.resolve_declaration_types(&resolver));
+
+        self.globals
+            .for_all_items_mut(|x| x.resolve_declaration_types(&resolver));
+
+        for item in &mut self.functions {
+            item.resolve_declaration_types(&resolver);
         }
     }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,7 +20,7 @@ use clap::{App, Arg, ArgMatches};
 mod logging;
 mod bindgen;
 
-use bindgen::{Bindings, Builder, Cargo, Config, Error, Language};
+use bindgen::{Bindings, Builder, Cargo, Config, Error, Language, Style};
 
 fn apply_config_overrides<'a>(config: &mut Config, matches: &ArgMatches<'a>) {
     // We allow specifying a language to override the config default. This is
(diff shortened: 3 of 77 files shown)


eqrion:
There has been some back and forth on this issue. Originally we output tag names (see #90), but that was changed to simplify the output. I definitely see the use case for both.

Would this make sense to add the config option `config.structure.generate_tag` and then output tag names like you do in this PR when it is enabled? But then also retain the ability to only emit typedefs if desired. I'm okay with generating tags being the default.

We could also start exclusively using tags when that config option is set, like you mentioned. It's not a simple change, but shouldn't be terrible.

Fulkerson:
I will look into making it fully configurable. I had missed the Opaque forward declacation. If one wants a struct tag name one probably also want forward declared opaque types.

I think creating an enum in `config.structure` which can be `[TypeOnly, TagAndType]` which could be extended with `TagOnly` in the future could work. What do you think?

> It's not a simple change, but shouldn't be terrible

I looked at it a bit, but I'm still quite new at rust and I have a hard time wrapping my head around the code. From what I understand GenericPath is going to be the type that fields and function arguments are. So in order to know whether to output Type, struct Type or perhaps even enum Enum sometime, you need to know what the path actually is. 

And it's about there that I gave up. It's not really that important to me anyway.

RReverser:
Maybe we should just append `_t` to types as per popular C conventions? (and in C++ this doesn't matter anyway)

Fulkerson:
You can't have different type name in rust and C. That will result in linking error. Also `_t` is reserved in C so if you see anyone outside of the c standard using it they are doing it wrong!

I have managed to modify cbindgen to include union, struct and enum and it is configurable. I will clean it up and make a new pull request.

RReverser:
It won't, type names are not part of linking information and you can define it with any name and layout on the other side as long as it's compatible.

Fulkerson:
You are right! I got confused for a bit there.

sajattack:
:+1: https://github.com/redox-os/relibc just started using cbindgen and I ran into a situation where I had a struct and a function with the same name. The typedef struct doesn't work in this case and I'd like to be able to have structs generated in the format you specified, with parameters prefixed with the struct keyword.

Fulkerson:
I've pushed my current quick and dirty attempt. It worked up until the concept of "tagged enums" were introduced. I don't quite know what that is yet and haven't gotten it to work.

The idea is quite simply to store the knowledge of what kind of c type something is in a global cache and go over all types one last time before use to set the proper type. The tricky part is there doesn't seem to be a one place where all "things" are stored. The closest is the library, but in the tagged enum case it seems the types never actually land there.

src/bindgen/library.rs
```diff
@@ -266,6 +268,55 @@ impl Library {
         }
     }
 
+    fn set_ctype(&mut self) {
+        if self.config.style.generate_typedef() {
+            return;
+        }
+
+        let mut struct_names = HashSet::new();
```
Fulkerson: These should probably be made into some kind of cache that can be sent  around with convenience methods.

Fulkerson:
I managed to make the tests pass again. I'll try to clean it up a bit more.

Fulkerson:
I think I'm finally finished! I am ready for review.

Fulkerson:
@sajattack Please try this pull request out and see if it works :)

sajattack:
@Fulkerson So does Tag Only insert the struct keyword in front when it's used?

Fulkerson:
@sajattack Yes! And enums and unions. The only place where it still uses typedefs are sized non repr C enums. They are still typedefed as uint8_t for example.

sajattack:
What's the syntax for the config option? I tried:
```
[struct]
style = "TagOnly"
```
but that wasn't it

Fulkerson:
# A rule to use to select style of declaration in C, tagname vs typedef
    style = "[Both|Type|Tag]"

So put `style = "Tag"` in the config next to `language = "C"`

sajattack:
Doesn't work for structs within structs.

src/bindgen/ctyperesolver.rs
```diff
@@ -0,0 +1,62 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+use std::collections::HashSet;
+
+pub struct CTypeResolver {
+        structs: HashSet<String>,
```
eqrion: Can you decrease the indent here to match the rest of the file?

src/bindgen/ctyperesolver.rs
```diff
@@ -0,0 +1,62 @@
...
+            Some(CType::Struct)
+        } else if self.enums.contains(name) {
+            Some(CType::Enum)
+        } else if self.unions.contains(name) {
+            Some(CType::Union)
+        } else {
+            None
+        }
+    }
+}
```
eqrion: And add a newline at the end of the file?

src/bindgen/ctyperesolver.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use std::collections::HashSet;
+
+pub struct CTypeResolver {
+        structs: HashSet<String>,
+        enums: HashSet<String>,
+        unions: HashSet<String>,
+}
+
+#[derive(Clone, Copy, PartialEq, PartialOrd, Eq, Ord, Debug, Hash)]
```
eqrion: A little bikesheddy here, but would you be willing to rename CType to DeclarationType? I think it's an easier name and fits the style better. And CTypeResolver to DeclarationTypeResolver.

I know there is CDeclarator inside cdecl.rs, but that's an internal type for just that module.

src/bindgen/ir/item.rs
```diff
@@ -21,6 +22,8 @@ pub trait Item {
 
     fn container(&self) -> ItemContainer;
 
+    fn populate_ctyperesolver(&self, _resolver: &mut CTypeResolver) { unimplemented!() }
+    fn set_ctype(&mut self, _resolver: &CTypeResolver) { unimplemented!() }
```
eqrion: Along with the rename of ctype above, I think changing `populate_ctyperesolver` to `add_declaration_types` or `collect_declaration_types` and `set_ctype` to `resolve_declaration_types` would match the other functions here and might be clearer.

src/main.rs
```diff
@@ -111,6 +126,14 @@ fn main() {
                 .help("Specify the language to output bindings in")
                 .possible_values(&["c++", "C++", "c", "C"]),
         )
+        .arg(
+            Arg::with_name("style")
+                .short("s")
+                .long("style")
+                .value_name("STYLE")
+                .help("Specify the language to output bindings in")
```
eqrion: I think you forgot to update the help here. Maybe "Specify the declaration style to use for bindings"?

src/bindgen/ctyperesolver.rs
Fulkerson: Done!

src/bindgen/ctyperesolver.rs
Fulkerson: Done!

src/bindgen/ir/item.rs
Fulkerson: Done

src/main.rs
Fulkerson: Done!

src/bindgen/ctyperesolver.rs
Fulkerson: And done!

eqrion:
I'm sorry I forgot about this PR! These changes look good, thank you for being patient with me.

PR #113, opened 2018-01-09
Emit correct prototypes for functions without arguments in C mode

Mrmaxmeier:
`fun test() -> bool` was declared as `bool test();`.
Omitting the list of arguments in the prototype instructs the compiler to allow any type of arguments in the declaration of the `test` function. `bool test(void)` in C is equivalent to `bool test()` in C++ and will only allow declarations without arguments.
GCC warns about this with the `-Wstrict-prototype` flag set.

This change only affects `Language::C`. AFAIK the current C++ behavior is prefered.

### DIFF
--- a/src/bindgen/cdecl.rs
+++ b/src/bindgen/cdecl.rs
@@ -109,7 +109,7 @@ impl CDecl {
         }
     }
 
-    fn write<F: Write>(&self, out: &mut SourceWriter<F>, ident: Option<&str>) {
+    fn write<F: Write>(&self, out: &mut SourceWriter<F>, ident: Option<&str>, void_prototype: bool) {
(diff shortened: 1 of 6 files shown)


eqrion:
Thanks!

brendanzab:
Awesome, this will clean up some of our warnings! Any chance we can get this deployed sometime soon?

I'm just wondering, should typedefs also have `(void)` argument lists as well? Seems to not have been fixed to. This isn't an problem for us now, but I just thought I'd highlight it! https://github.com/eqrion/cbindgen/blob/59104b1efa6ddb3cdc6ed07a61be1291ce202503/tests/expectations/cdecl.c#L5

PR #115, opened 2018-01-17
Increase parse error message to include more context

staktrace:
I was getting a parse error on webrender and there was insufficient context to figure out what was going on.


eqrion:
Reasonable, thanks! I'll see about that compile error and if it's an easy fix I'll batch them for a new release.

PR #117, opened 2018-01-25
Add a derive_constructor option to generate struct constructors

staktrace:


src/bindgen/ir/structure.rs
```diff
@@ -232,6 +232,47 @@ impl Source for Struct {
...
+                && !self.fields.is_empty()
+            {
+                if !wrote_start_newline {
+                    wrote_start_newline = true;
+                    out.new_line();
+                }
+
+                out.new_line();
+
+                let arg_renamer = RenameRule::GeckoCase;
```
eqrion: Could we use `config.function.rename_args` here? Just so it's customizable.

staktrace:
Actually it might still be useful, if I add another option to add IPDL-friendly goop to the structs (which would consist of a private no-arg constructor and a friend declaration for the ParamTraits). I'll try that at some point.

src/bindgen/ir/structure.rs
staktrace: Updated patch to pull out a lambda that uses the config.function.rename_args if it exists.

eqrion:
Thanks!

PR #118, opened 2018-01-27
Non-generic support for tagged enums

RReverser:
I've come to realise that I can't find time to continue refactoring generics or to implement them by duplicating methods from other item types, so decided to at least rebase my branch with tagged enums as they are (without generics) to close #78.

Sorry for this taking so long and being somewhat unfinished.

PR #125, opened 2018-01-30
Fixes and improvements for tagged enums

### LINKED ISSUE #119: Implement repr(C, u8) enums
Same as repr(C), but sets the type of the tag enum (recommended).

### LINKED ISSUE #122: The bonus tag accessor for repr(u8) should be wrapped in a struct
Very strictly speaking it's only valid to access a prefix, and bare `Tag` isn't a prefix of `{ Tag; ... }`. But `{ Tag; }` is.

### LINKED ISSUE #126: C-like representation for tagged enums incorrectly emits union instead of struct
Tracking issue for https://github.com/eqrion/cbindgen/pull/118/files#r164882410

RReverser:
- Implement support for repr(C, Int) (fixes #119)
 - Fix wrapper kind for C-like tagged enum (fixes #126)
 - Wrap tag into anonymous struct in C++ (fixes #122)

cc @Gankro

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -353,7 +353,7 @@ impl Source for Enum {
             out.new_line();
 
             if config.language == Language::C {
-                out.write("typedef union");
+                write!(out, "typedef {}", if separate_tag { "struct" } else { "union" });
                 out.open_brace();
             }
 
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -10,7 +10,7 @@ use bindgen::config::{Config, Language};
 use bindgen::dependencies::Dependencies;
 use bindgen::library::Library;
 use bindgen::ir::{AnnotationSet, Cfg, CfgWrite, Documentation, GenericParams, GenericPath, Item,
-                  ItemContainer, Repr, Struct, Type};
+                  ItemContainer, Repr, ReprStyle, ReprType, Struct, Type};
 use bindgen::rename::{IdentifierType, RenameRule};
 use bindgen::utilities::find_first_some;
 use bindgen::writer::{Source, SourceWriter};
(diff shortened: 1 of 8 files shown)


RReverser:
Ugh, found a bug in previous commit - https://github.com/eqrion/cbindgen/pull/118/files#r164882410 - please don't merge yet.

Gankro:
Output in the tests looks right. Anonymous struct is slick! 👍

Quick questions since I don't know the code:

* What will this do if `repr(C, u8, u16)` is encountered? (this is unspecified by rust, so we should conservatively consider the repr opaque)
* What will this do if `repr(u8, new_repr_released_in_2020)` is encountered? (again should conservatively be opaque)
* What will this do if `#[repr(C)] #[repr(u8)]` is encountered? (should == `#[repr(C, u8)]`)

RReverser:
As for `repr(C, u8, u16)` and `repr(u8, new_repr_released_in_2020)` - Rust already emits warnings for that "conflicting representation hints" anyway, so I'd hope people just wouldn't have them in their code (FWIW cbindgen has similar assumptions about valid Rust in some other places too).

As for `#[repr(C)] #[repr(u8)]` - yes, that will work as expected.

Gankro:
The warnings tell users that they have a version of rustc that doesn't support that combination, but my concern is the opposite: they have a (future) version of rustc which *does*, but a copy of cbindgen which hasn't been taught about that yet and therefore miscompiles their code.

RReverser:
>  they have a (future) version of rustc which does

I assume this concern applies to your second example not `repr(C, u8, u16)`?

I think we can try to fallback to opaque on unknown reprs, yeah.

Gankro:
There's nothing that *fundamentally* prevents `(u8, u16)` from gaining a meaning. Also the warning that's emitted for it uses a weird mechanism that isn't picked up by `deny(warnings)`.

RReverser:
> There's nothing that fundamentally prevents (u8, u16) from gaining a meaning.

Heh, I'd love to see how that would make sense in future. Anyway, I made a patch to forbid both cases and fall back to opaque.

Gankro:
Thanks! 😄

eqrion:
This looks good. Thanks for following through with this work!

RReverser:
No problem, I'm interested in it as well :) Could you please publish a minor release with these fixes?

eqrion:
[Done](https://github.com/eqrion/cbindgen/releases/tag/v0.4.3)

PR #130, opened 2018-02-08
Add more libc primitive types

### LINKED ISSUE #129: Handling of Constants
Is there a clean way to handle constants (e.g. log levels)?

I could create an enum:

```rust
#[no_mangle]
#[repr(u8)]
pub enum my_log_level_t {
    LEVEL_TRACE = 0,
    LEVEL_DEBUG = 1,
    LEVEL_INFO = 2,
    LEVEL_WARN = 3,
    LEVEL_ERROR = 4,
    LEVEL_OFF = 5,
}

#[no_mangle]
pub extern "C" fn my_log_init(level: my_log_level_t) -> bool {
    ...
}
```

However, this way C code could still pass in for example the value `6`, which results in undefined behavior and a crash. So I'd prefer to work with plain u8 constants.

But this does not end up in the generated header file:

```rust
pub const LEVEL_TRACE: uint8_t = 0;
```

Would it be an option to convert `pub const` types to `#define`s?

dbrgn:
I only added int / uint types, but not types like `wchar_t` or `in_port_t`.

Fixes #129


eqrion:
This looks good. Thanks!

PR #135, opened 2018-02-21
Free cbindgen from rustc-internal dynamic lib dependencies

staktrace:
This updates cbindgen to use standalone-syn which in turn uses standalone-quote and standalone-proc-macro2, all of which use rustc-ap-proc_macro, which is a fork of the libproc_macro in rustc. This is a workaround for rust-lang/rust#47931 which allows the cbindgen binary to run without requiring all users to hack the LD_LIBRARY_PATH to pick up rustc internal dynamic libraries.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,7 @@
 language: rust
 cache: cargo
 rust:
-  - stable
-  - beta
   - nightly
-matrix:
-  allow_failures:
-  - rust: nightly
 addons:
     apt:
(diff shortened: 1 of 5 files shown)


staktrace:
Oh I should probably update the travis.yml file as well, it will likely fail builds on non-nightly

eqrion:
Awesome, thanks Kats! It works locally for me, I'll merge and try and publish this.

PR #140, opened 2018-03-03
Fix to_repr_c of PrimitiveType::USize to match PrimitiveType::ISize

jackpot51:


RReverser:
Can you provide motivation for this change?

jackpot51:
Currently to_repr_c converts  `PrimitiveType::USize => "size_t"`, `PrimitiveType::ISize => "intptr_t"`, and `PrimitiveType::SizeT => "size_t"`

This PR makes it so that `PrimitiveType::USize => "uintptr_t"`, which removes the assumption that `usize` and `size_t` are the same type. There is no requirement for `size_t` to be the address width of the architecture, whereas `uintptr_t` is exactly that.

jackpot51:
Thanks for merging @eqrion !

PR #142, opened 2018-03-05
Add an option to specify the lockfile

### LINKED ISSUE #141: support workspaces
The rust code in Mozilla-central has moved into a workspace, so the cbindgen step for regenerating the webrender bindings no longer works :(

staktrace:
Fixes #141.

For projects in workspaces, the Cargo.lock file may not be in the same
folder as Cargo.toml. This patch allows the user to explicitly pass the
path to the Cargo.lock file so that this scenario can be handled.

r? @eqrion


eqrion:
Thanks!

PR #150, opened 2018-03-11
Allow generating structs without a typedef

dlrobertson:
Add a configuration option that allows users to specify structs that
should not be decaired with a typedef. This is particularly useful when
a function that is accompanied by a struct of the same name.

I hit this issue today with `sigaction()` and `struct sigaction`. This was
my first thought/attempt at a solution. I'd be more than happy to attempt
to fix this with a different approach.

### DIFF
--- a/src/bindgen/cdecl.rs
+++ b/src/bindgen/cdecl.rs
@@ -4,6 +4,7 @@
 
 use std::io::Write;
 
+use bindgen::config::{Config};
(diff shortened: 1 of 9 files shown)


sajattack:
Possible duplicate of #112

dlrobertson:
Ah! Should have checked the opened PRs. I like #112 way better

PR #152, opened 2018-03-13
Fix handling of dependencies that are renamed

DataAnalysisCosby:
Our use case involves running cbindgen on a project that has internal dependencies which are renamed to be distinct from the spelling of their directories. This commit fixes that problem.

### DIFF
--- a/src/bindgen/parser.rs
+++ b/src/bindgen/parser.rs
@@ -104,7 +104,7 @@ struct Parser {
 }
 
 impl Parser {
-    fn should_parse_dependency(&self, pkg_name: &String) -> bool {
+    fn should_parse_dependency(&self, pkg_name: &String, orig_name: &String) -> bool {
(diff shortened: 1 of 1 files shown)


eqrion:
Thank you for the pull request!

After reading the code here and looking at the history, the current code for renaming extern crates is completely broken and it's my fault after the syn update. This code used to be needed before the syn update, and in the transition I flipped it backwards here, we shouldn't care at all what the renamed crate's new name is.

While debugging this I got the fix working and added a test case for it. Don't mean to steal this from you, but I'd feel worse making you do it again. The fix should be in cf3529fa84326d34917eb4b40620cd5e8101a0f8.

Again thanks for finding this!

DataAnalysisCosby:
No problem, thanks for fixing this so quickly! I figured you'd have a much better idea on how to fix this properly than I did.

PR #154, opened 2018-03-14
Improve module path parsing

dlrobertson:
If a module has a path attribute use this as the module path to parse.


eqrion:
This is great, and thank you for the test!

dlrobertson:
Np. Thanks for maintaining this project!

PR #157, opened 2018-03-26
Expand in normal target dir by default

### LINKED ISSUE #156: Consider not changing CARGO_TARGET_DIR by default for expansion
In https://github.com/eqrion/cbindgen/issues/72#issuecomment-343047353 you said that you initially changed CARGO_TARGET_DIR because it was causing problems with a custom build system.

Would you consider changing the default to use the normal target dir? I'm aware that I can set CARGO_EXPAND_TARGET_DIR, but it took me a while to discover this option... I think it'd be better if the default were fast. I'd imagine that the people for which the normal target dir causes a problem are the minority.

There are some minor issues that need to be worked out in that case, but I can prepare a PR for those.

untitaker:
Fix #156

### DIFF
--- a/src/bindgen/cargo/cargo_expand.rs
+++ b/src/bindgen/cargo/cargo_expand.rs
@@ -8,9 +8,6 @@ use std::path::Path;
 use std::process::Command;
 use std::str::{Utf8Error, from_utf8};
 
-extern crate tempdir;
-use self::tempdir::TempDir;
-
 #[derive(Debug)]
 /// Possible errors that can occur during `rustc --pretty=expanded`.
 pub enum Error {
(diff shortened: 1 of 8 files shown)


src/bindgen/cargo/cargo_expand.rs
```diff
@@ -37,39 +34,31 @@ impl From<Utf8Error> for Error {
...
-        } else {
-            Ok(src)
-        }
-    };
+    let src = from_utf8(&output.stdout)?.to_owned();
+    let error = from_utf8(&output.stderr)?.to_owned();
 
-    if let Ok(ref path) = env::var("CARGO_EXPAND_TARGET_DIR") {
-        run(&Path::new(path))
+    if src.len() == 0 {
```
untitaker: This can be true if `cargo rustc` is run twice consecutively. In that case `stderr` will be just "Finished in 0.0 seconds" and the exit code will be 0.

I'll add my fix in a separate commit, but there are several discussion-worthy solutions.

untitaker: @eqrion where is the best place to add tests for this?

untitaker: Actually I can't reproduce this problem at all on a different machine...

eqrion:
@untitaker Thanks for doing this! Now that I think about it, I'd actually prefer to keep the code for tempdir, but just not use it by default.

Specifically:
1. Add an option to `Cargo` for `use_tempdir` in the struct and constructor
1. Pipe that option into `cargo_expand::expand` and use it
1. Set that to false in uses of `Cargo` in `Builder::generate`
1. Add a command line option in `main.rs` for `--temp-dir` or maybe `--clean`?
1. Use that option in `Cargo::new` in `main.rs`

Let me know if that makes sense or not.

untitaker:
@eqrion do you insist on implementing your suggestion? I will do it if it means that this gets merged

eqrion:
@untitaker Yes, I'd prefer that.

untitaker:
@eqrion let me know if this is what you had in mind. I'm not entirely sure if adding it to the TOML config is a good idea, or how the envvar CARGO_EXPAND_TARGET_DIR fits into the picture wrt consistency.  Perhaps it should be a CLI option, but I think we could also remove it since there is no known usecase left for it that isn't already covered by `--clean`.

eqrion:
This is perfect, thank you! As for CARGO_EXPAND_TARGET_DIR, I'm don't exactly remember who needed that, but it's good to keep it around if it's not much trouble.

PR #158, opened 2018-04-04
Switch back to mainline syn instead of using standalone-syn

staktrace:
The new syn/quote (with the proc-macro dependency extracted as an optional feature) is on crates.io now, so we can switch back to using that.


eqrion:
Thanks for following up with this and finishing the fight!

PR #178, opened 2018-07-25
Add more control over the generated items.

emilio:
I'm going to need this for the style system, otherwise we generate a bunch of constants and such that we don't need.

Also I fixed a stylistic nit where the warning message was printed redundantly, and avoided generating empty extern blocks.

### DIFF
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -139,6 +139,40 @@ impl FromStr for Style {
 
 deserialize_enum_str!(Style);
 
+/// Different item types that we can generate and filter.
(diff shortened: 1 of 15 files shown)


jwatt:
I guess there should be some documentation on the new options, probably in README.md?

src/bindgen/config.rs
```diff
@@ -139,6 +139,40 @@ impl FromStr for Style {
 
 deserialize_enum_str!(Style);
 
+/// A style of Style to use when generating structs and enums.
```
eqrion: nit: Fix this doc comment.

PR #179, opened 2018-07-26
remove serde_derive dependency

ignatenkobrain:
It was pinned to be used with standalone-syn. Since we have switched
back to syn, it is no longer needed.

Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


eqrion:
I'm going to close this as there's no good path forward. Once #203 is resolved we can remove the pin.

PR #181, opened 2018-07-26
Use Cargo.lock from workspace root

Elarnon:
cbindgen currently assumes that the `Cargo.lock` is in directly in the
crate directory as a sibling to `Cargo.toml`; however that is usually
not the case inside a workspace.

Instead, this PR extracts the workspace root from the output of `cargo
metadata` [1] (already used to get a list of packages) and uses the
`Cargo.lock` from there.

 1. This is available since Rust 1.24; see rust-lang/cargo#4940


eqrion:
This looks good, thanks!

PR #182, opened 2018-07-27
Add ability to specify features to use for macro expansion

Elarnon:
Currently, `cbindgen` uses the `--all-features` flag when expanding a
dependent crate. However, this may not be desirable in a number of
cases:

 - Some C APIs may be gated by a feature flag and would not be present
   in the final cdylib depending on the features provided (for instance
   one could want to have the ability to build a "debug" version of the
   library which provides extra unstable hooks). In such cases, a
   programmatic `cbindgen` call in a build script would want to use only
   the features that will get used in the current build.

 - Some features may bring in large dependencies and/or potentially
   increase compilation time without affecting the FFI surface, and it
   would be faster and more efficient to disable them when running
   `cbindgen`.

 - Some features may require external libraries and/or hardware (e.g.
   dependencies on GPU libraries such as CUDA) that may not be available
   on the current machine without affecting the FFI surface.

To alleviate this problem, this PR adds an extended version of the
`parse.expand` configuration key, allowing control over the features
used when expanding in a way similar to the way cargo handles extended
dependencies (although note that there is a single version of each key,
since the features refer to the features of the current crate). So for
instance instead of writing `expand = ["euclid"]` one would write:

```
[parse.expand]
crates = ["euclid"]
```

which is equivalent to:

```
[parse.expand]
crates = ["euclid"]
all_features = false
default_features = true
features = ["feature1", "feature2"]
```

Note that `all_features` is set to `false` by default in order to match
cargo's behavior.

For backwards compatibility, the old syntax `expand = ["euclid"]` is
still supported and is equivalent to:

```
[parse.expand]
crates = ["euclid"]
all_features = true
default_features = true
features = null
```

In this case, `all_features` is set to `true` in order to match the
previous behavior of cbindgen.


README.md
```diff
@@ -98,13 +98,28 @@ parse_deps = true
...
 
+[parse.expand]
+# A list of crate names that should be run through `cargo expand` before
+# parsing to expand any macros
+crates = ["euclid"]
+# If enabled,  use the `--all-features` option when expanding. Ignored when
+# `expand_features` is set. Disabled by default, except when using the
+# `expand = ["euclid"]` shorthand for backwards-compatibility.
+all_features = false
+# When `enable_all_features` is disabled and this is also disabled, use the
```
eqrion: Do you mean `all_features` here?

README.md
```diff
@@ -98,13 +98,28 @@ parse_deps = true
...
 # Whether to use a new temporary target directory when running `rustc --pretty=expanded`.
 # This may be required for some build processes.
 clean = false
 
+[parse.expand]
+# A list of crate names that should be run through `cargo expand` before
+# parsing to expand any macros
+crates = ["euclid"]
+# If enabled,  use the `--all-features` option when expanding. Ignored when
+# `expand_features` is set. Disabled by default, except when using the
```
eqrion: Do you mean `features` here?

eqrion:
Also cc @emilio as I think he mentioned that `--all-features` was a problem for stylo.

Elarnon:
README updated without the typos.

Elarnon:
Rebased onto master and ran rustfmt.

eqrion:
Thanks!

Issue #186, opened 2018-08-07
Consts missing prefix when part of a type name, e.g. `[u8; BUF_LEN]`


regexident:
Input:

```rust
#![allow(non_camel_case_types)]
#![allow(non_upper_case_globals)]

pub const buffer_size: … = 16;

pub type buffer_t = [u8; buffer_size];

#[no_mangle]
pub unsafe extern "C" fn dummy(
    _buffer: &buffer_t,
) {}
```

Output:
```c
#include <cstdint>
#include <cstdlib>

static const uintptr_t prefix_buffer_size = 16;

using prefix_buffer_t = uint8_t[buffer_size];

extern "C" {

void dummy(const prefix_buffer_t *_buffer);

} // extern "C"
```

Notice how …

```c
using prefix_buffer_t = uint8_t[buffer_size];
```

… is missing the `prefix_` before `buffer_size`.

PR #190, opened 2018-08-10
Fixed missing prefix for named consts in `[T; <Const>]` types

regexident:
Fixes https://github.com/eqrion/cbindgen/issues/186.

cc @emilio


eqrion:
Thanks for fixing this!

Issue #175, opened 2018-07-23
Missing `#if defined(…) … #endif` guards in header of items from behind `#[cfg(…)]`


regexident:
Given a `Cargo.toml` like this …

<details open>

```toml
# Cargo.toml

[package]
name = "cbindgen_test"
# ...

[build-dependencies]
cbindgen = "~0.6"

[features]
foobar = []
```
</details>

… and a `cbindgen.toml` like this …

<details open>

```toml
# cbindgen.toml

[defines]
"foo" = "FOO"
"bar" = "BAR"
```
</details>

… and a `lib.rs` like this …

<details open>

```rust
// lib.rs

#[cfg(foo)]
pub const DOES_WORK: i32 = 1;

#[cfg(bar)]
pub const DOES_WORK: i32 = 2;

#[cfg(foo)]
#[no_mangle]
pub unsafe extern "C" fn does_work(dw: &DoesWork) {}

#[cfg(bar)]
#[no_mangle]
pub unsafe extern "C" fn does_work(dw: &DoesWork) {}

#[cfg(foo)]
#[repr(C)]
pub struct DoesWork {}

#[cfg(bar)]
#[repr(C)]
pub struct DoesWork {}

#[cfg(feature = "foobar")]
pub mod foo {
    #[cfg(foo)]
    pub const DOES_NOT_WORK: i32 = 1;

    #[cfg(bar)]
    pub const DOES_NOT_WORK: i32 = 2;

    #[cfg(foo)]
    #[no_mangle]
    pub unsafe extern "C" fn does_not_work(dnw: &DoesNotWork) {}

    #[cfg(bar)]
    #[no_mangle]
    pub unsafe extern "C" fn does_not_work(dnw: &DoesNotWork) {}

    #[cfg(foo)]
    #[repr(C)]
    pub struct DoesNotWork {}

    #[cfg(bar)]
    #[repr(C)]
    pub struct DoesNotWork {}
}
```
</details>

… one ends up with a header like this …

<details open>

```c
// header.h

#include <cstdint>
#include <cstdlib>

static const int32_t DOES_NOT_WORK = 1;

static const int32_t DOES_NOT_WORK = 2;

#if defined(FOO)
static const int32_t DOES_WORK = 1;
#endif

#if defined(BAR)
static const int32_t DOES_WORK = 2;
#endif

struct DoesNotWork {

};

struct DoesNotWork {

};

#if defined(FOO)
struct DoesWork {

};
#endif

[...]

PR #191, opened 2018-08-12
Fix "Missing `#if defined(…) … #endif` guards in header of items from behind `#[cfg(…)]`"

regexident:
Fixes https://github.com/eqrion/cbindgen/issues/175.

<details open>

```toml
# cbindgen.toml
[defines]
"foo" = "FOO"
"bar" = "BAR"
```

</details>

<details open>

```rust
// lib.rs

#[cfg(foo)]
pub const FOO: i32 = 1;

#[cfg(foo)]
#[no_mangle]
pub unsafe extern "C" fn foo(dw: &Foo) {}

#[cfg(foo)]
#[repr(C)]
pub struct Foo {}

#[cfg(feature = "foobar")]
pub mod foo {
    #[cfg(bar)]
    pub const BAR: i32 = 2;

    #[cfg(bar)]
    #[no_mangle]
    pub unsafe extern "C" fn bar(dnw: &Bar) {}

    #[cfg(bar)]
    #[repr(C)]
    pub struct Bar {}
}
```

</details>

Before this PR:

<details open>

```c
// lib.h

#include <stdint.h>
#include <stdlib.h>
#include <stdbool.h>

#define BAR 2

#if defined(FOO)
#define FOO 1
#endif

typedef struct {

} Bar;

#if defined(FOO)
typedef struct {

} Foo;
#endif

void bar(const Bar *dnw);

#if defined(FOO)
void foo(const Foo *dw);
#endif
```

</details>

After this PR:

<details open>

```c
// lib.h

#include <stdint.h>
#include <stdlib.h>
#include <stdbool.h>

#if defined(BAR)
#define BAR 2
#endif

#if defined(FOO)
#define FOO 1
#endif

#if defined(BAR)
typedef struct {

} Bar;
#endif

#if defined(FOO)
typedef struct {

} Foo;
#endif

#if defined(BAR)
void bar(const Bar *dnw);
#endif

#if defined(FOO)
void foo(const Foo *dw);
#endif
```

</details>

Notice how `#if defined(BAR) … #endif` now shows up as expected.

Another change this PR introduces is a warning for missing entries in `[defines]`:

In the case of this snippet:
```
WARN: Missing `[defines]` entry for `feature = "foobar"` in cbindgen config.
WARN: Missing `[defines]` entry for `feature = "foobar"` in cbindgen config.
WARN: Missing `[defines]` entry for `feature = "foobar"` in cbindgen config.
WARN: Missing `[defines]` entry for `feature = "foobar"` in cbindgen config.
```

( @emilio / @Gankro: This is the bug I mentioned to you at last week's Rust Berlin. 🙂)

### DIFF
--- a/src/bindgen/ir/cfg.rs
+++ b/src/bindgen/ir/cfg.rs
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use std::fmt;
 use std::io::Write;
 
 use syn;
@@ -9,6 +10,7 @@ use syn;
 use bindgen::config::Config;
 use bindgen::writer::SourceWriter;
 
+#[derive(PartialEq, Eq)]
 enum DefineKey<'a> {
     Boolean(&'a str),
     Named(&'a str, &'a str),
@@ -50,6 +52,36 @@ pub enum Cfg {
     Any(Vec<Cfg>),
     All(Vec<Cfg>),
     Not(Box<Cfg>),
+}
+
+impl fmt::Display for Cfg {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Cfg::Boolean(key) => write!(f, "{}", key),
+            Cfg::Named(key, value) => write!(f, "{} = {:?}", key, value),
+            Cfg::Any(cfgs) => {
+                write!(f, "any(")?;
+                for (index, cfg) in cfgs.iter().enumerate() {
+                    if index > 0 {
+                        write!(f, ", ")?;
+                    }
+                    write!(f, "{}", cfg)?;
+                }
+                write!(f, ")")
+            }
+            Cfg::All(cfgs) => {
+                write!(f, "all(")?;
+                for (index, cfg) in cfgs.iter().enumerate() {
+                    if index > 0 {
+                        write!(f, ", ")?;
+                    }
+                    write!(f, "{}", cfg)?;
+                }
+                write!(f, ")")
+            }
+            Cfg::Not(cfg) => write!(f, "not({})", cfg),
+        }
+    }
 }
 
 impl Cfg {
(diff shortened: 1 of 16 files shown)


src/bindgen/ir/cfg.rs
```diff
@@ -167,6 +198,64 @@ impl Cfg {
         }
     }
 
+    fn matched_defines(&self, config: &Config, unmatched: &mut Vec<Cfg>) -> Option<Cfg> {
```
emilio: This could get some comments :P

src/bindgen/ir/cfg.rs
```diff
@@ -286,21 +375,33 @@ impl CfgWrite for Option<Cfg> {
...
         if let &Some(ref cfg) = self {
-            // TODO
             if !cfg.has_defines(config) {
                 return;
             }
 
-            out.new_line();
-            out.write("#endif");
+            let mut unmatched = vec![];
+            if let Some(_) = cfg.matched_defines(config, &mut unmatched) {
```
emilio: nit: `is_some`?

src/bindgen/ir/cfg.rs
```diff
@@ -167,6 +198,64 @@ impl Cfg {
         }
     }
 
+    fn matched_defines(&self, config: &Config, unmatched: &mut Vec<Cfg>) -> Option<Cfg> {
+        match self {
```
emilio: nit: I'd match on `*self` and remove all the `&` in the left.

src/bindgen/ir/cfg.rs
```diff
@@ -52,6 +53,43 @@ pub enum Cfg {
     Not(Box<Cfg>),
 }
 
+impl fmt::Display for Cfg {
```
emilio: Is this used anywhere? Or is it just for debugging?

src/bindgen/ir/cfg.rs
```diff
@@ -205,6 +205,64 @@ impl Cfg {
         }
     }
 
+    fn matched_defines(&self, config: &Config, unmatched: &mut Vec<Cfg>) -> Option<Cfg> {
+        match self {
+            &Cfg::Boolean(ref cfg_name) => {
+                for (key, ..) in &config.defines {
+                    let key = DefineKey::load(key);
+                    match key {
```
emilio: nit: maybe use `if let`?

src/bindgen/ir/cfg.rs
```diff
@@ -205,6 +205,64 @@ impl Cfg {
...
+                        DefineKey::Named(..) => {}
+                    }
+                }
+                unmatched.push(self.clone());
+                None
+            }
+            &Cfg::Named(ref cfg_name, ref cfg_value) => {
+                for (key, ..) in &config.defines {
+                    let key = DefineKey::load(key);
+                    match key {
```
emilio: nit: maybe use `if let`?

src/bindgen/ir/cfg.rs
```diff
@@ -205,6 +205,64 @@ impl Cfg {
...
+                }).collect();
+                match cfgs.len() {
+                    0 => None,
+                    1 => cfgs.into_iter().next(),
+                    _ => Some(Cfg::All(cfgs)),
+                }
+            },
+            &Cfg::Not(ref cfg) => {
+                cfg.matched_defines(config, unmatched).map(|cfg| {
+                    Cfg::Not(Box::new(cfg.clone()))
```
emilio: nit: This clone seems redundant to me.

src/bindgen/ir/cfg.rs
```diff
@@ -382,21 +382,26 @@ impl CfgWrite for Option<Cfg> {
                 return;
             }
 
-            out.write("#if ");
-            cfg.write_condition(config, out);
-            out.new_line();
+            let mut unmatched = vec![];
```
emilio: I think you can use `&mut vec![]` below to avoid naming this if you want, or `matched_defined` could take an `Option`.

src/bindgen/ir/cfg.rs
```diff
@@ -388,6 +388,10 @@ impl CfgWrite for Option<Cfg> {
                 cfg.write_condition(config, out);
                 out.new_line();
             }
+
+            for cfg in unmatched {
+                warn!("Missing `[defines]` entry for `{}` in cbindgen config.", cfg);
```
emilio: I guess it's used here now :)

emilio: It's slightly weird that this warns about missing defines where we generate code, but not when we don't (in the `!has_defines()` case)... Should `write_condition` be adapted to just skip unknown conditions (and simplify out unknown ones ahead of time, or replace them by `1` maybe) instead of the `matched_defines` business?

looks like it's easy to overlook that you need to call `matched_defines` to get a useful config. Anyway not a big deal :)

src/bindgen/ir/cfg.rs
regexident: It's used in error/warning messages.

src/bindgen/ir/cfg.rs
```diff
@@ -286,21 +375,33 @@ impl CfgWrite for Option<Cfg> {
                 return;
             }
 
-            out.write("#if ");
-            cfg.write_condition(config, out);
-            out.new_line();
+            let mut unmatched = vec![];
+            if let Some(cfg) = cfg.matched_defines(config, &mut unmatched) {
```
eqrion: Do we still need the `if !cfg.has_defines(config)` line above, if we have this?

src/bindgen/ir/cfg.rs
eqrion: Yes, some comments here would help. The method seems sound though.

eqrion:
This overall looks good to me, just a few questions and nits before we can merge this.

src/bindgen/ir/cfg.rs
regexident: We don't. Fixed.

regexident:
I pushed another commit that introduces an additional type `Condition` (a `Cfg`, but with `cfg_name` and `cfg_value` replaced with the respective "define" from `Config`).

`CfgWrite` has become `ConditionWrite` now and `matched_defines` became `to_condition`.
With this it's now impossible to forget to "sieve" the `Cfg` before writing. We have a strong type system, we might as well make use of it! 😎

regexident:
(`Condition` would also allow for https://github.com/eqrion/cbindgen/issues/193.)

src/bindgen/ir/cfg.rs
```diff
@@ -166,139 +198,166 @@ impl Cfg {
...
-                        DefineKey::Boolean(key_name) => if cfg_name == key_name {
-                            define = define_value;
-                        },
-                        DefineKey::Named(..) => {}
-                    }
-                }
+#[derive(Debug, Clone)]
+pub enum Condition {
+    Boolean(String),
+    Named(String),
```
eqrion: Is there any difference between `Boolean` and `Named`?

regexident: No. I merged them into `Condition::Define`. :)

src/bindgen/config.rs
```diff
@@ -177,6 +177,30 @@ impl FromStr for ItemType {
 
 deserialize_enum_str!(ItemType);
 
+/// A language type to generate bindings for.
```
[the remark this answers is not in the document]
regexident: Removed. Along with the type. (It wasn't used anywhere, yet.)

regexident:
I'm not yet familiar enough with the whole of cbindgen's IR to comment on whether `Condition` > `Cfg` would be desirable, other than your argumentation making sense to me at first glance.

One issue I see with resolving `[defines]` from `Config` at parse-time would be that for emitting warnings one would have to walk the parse tree after parsing to check for missing `[defines]`. Unless one decides to make missing `[defines]` impossible by defaulting to auto-generated ones if missing (which might be a breaking change, though?).

> But, I don't think all that work is a necessary refactoring to land this PR. So feel free to fix the couple of nits and I can merge this and file a followup issue.

This should be fixed. :)

regexident:
Squashed.

eqrion:
Thanks!

PR #197, opened 2018-08-17
Point to a solution when we hit a parse error.

emilio:
When I hit this, I initially started bisecting my changes to see where the error
was.

It'd be much nicer if somebody had told me that `rustc -Z parse-only` existed.


eqrion:
Thanks!

Issue #7, opened 2017-05-09
Handle item paths better


rlhunt:
Each Rust `Item` has a path that includes the crate and mods needed to get to it. Currently `cbindgen` just ignores all segments of a path except the last one, which is the name of the item.

This works as long as there are no duplicates in different modules, but isn't correct and will probably cause someone an issue someday.

eqrion:
My comments earlier were wrong.

`cbindgen` used to refuse to parse paths that had multiple segments. This wasn't great so I fixed it in 015572877a9eec5b538c71ff73ddedae40e6210c.

AdamRzepka:
Any updates on this? I've got multiple `struct Config`s in different modules and want to export them all.

- What's the best workaround currently?
- Do you have any specific plans on implementing this?
- If it's not very hard, and you have some  I could try to implement the feature, equipped with some instructions.

eqrion:
Unfortunately, I'm not sure there is a workaround yet besides renaming the types. My first thought would have been to use the structs behind type aliases with different names, but that won't work if you want to expose them..

I don't currently have specific plans on implementing this. Unfortunately this is a tricky issue with a couple different options.

1. Implement a custom path resolution algorithm on top of syn
2. Use [rls-analysis](https://github.com/nrc/rls-analysis) to hook into the work going on there for goto definition
3. Write a rustc integration to get this data

I've been told by @Gankro that option 1 is not going to be fun as this is a very complicated and not well understood part of rustc and has evolved. I think this part of the rust reference is [relevant](https://doc.rust-lang.org/stable/reference/items/use-declarations.html).

Using rls-analysis could be viable, but needs investigation. My concern with using rls would be if it is using a lossy/mostly correct goto definition feature and not focused on strict correctness.

I have no idea of how feasible option 3 would be.

regexident:
### **Issue:** Import renames à la `use Foo as Bar;` are not resolved properly.

A minimalistic crate like this …

<details open>

```rust
// lib.rs

pub struct Foo { pub field: bool }

use Foo as Bar;

#[no_mangle]
pub unsafe extern "C" fn placebo(_bar: &Bar) {}
```

</details>

… generates a header like this …

<details open>

```c
// lib.h

#include <cstdint>
#include <cstdlib>

extern "C" {

void placebo(const Bar *_bar);

} // extern "C"
```

… emitting the following warning:

> WARN: Can't find `Bar`. This usually means that this type was incompatible or not found.

[...]

PR #201, opened 2018-08-22
Replaced `Item`s’ `String`-based names with proper `Path`s

regexident:
If we ever want to fix https://github.com/eqrion/cbindgen/issues/7 ("Handle fully qualified paths and `use` aliases"), we need to replace `Item`'s  reliance on its `self.name` field for type resolving with a proper `self.path` field holding a qualified path.

This PR replaces `pub name: String` on implementors of `Item` with `pub path: Path`, while (for now!) changing `Path` from `type Path = String` to `struct Path { name: String }`.

While this does not fix any existing type resolution bugs (let alone #7), it levels the field for any future fixes thereof and makes cbindgen more robust overall, in regards to type-safety.

**Important:** It also changes how cbindgen does renaming from changing its actual `self.name` field to having an auxiliary `self.export_name` field (that by default is set to be equal to the last path segment) which then is used in `fn write(…)`. This way we don't burn any bridges during the renaming phase and keep all the precious type information intact.

**Note:** Given that this PR touches pretty much every file in significant ways I'd rather merge all the existing (non-stale and acceptable) PRs, before merging this one. I'll then make sure to rebase this one accordingly.


regexident:
@eqrion I just did a rebase. Do you think we could merge this in, now that we're basically down to PR inbox-zero?

eqrion:
@regexident You may be interested in [this](http://dreamingofbits.com/post/future-directions-for-cbindgen-rust-ffi/).

Bu yes, I will take a look at this PR soon because it'd be good to get it merged either way.

regexident:
Looks like my rebase broke `cargo fmt --check`. Fixed.

Interesting read, @eqrion!

I very much agree with the points you're making. Looking forward to contributing to rust-ffi then!

I'd still like to merge this PR though, as it makes the whole renaming logic more explicit, even if the `String` to `Path` transition stays practically ineffective.

src/bindgen/ir/constant.rs
```diff
@@ -152,20 +155,44 @@ impl Constant {
...
+        path: Path,
+        ty: Type,
+        value: LiteralExpr,
+        cfg: Option<Cfg>,
+        annotations: AnnotationSet,
+        documentation: Documentation,
+    ) -> Self {
+        let export_name = path.name().to_owned();
+        Self {
+            path: path,
```
eqrion: I think it might be more idiomatic rust to change `path: path` to `path` in this constructor as they all have the same name. Ditto for all the other field initializers.

src/bindgen/ir/generics.rs
```diff
@@ -51,3 +54,87 @@ impl Source for GenericParams {
         }
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct Generic {
```
eqrion: Is there a reason you renamed this from `GenericPath` to just `Generic`? I think I prefer the old name and keeping it in `path.rs`, but it's not super important.

src/bindgen/ir/global.rs
```diff
@@ -31,14 +32,34 @@ impl Static {
...
+        path: Path,
+        ty: Type,
+        mutable: bool,
+        cfg: Option<Cfg>,
+        annotations: AnnotationSet,
+        documentation: Documentation,
+    ) -> Self {
+        let export_name = path.name().to_owned();
+        Self {
+            path: path,
```
eqrion: nit: drop the duplicated names for field initializers

src/bindgen/ir/opaque.rs
```diff
@@ -20,33 +20,56 @@ use bindgen::writer::{Source, SourceWriter};
...
     pub generic_params: GenericParams,
     pub cfg: Option<Cfg>,
     pub annotations: AnnotationSet,
     pub documentation: Documentation,
 }
 
 impl OpaqueItem {
-    pub fn new(
-        name: String,
+    pub fn load(
```
eqrion: Is there a reason this function is needed? Seeing as it just returns `Ok(OpaqueItem::new(..))`

src/bindgen/ir/ty.rs
```diff
@@ -229,24 +235,25 @@ impl Type {
...
                     Type::Primitive(prim)
                 } else {
-                    Type::Path(path)
+                    Type::Path(generic)
                 }
             }
             &syn::Type::Array(syn::TypeArray {
                 ref elem,
-                len: syn::Expr::Path(ref path),
+                len: syn::Expr::Path(syn::ExprPath { ref path, .. }),
```
eqrion: I haven't looked at the syn docs, but why is this change needed?

src/bindgen/ir/ty.rs
```diff
@@ -229,24 +235,25 @@ impl Type {
...
+                let generic = Generic::load(&path.path)?;
 
-                if path.name == "PhantomData" {
+                if generic.name() == "PhantomData" {
                     return Ok(None);
                 }
 
-                if let Some(prim) = PrimitiveType::maybe(&path.name) {
-                    if path.generics.len() > 0 {
+                // FIXME: this `generic_path.name()` seems wrong!
```
eqrion: Could you explain what's wrong about this check?

src/bindgen/ir/ty.rs
```diff
@@ -255,9 +262,9 @@ impl Type {
                     Some(converted) => converted,
                     None => return Err("Cannot have an array of zero sized types.".to_owned()),
                 };
-
-                let path = GenericPath::load(&path.path)?;
-                let len = ArrayLength::Name(path.name);
+                // FIXMe: An array's length cannot really be generic, or can it?
```
eqrion: Kind of, I think you can reference associated constants of generic paths.

```rust
struct Foo<T> {
  data: T,
}
trait TraitConst {
  const Value: i32;
}
impl<T> TraitConst for Foo<T> {
  const Value: i32 = 1;
}
fn main() {
  let x = [i32; Foo::<i32>::Value as usize];
}
```

I don't think we support the actual semantic structure of rust paths, which are arbitrary sequences of idents, generic specifiers, and lifetime specifiers. So in the current state of things, it makes sense for this to just use `Path`.

src/bindgen/ir/ty.rs
```diff
@@ -330,13 +337,17 @@ impl Type {
...
         let mut simplified = None;
-
-        if let &mut Type::Path(ref mut path) = self {
-            if path.name == "Option" && path.generics.len() == 1 && path.generics[0].is_repr_ptr() {
-                simplified = Some(path.generics.pop().unwrap());
+        if let &mut Type::Path(ref generic) = self {
+            let option_path = Path::new("Option");
+            if generic.path() == &option_path
+                && generic.generics().len() == 1
+                && generic.generics()[0].is_repr_ptr()
```
eqrion: This isn't your fault, but the logic of this function is wrong. It will convert `Option<* const i32>` to `const int32_t*` which isn't correct. The transformation is only valid if the pointer is non-nullable (like &i32). I'll file an issue.

src/bindgen/ir/ty.rs
```diff
@@ -408,14 +420,15 @@ impl Type {
...
-                for generic_value in &path.generics {
+            &Type::Path(ref generic) => {
+                for generic_value in generic.generics() {
                     generic_value.add_dependencies_ignoring_generics(generic_params, library, out);
                 }
-                if !generic_params.contains(&path.name) {
-                    if let Some(items) = library.get_items(&path.name) {
-                        if !out.items.contains(&path.name) {
-                            out.items.insert(path.name.clone());
+                let path = generic.path().clone();
```
eqrion: Is this copy necessary?

src/bindgen/ir/ty.rs
```diff
@@ -458,15 +471,15 @@ impl Type {
...
             }
-            &Type::Path(ref path) => {
-                if path.generics.len() == 0 || out.contains(&path) {
+            &Type::Path(ref generic) => {
+                if generic.generics().len() == 0 || out.contains(&generic) {
                     return;
                 }
-
-                if let Some(items) = library.get_items(&path.name) {
+                let path = generic.path().clone();
```
eqrion: Is this copy necessary?

src/bindgen/ir/opaque.rs
regexident: It's merely to make `OpaqueItem` be more in line with the rest of `ir::`.

And to hide the …

```rust
…
LiteralExpr::load(&item.expr)?,
Cfg::append(mod_cfg, Cfg::load(&item.attrs)),
AnnotationSet::load(&item.attrs)?,
Documentation::load(&item.attrs),
…
```

… shenanigans. Which is what all others are doing in `fn load(…)`, too.

src/bindgen/ir/ty.rs
regexident: Good to know. I'll keep it as is then.

regexident:
I'm having a rather hard time keeping up with all the recently merged PRs.
Could we focus on getting this rebased and merged, before merging any further PRs, or else just abandon it altogether?
With both branches diverging more and more with every PR rebasing is slowly becoming unmanageable. :(

PR #204, opened 2018-08-30
Add no_includes option

konstin:
Python's cffi can't handle include directives, which is why milksnake and pyo3-pack currently need to strip the includes from the header before they can use it. This pull request introduces a `no_includes` flag which simply disables the part that would normally write the includes.

I'm not sure if the flags should only disable the default includes (`stdint.h`/`stdlib.h`/`stdbool.h`) or all includes; For now i've chosen all.

For reference, [milksnake's include filtering](https://github.com/getsentry/milksnake/blob/ef0723e41df23d8f6357570c69c1e69cb31f9e9e/milksnake/ffi.py#L8-L16).

### DIFF
--- a/tests/rust/no_includes.rs
+++ b/tests/rust/no_includes.rs
@@ -1,0 +1,3 @@
+#[no_mangle]
+pub extern "C" fn root() {
+}
--- a/src/bindgen/bindings.rs
+++ b/src/bindgen/bindings.rs
@@ -65,9 +65,7 @@ impl Bindings {
         }
     }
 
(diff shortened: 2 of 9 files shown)


tests/rust/no_includes.rs
```diff
@@ -0,0 +1,4 @@
+#[no_mangle]
+pub extern "C" fn hello_world() {
```
eqrion: Could you rename this to `root` and drop the println!()? That would match our other tests.

konstin:
Fixed and squashed

eqrion:
Thanks!

PR #207, opened 2018-09-10
Bump to serde_derive 1.0.58

cuviper:
This will in turn build with `syn 0.13` and `quote 0.5`, which is still
distinct from cbindgen's own dependencies, so it won't get "infected" by
the proc-macro feature.


cuviper:
I've confirmed that the resulting binary is still independent of rustc shared libraries on `x86_64-unknown-linux-gnu` and `x86_64-pc-windows-msvc`, but someone should probably check macos too.

eqrion:
@cuviper I can confirm it works on OSX as well. Thanks!

PR #216, opened 2018-10-09
Skip lifetime type args

Mrmaxmeier:
The new test fails on master with `ERROR: Cannot use fn lifetime_arg::root ().`.

This matches the behavior of `cbindgen v3.3`. I'm not sure which version introduced the change.


eqrion:
This looks good, thanks!

PR #219, opened 2018-10-10
Add support for generic enums.

emilio:
Going to need this if I ever aim to generate TransformOperation bindings and
remove a bunch of slow and ugly Gecko code:

  https://searchfox.org/mozilla-central/rev/80ac71c1c54af788b32e851192dfd2de2ec18e18/servo/components/style/values/generics/transform.rs#189

With the caveat that I'll need to remove the options, but I can manage to do
that.

This also fixes a bunch of renaming bugs that I found while at it.

This patch has the gotcha that we need to remove the assertion of no-underscores
in mangled names... But I think it should be fine, and I'd rather not do a more
breaking change.

You can generate conflicting names in C using enum variants with the same name
as a struct regardless, for example, so I don't think this is terribly
important.

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -14,6 +14,8 @@ use bindgen::ir::{
     ItemContainer, Repr, ReprStyle, ReprType, Struct, ToCondition, Type,
 };
 use bindgen::library::Library;
+use bindgen::mangle;
+use bindgen::monomorph::Monomorphs;
 use bindgen::rename::{IdentifierType, RenameRule};
 use bindgen::utilities::find_first_some;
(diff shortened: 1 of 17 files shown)


emilio:
r? @eqrion 

(No rush on this one, not blocked whatsoever, yet... :P)

BorisChiou:
I also want this patch to drop servo/components/style/gecko_bindings/sugar/ns_timing_function.rs. :)

eqrion:
This looks good to me. Thanks for doing this!

eqrion:
Published in release [0.6.5](https://github.com/eqrion/cbindgen/releases/tag/v0.6.5).

emilio:
Thank you!

Issue #203, opened 2018-08-29
Find a way to update `serde_derive`


eqrion:
From #159 

> I thought we had an issue filed with exactly what is going on here, but we don't exactly. I'll summarize it here for reference and file an issue later.
>
> `cbindgen` depends on `syn` which depends on `proc-macro-2` because `syn` is geared to be used in proc-macros not random binaries for parsing rust code. `proc-macro-2` is the problematic library that causes this crash because it depends on internal compiler libs.
>
> `cbindgen` doesn't need any of the functionality from `syn` that uses `proc-macro-2` though, so we had a feature added to `syn` called `proc-macro` that we could use to disable this dependency. But disabling that feature is not enough to prevent `proc-macro-2` from being linked.
>
> Unfortunately `proc-macro-2` is used by `serde_derive` as a build time proc-macro crate, and cargo does feature flag resolution once for build dependencies and binary dependencies. So our feature flag gets overridden because `serde_derive` needs that flag enabled. Even though `serde_derive` only happens at compile time, and we want the feature disabled at run time.
>
> But not all `serde_derive`'s have this dependency. Which is why we pin to a version which doesn't yet have it.
>
> That's what rust-lang/cargo#2589 is about, and why it's still relevant. If that's fixed we could use the feature flag and update `serde_derive`.
>
> I don't think we could easily drop the `serde_derive` requirement either. It's used for parsing `cbindgen.toml` which isn't required if all you care about is `build.rs`. But it is used for parsing `Cargo.lock` which is required for most uses of the crate.
>
> This has been a major pain, and I'd love to find an answer to it.

cuviper:
I've updated the pinned version to 1.0.58 in #207, but the core problem is not resolved.

Would you be open to a PR that manually deserializes the necessary structs in cbindgen?  I know there are quite a few, but I'm somewhat inclined to just see how it goes...

sylvestre:
@eqrion @cuviper This issue is preventing us to upload cbindgen in Debian & Ubuntu.
We have serde_derive 1.0.70 in Debian and it would be tricky for us to backport 1.0.58

eqrion:
@sylvestre I wrote #222 to drop the serde_derive dependency using a hack. Hopefully this resolves the issue for you.

infinity0:
What's wrong with relaxing the dependency to `[dependencies.serde_derive] version = 1.0`? I couldn't figure this out from the huge explanation above.

eqrion:
> 
> 
[...]

PR #222, opened 2018-10-15
Remove serde_derive dependency, inlining expanded impls into source code

eqrion:
This commit removes the dependency on `serde_derive` so that
cbindgen doesn't depend on a stale package for OS's like debian.

A new file, README.serde_derive is added with instructions on
how to regenerate the impls with changes.


eqrion:
r? @emilio 

You obviously don't have to review the copy and pasted impl's, I'm more interested in a sanity check.

This will unblock the uploading of `cbindgen` to [Debian](https://github.com/eqrion/cbindgen/issues/203#issuecomment-429603319) and shift the burden of this issue from users to the maintainers.

There might be a CI failure, I didn't run this through rustfmt yet. Hopefully the autogenerated stuff is fine.

eqrion:
... and there are rustfmt errors. I'll sprinkle `#[cfg_attr(rustfmt, rustfmt_skip)]` over the commit and repush.

eqrion:
> Be aware that the generated code uses serde private APIs so is not guaranteed to compile or do the right thing against newer serde versions. That's fine within cbindgen as long as the impls are regenerated every time serde is bumped in Cargo.lock, but means that if you want other crates to use cbindgen as a library then you will need to pin a specific serde version in cbindgen's Cargo.toml. Otherwise things will break (frequently).

That's an understandable policy, I hadn't thought of that.

Thinking about it further, I'm not sure we can even rely on `Cargo.lock` for binaries as `cargo install cbindgen` doesn't use the lock file in my experience.

README.serde_derive
```diff
@@ -0,0 +1,25 @@
...
+5. Find the appropriate `impl`s and copy and paste them into the appropriate `mod`s.
+    1. Keep the `impl`s in source file order
+    2. Unindent the `impl`s to be aligned properly
+6. Update the comments with the version of `serde_derive` used to generate the `impl`s
+   1. You can find this in `Cargo.lock`
+7. Revert the changes to `Cargo.toml`
+8. Delete `expanded.rs`
+9. Test your changes
+
+This is very unfortunate, but it's better than keeping a stale dependency on `serde_derive`.
```
emilio: I think I disagree with this FWIW.

eqrion:
I suppose this change amounts to pinning `serde` instead of pinning `serde_derive`.

There is one benefit I could see to this, we can update the pinned version of `serde` when needed but currently can't with `serde_derive`. Still not a great solution.

src/bindgen/cargo/cargo_lock.rs
```diff
@@ -29,13 +29,17 @@ impl From<toml::de::Error> for Error {
...
+#[derive(Clone, Debug)]
 pub struct Lock {
     pub root: Option<Package>,
     pub package: Option<Vec<Package>>,
 }
 
-#[derive(Clone, Deserialize, Debug)]
+// Warning: Do not change this without regenerating serde_derive impls.
+// See issue #203 and README.serde_derive for more information.
+#[derive(Clone, Debug)]
```
emilio: These are all missing `#[cfg_feature]` bits right?

emilio:
Actually bindgen has the same issue (see https://github.com/rust-lang-nursery/rust-bindgen/pull/1409)... Sigh. I'd rather go fix https://github.com/rust-lang/cargo/issues/2589 instead fwiw.

dtolnay:
If the reason that this serde_derive dependency is a problem is that it enables proc-macro2's "proc-macro" feature and that ends up trying to dynamically link the compiler into cbindgen, then https://github.com/alexcrichton/proc-macro2/issues/139#issuecomment-429694118 may be a shorter path to fix that. We'd like for libproc_macro's build story to be no different from libcore or liballoc, meaning no more dynamic dependency on the compiler.

https://github.com/rust-lang/rust/pull/49219 is much of the work to get there.

eqrion:
I'm going to close this PR and put this on hold as it doesn't really sound like an attractive option. The work around libproc-macro sounds like a promising path.

PR #226, opened 2018-10-17
ir: Fix generic enums with untagged bodies.

### LINKED ISSUE #225: generic enum produce unnecessary template keywords
I tried to use generic enum for generics::transform::TimingFunction [1], and got the following code:
```
template<typename Integer, typename Number>
struct StyleGenericTimingFunction {
  enum class Tag : uint8_t {
    Keyword,
    CubicBezier,
    Steps,
  };

  template<typename Integer, typename Number>
  struct StyleKeyword_Body {
    StyleTimingKeyword _0;
  };

  template<typename Integer, typename Number>
  struct StyleCubicBezier_Body {
    Number x1;
    Number y1;
    Number x2;
    Number y2;
  };

  template<typename Integer, typename Number>
  struct StyleSteps_Body {
    Integer _0;
    StyleStepPosition _1;
  };

  Tag tag;
  union {
    StyleKeyword_Body keyword;
    StyleCubicBezier_Body cubic_bezier;
    StyleSteps_Body steps;
  };
  ...
}
```

It seems the `template<typename Integer, typename Number>`s for the inner struct cause the compilation errors, something like "error: declaration of 'Integer' shadows template parameter". I think we shouldn't generate the duplicated template parameters for the inner structs. Looks like I could pass the compilation after removing the duplicated inner template declarations.

@emilio 

[1] https://searchfox.org/mozilla-central/rev/3d989d097fa35afe19e814c6d5bc2f2bf10867e1/servo/components/style/values/generics/transform.rs#98
[2] Diff: https://gist.github.com/BorisChiou/c15171320e24efefb04efedb4de9bbf4

emilio:
Boris hit this because he used #[repr(C)] instead of #[repr(u8)].

We were incorrectly writing the generic arguments twice for those.

I missed this difference in #219.

Fixes #225.


emilio:
r? @eqrion 

Should be trivial, just overlooked the difference between tagged and untagged enum bodies.

emilio:
The generated union looks correct to me, fwiw... I don't understand what you're trying to imply there. The two struct layouts are correct.

  https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md#guide-level-explanation

contains the difference between them and the reasoning for the slightly more complicated representation when repr(C) isn't specified.

PR #233, opened 2018-10-22
Replaced Items’ String-based names with proper Paths

eqrion:
This is a rebased and updated version of #201.


regexident:
> @regexident Could you give this a glance? It's your work rebased and updated.

Looks good to me. Thanks for the rebase!

> As far as I can tell, this was a bug that we weren't doing this.

Yes, I stumbled upon it as well.

regexident:
I think the GitHub outage bricked the CI queue. You may have to manually re-trigger Travis-CI.

eqrion:
Fixed and merged. Thanks @regexident!

emilio:
This change should probably have been a breaking change. It made the `include` config option not work on the renamed path, but in the original one, which broke mozilla-central.

I'll update cbindgen on mozilla-central, so no action needed I guess, but wanted to point that out.

emilio:
Hmm, actually I think it may be better if we keep it the way it was... I'll see how hard it is to fix.

emilio:
Actually I think the new behavior is saner, let's keep that, add a test, and I'll take care of all versions building with old cbindgen versions.

Issue #232, opened 2018-10-22
Prefix for associated constants


kvark:
We are looking at the bitflags output of this struct:
```rust
bitflags! {
    #[repr(transparent)]
    pub struct TextureUsageFlags: u32 {
        const TRANSFER_SRC = 1;
        const TRANSFER_DST = 2;
        const SAMPLED = 4;
        const STORAGE = 8;
        const OUTPUT_ATTACHMENT = 16;
        const PRESENT = 32;
        const NONE = 0;
        const WRITE_ALL = 2 + 8 + 16;
    }
}
```
The result looks like this:
```cpp
#define WGPUOUTPUT_ATTACHMENT (TextureUsageFlags){ .bits = 16 }
#define WGPUPRESENT (TextureUsageFlags){ .bits = 32 }
```
There is "WGPU" prefix attached (which we specify in the options, but it's only means for types), and instead we'd want the type name to be the prefix, i.e.
```cpp
WGPUTextureUsageFlags_OUTPUT_ATTACHMENT
```
The current output is not very usable - it's not desired to have the type prefix affecting associated constants, and the way they are generated now there appears to be conflict between types (i.e. we have similar constants for `BufferUsageFlags` bitflags, and they overlap).

cc @grovesNL @eqrion @IGI-111

IGI-111:
Ah yes, that is one of the caveats, no explicit namespacing.

It should be relatively simple to add some (like a type prefix as you mention), but we should probably discuss the edge cases and document the specifics, because it's necessarily `cbindgen` specific behavior.

kvark:
Do you have any concerns (edge cases) in mind?

IGI-111:
Mostly collisions. Between namespaces and types, between classic constants and associated variable constants and so on.
`#define` collisions are especially hairy to debug so I'm inclined to be careful. But perhaps I'm just overthinking this.

I'm certainly not opposed to just doing the least surprising behavior by default, i.e.: just adding a type prefix. But we can probably avoid some pains with an optional prefix for typenames for instance. Something like

```toml
type_prefix = "__" # "" by default
```
That'd make `WGPU__TextureUsageFlags_OUTPUT_ATTACHMENT` rather than `WGPUTextureUsageFlags_OUTPUT_ATTACHMENT` in case you have some weirdly named types that overlap constants.

kvark:
[...]

PR #234, opened 2018-10-24
Add associated constant prefix

IGI-111:
Fix for #232
This adds a type prefix for associated constants to avoid namespace
collisions. It also adds error invocation in the rare but existing
cases where an collisions still happens in the constant namespace.


eqrion:
Thanks!

PR #236, opened 2018-10-30
Test the interaction between include and prefix configurations.

emilio:
This broke mozilla-central once, hopefully not twice :)


eqrion:
Looks good, thanks!

PR #237, opened 2018-10-30
Fix prefixing + formatting for enums 

lu-zero:
A configuration such as
``` toml
[enum]
prefix_with_name = true
```

Should map an enum such `Foo::Bar` to `FOO_BAR` or `PREFIX_FOO_BAR` if there is a prefix set.


lu-zero:
(moved to a topic branch)

PR #240, opened 2018-10-31
Propagate prefixes to struct literals

### LINKED ISSUE #238: Associated constants don't take the prefix into account properly
#234 is missing a test for preffixed bindings. This is what we get:
```cpp
typedef uint32_t WGPUTextureUsageFlags;
#define WGPUTextureUsageFlags_TRANSFER_DST (TextureUsageFlags){ .bits = 2 }
```
This doesn't compile, clearly, since there is no `TextureUsageFlags`.

cc @IGI-111

IGI-111:
Fix #238

This ensures that constants of a struct type have their type and the
type of their underlying value expressions renamed in the case of a
prefix.

### DIFF
--- a/src/bindgen/ir/constant.rs
+++ b/src/bindgen/ir/constant.rs
@@ -54,8 +54,13 @@ impl Literal {
             Literal::Struct {
                 name: _,
                 ref mut export_name,
-                fields: _,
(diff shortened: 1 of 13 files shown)


eqrion:
This looks great. Especially the thorough tests.

PR #243, opened 2018-11-01
Fix prefixing + formatting for enums

lu-zero:
A configuration such as
``` toml
[enum]
prefix_with_name = true
```

Should map an enum such `Foo::Bar` to `FOO_BAR` or `PREFIX_FOO_BAR` if there is a prefix set.

(I couldn't figure out how to move the former pr to this topic branch, sorry for the noise)

### DIFF
--- a/src/bindgen/rename.rs
+++ b/src/bindgen/rename.rs
@@ -65,10 +65,25 @@ impl RenameRule {
             RenameRule::PascalCase => text.to_owned(),
             RenameRule::CamelCase => text[..1].to_lowercase() + &text[1..],
             RenameRule::SnakeCase => {
+                // Do not add additional `_` if the string already contains `_` e.g. `__Field`
+                // Do not split consecutive capital letters
                 let mut result = String::new();
+                let mut add_separator = true;
+                let mut prev_uppercase = false;
                 for (i, c) in text.char_indices() {
-                    if c.is_uppercase() && i != 0 {
-                        result.push_str("_");
+                    if c == '_' {
+                        add_separator = false;
+                        prev_uppercase = false;
+                    }
+                    if c.is_uppercase() {
+                        if i != 0 && add_separator && !prev_uppercase {
+                            result.push_str("_");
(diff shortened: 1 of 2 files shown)


src/bindgen/rename.rs
```diff
@@ -79,9 +79,22 @@ impl RenameRule {
             RenameRule::ScreamingSnakeCase => {
                 // Same as SnakeCase code above, but uses to_uppercase
                 let mut result = String::new();
+                let mut add_separator = true;
```
eqrion: This is a bit confusing for me. Could you add a comment explaining what cases you're trying to handle. Additionally, do any other rename rules need to be updated to have compatible behaviors with consecutive upper case letters or underscores?

lu-zero: Basically assume you have something with an `_` in the name for some reason, one being having collated a prefix as part of the `export.prefix = "Prefix"` combined with `enum.prefix_with_name = true`.

You have an enum such as
``` c
#[repr(C)]
pub enum Unprefixed {
    Some,
    Other,
}
```

You'd enjoy producing
``` c
typedef enum {
    PREFIX_UNPREFIXED_SOME,
    ...
}
```
and not
``` c
PREFIX_UNPREFIXED__SOME
```

The screaming-snake-case has also a corner case situation when you want to not add a `_` between two consecutive capital letters.

lu-zero: SnakeCase would enjoy the same logic I guess.

eqrion: Interesting. So I think this is okay. It's a bit hard to reason about these functions. In the long run, we should use a dedicated crate for this. I think [`heck`](https://github.com/withoutboats/heck) should support everything we need. The most important thing will probably be testing compatibility.

But again, this is probably good for now. Could you add a comment in the code explaining what this extra logic is doing?

lu-zero: I reworked it a bit, is it fine like this?

PR #248, opened 2018-11-12
Remove member name from transparent associated constants

grovesNL:
This attempts to fix `repr(transparent)` for associated constants by unwrapping the first field on structs (see https://github.com/eqrion/cbindgen/issues/238#issuecomment-435923997 for more context).

I didn't see obvious ways to infer whether the struct the `impl` block corresponds with is `repr(transparent)`, and the order isn't guaranteed (`impl` prior to `struct`). So we have to parse the struct declaration first, and reference it later, which is why the `impl` blocks are loaded after the regular loop now.

I'm not sure whether the general approach is correct, so any feedback is greatly appreciated, especially ideas to simplify the branching or remove the `Vec` allocation.

### DIFF
--- a/src/bindgen/parser.rs
+++ b/src/bindgen/parser.rs
@@ -522,12 +522,10 @@ impl Parse {
                     self.load_syn_ty(crate_name, mod_cfg, item);
                 }
                 syn::Item::Impl(ref item_impl) => {
-                    let has_assoc_const = item_impl.items
-                        .iter()
-                        .any(|item| match item {
-                            syn::ImplItem::Const(_) => true,
-                            _ => false,
-                        });
+                    let has_assoc_const = item_impl.items.iter().any(|item| match item {
+                        syn::ImplItem::Const(_) => true,
+                        _ => false,
+                    });
                     if has_assoc_const {
                         impls_with_assoc_consts.push(item_impl);
                     }
@@ -537,12 +535,10 @@ impl Parse {
         }
 
         for item_impl in impls_with_assoc_consts {
-            let associated_constants = item_impl.items
-                .iter()
-                .filter_map(|item| match item {
-                    syn::ImplItem::Const(ref associated_constant) => Some(associated_constant),
-                    _ => None,
-                });
+            let associated_constants = item_impl.items.iter().filter_map(|item| match item {
+                syn::ImplItem::Const(ref associated_constant) => Some(associated_constant),
+                _ => None,
+            });
             self.load_syn_assoc_consts(
                 binding_crate_name,
                 crate_name,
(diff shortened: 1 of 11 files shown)


grovesNL:
cc @kvark @IGI-111 if you'd like to take a look

kvark:
The error handling seems a bit wonky (hard to catch the actual non-error code path by the eye), but the logic looks good to me.

IGI-111:
It definitely seems workable, but I'm concerned how big that Vec can grow.

Not sure we can avoid it since as you say the order isn't guaranteed.

grovesNL:
@eqrion @IGI-111 do you have any ideas for improvements on this?

kvark:
@eqrion ping

src/bindgen/parser.rs
```diff
@@ -636,36 +642,80 @@ impl Parse {
...
+            } else {
+                info!(
+                        "Skip impl block for {}::{} (impl blocks for associated constants are only defined for structs).",
+                        crate_name, impl_struct_path
+                    );
+                return;
+            }
+        } else {
+            error!(
+                    "Cannot find type for {}::{} (impl blocks require the struct declaration to be known).",
```
eqrion: Was it possible for users to have associated constants for `enum`s and not just `struct`s? It looks like the code didn't check for this constraint before, but I may be missing it.

If so, then this is kind of a breaking change. Not sure if that matters or not.

src/bindgen/parser.rs
```diff
@@ -524,21 +526,25 @@ impl Parse {
...
-                            self.load_syn_assoc_const(
-                                binding_crate_name,
-                                crate_name,
-                                mod_cfg,
-                                &item_impl.self_ty,
-                                item,
-                            );
-                        }
-                    }
+                    impls.push(item_impl);
```
eqrion: Can we minimize our memory usage here by filtering out `Impl`s that don't contain associated constants?

I'm not too worried about it though, as IIRC `rustc` needs to hold the full AST in memory, so if that's too much for us it should be too much for them as well.

tests/expectations/tag/const_conflict.c
```diff
@@ -2,4 +2,4 @@
 #include <stdlib.h>
 #include <stdbool.h>
 
-#define Foo_FOO 0
+#define Foo_FOO 42
```
eqrion: Is this change okay?

grovesNL: I left it for now unless it's preferable to early exit in this case. The current behavior is to emit an error (with `error!`) and continue.

src/bindgen/parser.rs
grovesNL: I think this should be handled now.

src/bindgen/parser.rs
grovesNL: I tried adding it to traits but it doesn't seem like there is much  in place for handling traits yet, (unless I've missed it). For example, I couldn't see how to know whether a trait has been implemented for a particular type, and without that I can't emit the associated constant for the concrete type.

grovesNL:
@eqrion I updated the PR. Please take a look, thanks!

eqrion:
Looks good, thanks!

PR #250, opened 2018-11-21
Escape reserved C/C++ keywords from struct fields and function arguments

### LINKED ISSUE #56: C/C++ reserved keywords should be escaped
If I have 
````
pub extern fn my_function(this: *mut MyStruct) -> bool
````
cbingen will generate
````
bool my_function(MyStruct* this);
````

But `this` is a reserved keyword in C++. It should be escaped. To `this_` would work.

eqrion:
Fixes #56. Feel free to suggest a better way to escape than postfixing a '_'.

r? @emilio 
cc @jrmuizel


eqrion:
We should probably start using the stable channel for rust-fmt, now that it's supported there.

PR #255, opened 2018-11-24
ir: Add support for negative enum discriminants.

emilio:
Pretty straight-forward, but it bit me today :)


eqrion:
Looks great, thanks.

PR #258, opened 2018-12-12
Properly propagate const into arrays.

jrmuizel:
i.e. 'type I = *const [i32; 16]' should map to 'typedef const int32_t (*I)[16]'
instead of 'typedef int32_t (*I)[16]'


jrmuizel:
@eqrion review ping

eqrion:
Thanks!

PR #266, opened 2019-01-03
Support references better

jrmuizel:
This adds better support for references in the type system
and uses them for enumeration helpers instead of string
concatenation.


jrmuizel:
I've updated the patch to fix the cmp.* helpers to return false.

eqrion:
Looks great. Could you fix the rustfmt issue? I really should switch travis over to using non-nightly rustfmt.

jrmuizel:
Done

eqrion:
Thanks!

PR #267, opened 2019-01-03
Manually copy arrays

jrmuizel:
C++ arrays can't be assigned. We detect this situation and manually copy
over the elements.


eqrion:
Looks good, thanks for the test!

PR #271, opened 2019-01-18
Unpin serde

### LINKED ISSUE #203: Find a way to update `serde_derive`

Mrmaxmeier:
This bumps the minimum required `rustc` version to 1.32.

Updating `serde_derive` breaks the builds for `rustc < 1.32`.  See #203 for details.
```console
$ cargo +stable -vV
cargo 1.31.0 (339d9f9c8 2018-11-16)
release: 1.31.0
commit-hash: 339d9f9c8f400010df3282ae5582bf3a0f739004
commit-date: 2018-11-16
$ cargo +stable build
[...]
$ ldd target/debug/cbindgen
	linux-vdso.so.1 (0x00007fff69b58000)
	libproc_macro-eb9b1f5dc4b73ebd.so => not found
	libsyntax-8b4b84ccec0d2e12.so => not found
	libsyntax_pos-31f39647344993a7.so => not found
	libstd-89cf9eb8d404bb7b.so => not found
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f0d4a1a8000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007f0d4a18e000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007f0d49fc8000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f0d4ac18000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007f0d49e43000)
```

After updating to today's stable release, this is no longer an issue. :tada:

```console
$ cargo +stable -vV
cargo 1.32.0 (8610973aa 2019-01-02)
release: 1.32.0
commit-hash: 8610973aaf48615ba7dc9a38a9a2795ba6f36a31
commit-date: 2019-01-02
$ cargo +stable build
[...]
$ ldd target/debug/cbindgen
	linux-vdso.so.1 (0x00007ffed69b0000)
	libdl.so.2 => /usr/lib/libdl.so.2 (0x00007f9412875000)
	librt.so.1 => /usr/lib/librt.so.1 (0x00007f941286b000)
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f941284a000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007f9412830000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007f941266c000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f94132f7000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007f94124e7000)
```

Closes #203.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,53 +3,58 @@ name = "ansi_term"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "winapi 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "atty"
-version = "0.2.10"
+version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.41 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.47 (registry+https://github.com/rust-lang/crates.io-index)",
  "termion 1.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "winapi 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
+name = "autocfg"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "bitflags"
-version = "1.0.3"
+version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "cbindgen"
 version = "0.6.8"
 dependencies = [
- "clap 2.31.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "log 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.64 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_derive 1.0.58 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_json 1.0.19 (registry+https://github.com/rust-lang/crates.io-index)",
- "syn 0.15.23 (registry+https://github.com/rust-lang/crates.io-index)",
- "tempfile 3.0.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


eqrion:
I'm not sure we can do this yet because of Firefox's [rust update policy](https://wiki.mozilla.org/Rust_Update_Policy_for_Firefox). Firefox currently requires only 1.29, and cbindgen is used during the build.

But yes, I definitely want to do this ASAP.

cc @bholley @emilio

bholley:
Yeah, from the stable+2weeks policy, we should be able to require 1.32 on Jan 31, which is a few days into the Firefox 67 Nightly cycle.

konstin:
What's the status here? Having this PR released would really useful as it would finally allow publishing crates depending cbindgen and recent serde versions.

eqrion:
I'll take a look at doing this now that we can require 1.32. Just got back from vacation.

eqrion:
So it looks like Firefox doesn't yet [require 1.32](https://bugzilla.mozilla.org/show_bug.cgi?id=1525733). I've posted a comment requesting that they increment the minimum required version.

eqrion:
Looks like the patch to require 1.32 has been pushed. Once it's merged we should be able to do a release with this.

PR #274, opened 2019-01-18
Add support for ssize_t

### LINKED ISSUE #273: ssize_t is not recognised by cbindgen
When trying to use `libc::ssize_t`, cbindgen complains:

```
WARN: Can't find ssize_t. This usually means that this type was incompatible or not found.
```

RReverser:
Fixes #273.


emilio:
I guess we don't have any for `size_t` either... Still if you could submit them it'd be great :)

PR #282, opened 2019-01-21
Update few more APIs to use AsRef<Path>

### LINKED ISSUE #280: Config::from_file is using str
Most cbindgen APIs use idiomatic `Path` (either directly with `&Path` or via `AsRef<Path>` generics), but `Config::from_file` is still using `&str`. I believe it should be aligned with others.

RReverser:
This makes few more final APIs that work with files consistent with the rest.

Fixes #280.


eqrion:
Thanks!

PR #289, opened 2019-02-04
Added support for includes without default

### LINKED ISSUE #288: Including libs while disabling the default ones
Hey,
I want to add one include without adding all of these:
```
#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>
```

the problem is if I set `no_includes` to true it ignore `includes` and `sys_includes` but If I set it to true I get all of these default includes, is this wanted? what's the right way to do it?

elichai:
Resolves #288

### DIFF
--- a/src/bindgen/bindings.rs
+++ b/src/bindgen/bindings.rs
@@ -112,7 +112,6 @@ impl Bindings {
                 out.write("#include <cstdlib>");
                 out.new_line();
             }
-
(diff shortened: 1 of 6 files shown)


elichai:
ping @eqrion

eqrion:
Looks good, thanks!

PR #291, opened 2019-02-07
Remove hard dependency on serde 1.0.58 (closes #203)

### LINKED ISSUE #203: Find a way to update `serde_derive`

vadixidav:
This PR closes #203 by simply removing the version requirement. This seems to work on current stable 1.32.0 and nightly.


eqrion:
I believe this is a duplicate of #271.

PR #293, opened 2019-02-24
Improve support for bitflags.

### LINKED ISSUE #100: Handle bitflags
It would be nice to be able to use [`bitflags!`](https://github.com/rust-lang-nursery/bitflags) and have nice output generated. Ideally the output of macro expansion would just naturally work.

Currently the output of a `bitflags!` macro is a struct containing an integer, some trait implementations, and an inherent impl that contains the associated constants for each constant from the `bitflags!` macro.

I believe if we were to parse associated constants and some more trait implementations, the expanded output would work. The associated constants part is a little tricky because the constants are rust structs initializers and not integer literals.

Alternatively, we could parse the `bitflags!` macro and output some specially crafted output. This may be easier, but I'm not sure yet.

### LINKED ISSUE #284: Special-case the bitflags! macro
cbindgen now has support for associated constants, but getting that for bitflags! requires macro expansion, which in turn requires nightly Rust. This means we can't use it in Firefox.

As a result, we have to duplicate various lists of bit values, which is error-prone. I think it would be worthwhile to handle bitflags as a special case to auto-generate what we manually do for FontInstanceFlags [1].

CC @emilio @eqrion 

[1] https://searchfox.org/mozilla-central/rev/6c784c93cfbd5119ed07773a170b59fbce1377ea/gfx/webrender_bindings/webrender_ffi.h#99

emilio:
Manually expand the macro using syn and a bit of custom code.

Also, add a way of using bitflags conveniently in C++ (opt-in, at least for now for now, since it uses inline variables, which are only available from C++17).

This also fixes a couple other bugs related to constants and handling or `repr(transparent)`, and adds tests for those.

Fixes #284.


src/bindgen/parser.rs
```diff
@@ -893,4 +913,39 @@ impl Parse {
...
+            Ok(b) => b,
+            Err(e) => {
+                warn!("Failed to parse bitflags invocation: {:?}", e);
+                return;
+            }
+        };
+
+        let (struct_, impl_) = bitflags.expand();
+        self.load_syn_struct(crate_name, mod_cfg, &struct_);
+        self.load_syn_assoc_consts_from_impl(
```
eqrion: IIUC, we defer loading assoc-consts from impls until we have parsed all item definitions because they may be out of order.

This code should work fine, because we load the struct the line before, and the expanded impl will only reference that struct. Might be worth a comment though.

PR #295, opened 2019-03-05
parser: Make bitflags expansion opt-in.

emilio:
Otherwise we may break code that does its own bitflags parsing.

In particular, this will prevent Gecko builds from breaking due to:

  https://searchfox.org/mozilla-central/rev/92d11a33250a8e368c8ca3e962e15ca67117f765/gfx/webrender_bindings/webrender_ffi.h#67

(Which should go away once we start opting-in).

### DIFF
--- a/src/bindgen/builder.rs
+++ b/src/bindgen/builder.rs
@@ -295,7 +295,7 @@ impl Builder {
         }
 
         for x in &self.srcs {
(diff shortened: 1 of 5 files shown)


emilio:
r? @eqrion 

(If this and #294 look good to you, I'd appreciate a release afterwards :P)

eqrion:
Thanks!

PR #304, opened 2019-03-22
bindgen: write header even when no_includes==true

astraw:
Currently (release 0.8.2), if `config.header` is set to `Some(x)` and `config.no_includes` is set to `true`, the header is not written. I think this is a bug and this patch fixes that.

### DIFF
--- a/src/bindgen/bindings.rs
+++ b/src/bindgen/bindings.rs
@@ -158,6 +158,7 @@ impl Bindings {
         if !self.config.no_includes
             || !self.config.includes.is_empty()
             || !self.config.sys_includes.is_empty()
+            || !self.config.header.is_none()
         {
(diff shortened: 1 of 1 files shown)


src/bindgen/bindings.rs
```diff
@@ -158,6 +158,7 @@ impl Bindings {
         if !self.config.no_includes
             || !self.config.includes.is_empty()
             || !self.config.sys_includes.is_empty()
+            || !self.config.header.is_none()
```
emilio: `self.config.header.is_some()`? :)

jschwe:
#352 claims to have already solved this (3 years ago). Can this PR be closed?

emilio:
Yes, let's close this, thanks!

Issue #302, opened 2019-03-20
Configure comment style, or suppress the `*` per line.


ralfbiedert:
For `C` projects, documentation is currently generated like this:

```
  /**
   * Result of f() 
   */
  my_ray combined;
```

In our project we feed the generated header into other tools, which then generate HTML. However, the `*` each line mess up the output format and require manual cleanup each time we regenerate. 

It would be nice to configure the comment style, e.g., being able to select:

- The current style
- `//` each line
- Only a `/*` at the start and an `*/` in the end.

eqrion:
@emilio Can this be closed now?

emilio:
Yes. Fixed by #305.

PR #305, opened 2019-03-23
Documentation style changes

ralfbiedert:
Addresses #302. 

Note that this changes the documentation style generated for `C` bindings to be line with [these conventions](https://en.wikibooks.org/wiki/C++_Programming/Code/Style_Conventions/Comments). 

It also introduces the configuration `documentation_style`, that can be `c`, `cxx`, `doxylight` or `auto` (default).

I'm happy rename `doxylight` (it's not really Doxygen due to lack of `@` attributes, but resembles it), and change what `auto` behavior should be.


tests/expectations/tag/associated_in_body.c
```diff
@@ -3,9 +3,9 @@
 #include <stdint.h>
 #include <stdlib.h>
 
-/**
- * Constants shared by multiple CSS Box Alignment properties
```
emilio: Is there a good reason to change the default behavior?

ralfbiedert: I proposed it because it seemed slightly more in line with the most 'unopinionated' defaults I found online (c. [Wikipedia](https://en.wikipedia.org/wiki/Comment_(computer_programming)#Styles), [Wikibooks](https://en.wikibooks.org/wiki/C++_Programming/Code/Style_Conventions/Comments), and the [C++ reference](https://en.cppreference.com/w/cpp/comment)), and more tooling friendly than the "doxygen" style.

Whether that was a _good_ reason I don't know. As mentioned, I don't mind either way, just let me know what you want.

emilio: As a general rule I'd prefer to keep the behavior the same and allow opting into the behavior change...

Though in this case the risk of breaking something unintendedly seems close to zero...  :)

Maybe @eqrion has opinions?

ralfbiedert: I changed to default for `C`, added a README entry, and renamed `Doxylight` to `Doxy`.

src/bindgen/config.rs
```diff
@@ -100,6 +100,32 @@ impl FromStr for Layout {
 
 deserialize_enum_str!(Layout);
 
+/// How the documentation
```
emilio: This comment seems unfinished?

src/bindgen/ir/documentation.rs
```diff
@@ -39,25 +39,57 @@ impl Source for Documentation {
             return;
         }
 
-        if config.language == Language::C {
-            out.write("/**");
-            out.new_line();
+        let style = match &config.documentation_style {
```
emilio: Just `match config.documentation_style`, and then you can remove `*other` and just use `other`?

ralfbiedert: Yes, done, thanks for review!

Issue #303, opened 2019-03-20
Allow custom `#ifdefs` around function signatures


ralfbiedert:
In our project we would like to generate APIs like this:

```
#ifdef XXX
#define MYDLL declspec(dllimport)
#else

MYDLL int f(int);
```

The function `f` comes from Rust, and the definition should be generated automatically. However, we would like to somehow annotate that `f` receives `MYDLL`, which should be an arbitrary block that can be user-defined. 

Depending on how much cross-compilation is going on, the `MYDLL` block can get large, so it should probably be definable in a separate file.

emilio:
I believe this should be already doable. There's the `fn.prefix` and `fn.postfix` settings that can be set to `MYDLL`. Is that not enough?

emilio:
(or per function like https://searchfox.org/mozilla-central/rev/7c20ad925005fbad7b8e08813115f1ec7fa1c248/gfx/webrender_bindings/src/bindings.rs#689)

ralfbiedert:
Ah, didn't realize `fn.prefix` did this. Do you know how I could then define the custom `ifdef` block that goes along with that?

emilio:
There's the includes and `sys_includes` options to allow you to include arbitrary files.

emilio:
Looks like those are not documented in the README, a patch for doing that would be amazing :)

ralfbiedert:
I just checked the code for `sys_includes`, and from [what it looks like](https://github.com/eqrion/cbindgen/blob/90d26107b8af6b2ac10652494777d238ace04678/src/bindgen/bindings.rs#L145) these only `#include` other files, and not embed them. However, we would like to deliver a single header file, so I think this issue is not fully solved.

emilio:
I think that's a different feature request, but sure.

Issue #211, opened 2018-09-27
C: inject custom code block after includes


grovesNL:
It would be really useful to be able to have a hook after C includes are added (somewhere like https://github.com/eqrion/cbindgen/blob/master/src/bindgen/bindings.rs#L123) but prior to any of the generated type declarations.

Our use case is to have a `#ifdef` to redefine an opaque handle type, like:
```C
#ifdef WGPU_REMOTE
    typedef uint32_t WGPUId;
#else
    typedef void *WGPUId;
#endif
```
However this needs to occur prior to any of the declarations (because they reference this type), but after includes (because it references `uint32_t`).

PR #306, opened 2019-03-24
Embed custom files.

ralfbiedert:
Addresses #303. 

Note that this is a breaking change, as `generate_with_config` now requires a `config_root` to be passed, which specifies relative to what embed requests should be loaded.

It could theoretically be done without breaking that call, but then either

- `config` would have to contain implementation details about where it was loaded from, or
- `[embed]` is likely to break when used as it doesn't know where to get files from

### DIFF
--- a/src/bindgen/bindings.rs
+++ b/src/bindgen/bindings.rs
@@ -6,6 +6,7 @@ use std::fs;
 use std::fs::File;
 use std::io::{Read, Write};
 use std::path;
+use std::path::PathBuf;
 
 use bindgen::config::{Config, Language};
 use bindgen::ir::{
@@ -16,6 +17,7 @@ use bindgen::writer::{Source, SourceWriter};
 /// A bindings header that can be written.
 pub struct Bindings {
     pub config: Config,
+    root: Option<PathBuf>,
     /// The map from path to struct, used to lookup whether a given type is a
     /// transparent struct. This is needed to generate code for constants.
     struct_map: ItemMap<Struct>,
@@ -28,6 +30,7 @@ pub struct Bindings {
 impl Bindings {
     pub(crate) fn new(
         config: Config,
+        root: Option<PathBuf>,
         struct_map: ItemMap<Struct>,
         constants: Vec<Constant>,
         globals: Vec<Static>,
@@ -36,6 +39,7 @@ impl Bindings {
     ) -> Bindings {
         Bindings {
             config,
+            root,
             struct_map,
             globals,
             constants,
@@ -83,6 +87,25 @@ impl Bindings {
         } else {
             false
         }
+    }
+
+    pub fn embed_file<P: AsRef<path::Path>, F: Write>(&self, out: &mut SourceWriter<F>, path: P) {
+        let mut src_content = String::new();
+
+        let src_file = self
+            .root
+            .as_ref()
+            .expect("Bindings::root must be set when embedding files.")
+            .join(path);
+
+        File::open(src_file)
+            .expect("Cannot open file to embed.")
+            .read_to_string(&mut src_content)
+            .expect("Cannot read file to embed.");
+
+        out.new_line_if_not_start();
+        write!(out, "{}", src_content);
+        out.new_line();
     }
 
     pub fn write_headers<F: Write>(&self, out: &mut SourceWriter<F>) {
@@ -160,6 +183,10 @@ impl Bindings {
             || !self.config.sys_includes.is_empty()
         {
             self.write_headers(&mut out);
+        }
+
+        if let Some(file) = &self.config.embed.before {
+            self.embed_file(&mut out, file);
         }
 
         if self.config.language == Language::Cxx {
--- a/src/bindgen/builder.rs
+++ b/src/bindgen/builder.rs
@@ -16,6 +16,7 @@ use bindgen::parser::{self, Parse};
 pub struct Builder {
     config: Config,
     srcs: Vec<path::PathBuf>,
+    root: Option<path::PathBuf>,
     lib: Option<(path::PathBuf, Option<String>)>,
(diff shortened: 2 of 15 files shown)


ralfbiedert:
Note this also addresses #211.

src/bindgen/bindings.rs
```diff
@@ -85,6 +89,25 @@ impl Bindings {
         }
     }
 
+    pub fn embed_file<P: AsRef<path::Path>, F: Write>(&self, out: &mut SourceWriter<F>, path: P) {
+        let mut src_content = String::new();
+
+        let src_file = self
+            .root
+            .as_ref()
+            .expect("Bindings::root must be set when embedding files.")
```
emilio: I don't know, using cwd by default, or config location otherwise doesn't seem too bad.

If I specify the path as an absolute file, which is more than likely when ran from build.rs, it doesn't seem necessary to have to provide a root path... But I'm very bad at API design myself, and this seems ok I guess.

Maybe @eqrion can take a look and comment on this?

src/bindgen/builder.rs
```diff
@@ -152,6 +154,12 @@ impl Builder {
         self
     }
 
+    #[allow(unused)]
```
emilio: Why the `#[allow]`?

src/bindgen/config.rs
```diff
@@ -536,6 +536,22 @@ impl Default for ParseConfig {
...
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "snake_case")]
+#[serde(deny_unknown_fields)]
+#[serde(default)]
+pub struct EmbedConfig {
+    /// An optional file to embed before our generated parts (but after the guards)
+    pub before: Option<String>,
+}
+
+impl Default for EmbedConfig {
```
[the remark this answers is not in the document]
ralfbiedert: I considered it, but there were so many other structs inside `config.rs` that didn't `derive` that I wondered if that was just the preferred style. I think the whole file could use a cleanup, also parsing and case handling.

src/bindgen/bindings.rs
ralfbiedert: I feel strongly that the directory hosting the used `cbindgen.toml` must be the relative anchor for embeds, as that's the only thing you can rely on without thinking too much. Any working dir is random, and the `Cargo.toml` can get mildly ambiguous when generating bindings for a sub-project.

src/bindgen/builder.rs
ralfbiedert: Because every other method for that struct also has it. I didn't want to break that pattern :)

src/bindgen/config.rs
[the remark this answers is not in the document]
eqrion: Yup, that'd be a great clean up but isn't necessary for now.

src/bindgen/bindings.rs
eqrion: I think having these paths be relative to 'cbindgen.toml' makes sense. I'm not great at API design either, so it's not a strong feeling.

src/lib.rs
```diff
@@ -23,19 +23,22 @@ use std::path::Path;
...
+    let config = ConfigLoader::from_root_or_default(crate_dir.as_ref());
+    let root = config.root().clone();
 
-    generate_with_config(crate_dir, config)
+    generate_with_config(crate_dir.as_ref(), &root, config.into())
 }
 
 /// A utility function for build scripts to generate bindings for a crate with a
 /// custom config.
 pub fn generate_with_config<P: AsRef<Path>>(
```
eqrion: Can we keep this API unchanged and default to no root when providing an in-memory configuration?

ralfbiedert: I can do that, but then I would make embed panic if an embed was requested but no root was given.

ralfbiedert:
Just an update. After the problem with the unit tests on Mac #330 I finally got my hands on a Windows machine. 

When I try to run the test cases there (`cargo test`) on the latest master, I now receive a totally different set of error messages, such as

```
test_transparent stdout ----
Running: "D:\\Development\\Source\\_thirdparty\\cbindgen\\target\\debug\\cbindgen" "--lang" "c" "--cpp-compat" "--style" "type" "-o" "D:\\Development\\Source\\_thirdparty\\cbindgen\\tests\\expectations\\transparent.compat.c" "D:\\Development\\Sourc
e\\_thirdparty\\cbindgen\\tests\\rust\\transparent.rs"
Running: "C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\BuildTools\\VC\\Tools\\MSVC\\14.21.27702\\bin\\Hostx64\\x64\\cl.exe" "-D" "DEFINED" "-c" "D:\\Development\\Source\\_thirdparty\\cbindgen\\tests\\expectations\\transparent.compat.c" "
-o" "D:\\Development\\Source\\_thirdparty\\cbindgen\\tests\\expectations\\transparent.compat.o"
thread 'test_transparent' panicked at 'Output failed to compile: Output { status: ExitStatus(ExitStatus(2)), stdout: "transparent.compat.c\r\nD:\\Development\\Source\\_thirdparty\\cbindgen\\tests\\expectations\\transparent.compat.c(1): fatal error
C1034: stdarg.h: no include path set\r\n", stderr: "Microsoft (R) C/C++ Optimizing Compiler Version 19.21.27702.2 for x64\r\nCopyright (C) Microsoft Corporation.  All rights reserved.\r\n\r\ncl : Command line warning D9035 : option \'o\' has been d
eprecated and will be removed in a future release\r\n" }', tests\tests.rs:82:5
```

While the specific error message changes with the type of environment I load, the result is basically that I can't run the test cases to verify what I do works. 

I will close this pull request for now because we worked around the original embedding problem and I'm just sidetracking. I would like to keep #303 open though because it still has value, and it would have saved us work in the first place.

PR #307, opened 2019-03-25
ir: Add support for #[must_use].

emilio:


tests/rust/must-use.toml
```diff
@@ -0,0 +1,15 @@
+# Quite an abuse...
+autogen_warning = """
```
eqrion: Can we use the 'header' option for this?

emilio: Yup! Done

PR #312, opened 2019-03-30
Test crates

AxelNennker:
How about creating small test creates for all the integration tests and moving tests out of test.py.
So, in the end the integration tests can be done with usual `cargo test` instead of needing test.py

Doing this for all integration tests is work but it is mostly reshuffling files

### DIFF
--- a/src/bindgen/ir/constant.rs
+++ b/src/bindgen/ir/constant.rs
@@ -251,7 +251,7 @@ impl Constant {
         };
 
         if !can_handle(&ty, expr) {
-            return Err("Unhanded const definition".to_owned());
+            return Err("Unhandled const definition".to_owned());
         }
 
         let mut lit = Literal::load(&expr)?;
--- a/test_crates/constant/Cargo.lock
+++ b/test_crates/constant/Cargo.lock
@@ -1,0 +1,4 @@
+[[package]]
+name = "test"
+version = "0.0.1"
+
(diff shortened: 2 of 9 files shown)


emilio:
I don't think we should make adding tests harder. It should be possible to use `cargo test` and keeping all our tests as-is / making them equally easy to add.

AxelNennker:
two other advantages, besides being able to use `cargo test`,  I see with putting each integration test into its own crate are:

- the Rust files and the expected bindings are in one directory. So it is easier to see what belongs together.
- cbindgen works on crates and this way the cbindgen tests do as well.

emilio:
> * the Rust files and the expected bindings are in one directory. So it is easier to see what belongs together.

That's completely orthogonal to whether start using `cargo test`, right? Where to put the expectations can be changed at any time.

> * cbindgen works on crates and this way the cbindgen tests do as well.

`cbindgen` also works on plain files too, but anyhow, we already have support to test crates (all the subdirectories on `tests/rust` are crates).

The only reason not all tests are written that way (afaik) is that adding and changing tests should be convenient.

This is another approach that I think would work better:

  https://github.com/eqrion/cbindgen/compare/master...emilio:cargo-test

The end result is that `cargo test` would work, tests would not have to move, and adding a test will be similarly convenient. If we want to change the place test expectations live in afterwards, that's a different matter.

AxelNennker:
Cool

PR #314, opened 2019-04-08
Respect visibility of constants

### LINKED ISSUE #123: Don't include any constants besides those that are public for consumers outside of the module in the header file
Any constants that aren't public for consumers outside of the module shouldn't be included in the output header file. Currently, even private constants lead to `#define`s being created.

RReverser:
Fixes #123:
 - Adds visibility check for module-level constants.
 - Adds visibility check for associated constants.
 - Fixes bitflags expansion to produce public associated constants (as the real expansion does).

Admittedly, because it's a long-standing issue, it's now a breaking change for users who started to rely on current behaviour :(

[](https://xkcd.com/1172/)

That said, I think this is a right change as current behaviour is not compliant with Rust visibility rules and produces lots of noise for unintentionally exported constants.

### DIFF
--- a/src/bindgen/parser.rs
+++ b/src/bindgen/parser.rs
@@ -691,6 +691,12 @@ impl Parse {
         let impl_path = ty.unwrap().get_root_path().unwrap();
 
         for item in items.into_iter() {
+            if let syn::Visibility::Public(_) = item.vis {
+            } else {
+                warn!("Skip {}::{} - (not `pub`).", crate_name, &item.ident);
+                return;
+            }
+
             let path = Path::new(item.ident.to_string());
             match Constant::load(
                 path,
@@ -736,6 +742,12 @@ impl Parse {
                 "Skip {}::{} - (const's outside of the binding crate are not used).",
                 crate_name, &item.ident
             );
+            return;
+        }
+
+        if let syn::Visibility::Public(_) = item.vis {
+        } else {
+            warn!("Skip {}::{} - (not `pub`).", crate_name, &item.ident);
             return;
         }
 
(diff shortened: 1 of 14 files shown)


src/bindgen/parser.rs
```diff
@@ -691,6 +691,12 @@ impl Parse {
         let impl_path = ty.unwrap().get_root_path().unwrap();
 
         for item in items.into_iter() {
+            if let syn::Visibility::Public(_) = item.vis {
```
emilio: I think these would be nicer with `match`.

RReverser: I literally copied the visibility check as it already is in other places. Should I change all of them for consistency?

emilio: That'd be great, though I guess we can also change them all together in a followup PR. Your call :)

emilio: Also it looks like this needs a rebase for #318.

RReverser: I've rebased and fixed conflicts. As for visibility check - I started doing it but got a bit derailed in refactoring whole helper traits part and barely stopped :P

I think I'd rather keep this PR scoped and consistent with rest of code for now.

RReverser:
> Maybe we should add an option to keep the existing behavior in an opt-in way?

Hence the XKCD above with literally same suggestion :) But IMO not, if it's breaking someone's code, they're already not exporting it correctly.

emilio:
Yeah, fair :)

Ouch, looks like it needs a rebase again probably because of #321. I'll rebase myself if you don't get to it.

RReverser:
Oh. I won't get to it till tomorrow now.

emilio:
No worries, just rebased for you :)

Will merge as soon as it's green

emilio:
(Just realized @eqrion is on PTO)

RReverser:
Thanks!

PR #315, opened 2019-04-08
Add C99 doc comment style

RReverser:
In modern C (post-C99) it's common to just use `// double-slash comments`, but currently cbindgen provides only `/* block-comment */` and `/// triple-slash comment` variants.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ language = "[C|C++]"
 style = "[Both|Type|Tag]"
 # How the generated documentation should be commented.
 # C uses /* */; C++ uses //; Doxy is like C but with leading * per line.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -106,7 +106,7 @@ language = "[C|C++]"
 style = "[Both|Type|Tag]"
 # How the generated documentation should be commented.
 # C uses /* */; C++ uses //; Doxy is like C but with leading * per line.
```
emilio: Can you update this comment too?

RReverser: Yeah, although it looks like C++ comment was wrong in the first place? (it uses `///` not `//`)

RReverser: Updated both.

PR #316, opened 2019-04-08
Problem: conditional structure members are always present

yrashk:
Given something like:

```rust
struct Foo {
  #[cfg(unix)]
  bar: i32
}
```

we will always get a structure with `bar` field, regardless of
conditions.

Solution: pass conditional configuration for fields
and generate conditional #ifdef for them as well.

Current solution isn't absolutely perfect as it still inserts
semicolons after the conditions; but at least in the meanwhile this
seems to be an adequate solution.

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -67,7 +67,8 @@ impl EnumVariant {
         fn parse_fields(
             is_tagged: bool,
             fields: &syn::punctuated::Punctuated<syn::Field, syn::token::Comma>,
-        ) -> Result<Vec<(String, Type, Documentation)>, String> {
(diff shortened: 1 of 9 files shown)


emilio:
Ugh, sorry, I somehow missed this patch. Really sorry! :(

This looks sensible and a good fix to have, if you could rebase it.

yrashk:
Unfortunately, I don't have enough time to update this (if this is not a
trivial rebase) as I'm focused on other projects right now. Feel free to
use it as a base for your own patch should you deem it necessary, though.
I'll be happy if it's useful in any way.

petrochenkov:
Rebased in https://github.com/eqrion/cbindgen/pull/633, this PR can be closed.

PR #319, opened 2019-04-14
Use cargo test instead of test.py

emilio:
This makes `cargo test` also test all the stuff in `tests/rust`, and thus makes tests.py unnecessary.

Thanks a lot to @AxelNennker for helping out with this immensely.


emilio:
r? @eqrion 

This is basically a line-by-line port of tests.py to do the same as it does now, but from cargo test.

An easy way to confirm it works as expected is applying this diff:

```diff
diff --git a/tests/tests.rs b/tests/tests.rs
index dfea81f..1cedfe7 100644
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -89,7 +89,7 @@ fn run_compile_test(
     style: Option<Style>,
 ) {
     let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
-    let mut output = Path::new(&crate_dir).join("tests").join("expectations");
+    let mut output = Path::new(&crate_dir).join("tests").join("cargo-expectations");
     if let Some(style) = style {
         match style {
             Style::Both => {
```

Then running `cargo test`, then doing: `diff -r tests/expectations tests/cargo-expectations`, and ensuring it generates no output.

RReverser:
Yessss. I did this in the past, but ran into few cross-platform issues and so never sent a PR.

Could Windows and OS X be added to travis.yml before merging this? (probably best to do as part of this PR, as that's the main benefit of migrating away from Python runner)

emilio:
> Could Windows and OS X be added to travis.yml before merging this? (probably best to do as part of this PR, as that's the main benefit of migrating away from Python runner)

Does travis do windows these days? Didn't know :)

Can certainly give that a shot.

RReverser:
Yes, you just need to list operating systems under `os` key. Mind you, Windows can be currently slower than others, but I think it's worth it for such a tool as cbindgen that needs to produce cross-platform C code.

emilio:
I gave up on OSX / Windows CI for now because I don't have the time to figure out how to get a suitable compiler that supports all the C++ features that cbindgen uses. I suspect we could enable most of them passing flags to the default compiler, but that feels like a different PR :)

Issue #320, opened 2019-04-14
Single-quote chars are wrapped incorrectly


RReverser:
Just something I've noticed while looking at the code of #318 and confirmed by manual tests: trying to export a single quote char results in invalid output:

Rust:
```rust
pub const QUOTE: char = '\'';
```

C:
```c
#define QUOTE '''
```

C++
```cpp
static const wchar_t QUOTE = ''';
```

Unfortunately, noticed this only after the PR was merged, so raising as a separate tracking issue.

cc @AxelNennker

RReverser:
Same probably applies to other special chars like `\n` expanded into a newline. Worth extending tests to include various cases and use something like [`char::escape_default`](https://doc.rust-lang.org/std/primitive.char.html#method.escape_default) to cover all of them.

emilio:
Ah, nice catch.

PR #321, opened 2019-04-15
escape chars

AxelNennker:
Signed-off-by: Axel Nennker <axel.nennker@telekom.de>

Fix https://github.com/eqrion/cbindgen/issues/320

### DIFF
--- a/src/bindgen/ir/constant.rs
+++ b/src/bindgen/ir/constant.rs
@@ -151,7 +151,15 @@ impl Literal {
             syn::Expr::Lit(syn::ExprLit {
                 lit: syn::Lit::Char(ref value),
                 ..
-            }) => Ok(Literal::Expr(format!("'{}'", value.value().escape_default().to_string().replace('{', "").replace('}', "")))),
+            }) => Ok(Literal::Expr(format!(
+                "'{}'",
+                value
+                    .value()
+                    .escape_default()
+                    .to_string()
+                    .replace('{', "")
+                    .replace('}', "")
+            ))),
             syn::Expr::Lit(syn::ExprLit {
                 lit: syn::Lit::Int(ref value),
                 ..
--- a/src/bindgen/ir/constant.rs
+++ b/src/bindgen/ir/constant.rs
@@ -151,15 +151,25 @@ impl Literal {
             syn::Expr::Lit(syn::ExprLit {
                 lit: syn::Lit::Char(ref value),
                 ..
-            }) => Ok(Literal::Expr(format!(
-                "'{}'",
-                value
-                    .value()
-                    .escape_default()
-                    .to_string()
-                    .replace('{', "")
-                    .replace('}', "")
-            ))),
+            }) => {
+                let v = value.value();
+                if v.len_utf8() == 1 {
+                    Ok(Literal::Expr(format!(
+                        "'{}'",
+                        value.value().escape_default()
+                    )))
+                } else {
+                    Ok(Literal::Expr(format!(
+                        "L'{}'",
+                        value
+                            .value()
+                            .escape_default()
+                            .to_string()
+                            .replace('{', "")
+                            .replace('}', "")
+                    )))
+                }
+            }
             syn::Expr::Lit(syn::ExprLit {
                 lit: syn::Lit::Int(ref value),
                 ..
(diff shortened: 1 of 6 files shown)


tests/rust/constant.rs
```diff
@@ -1,6 +1,9 @@
 const FOO: i32 = 10;
 const BAR: &'static str = "hello world";
 pub const DELIMITER: char = ':';
+pub const QUOTE: char = '\'';
+pub const TAB: char = '\t';
+pub const NEWLINE: char = '\n';
```
RReverser: Can you add some Unicode char examples too?

AxelNennker: I could but thought to do this step by step. The docu has this nice '❤' character.
https://doc.rust-lang.org/std/primitive.char.html#method.escape_unicode
Maybe escape_default does all this for free.

RReverser: To clarify - we probably need at least two examples here, with different handling:
 - 1) Any character with code point <= 0xFFFF - fits into `wchar_t` and should be written out as such.
 - 2) Any character with large code point - either produces a multichar or at least errors out.

RReverser: It seems better to fix the issue at once (especially since this PR will automatically close it when merged).

Unfortunately, `escape_default` uses a different escaping than C/C++ so you might still need to do something manually here.

tests/expectations/both/constant.c
```diff
@@ -7,6 +7,14 @@
 
 #define FOO 10
 
+#define HEART '\u{2764}'
```
RReverser: So yeah, as test log on Travis currently says:

```
tests/expectations/constant.cpp:9:30: error: incomplete universal character name \u
 static const wchar_t HEART = '\u{2764}';
                              ^~~~~~~~~~
```

this will require some custom escaping.

AxelNennker: I changed the title to WIP because I am not sure what the correct way forward is.

Filtering the curly braces 
`value.value().escape_default().to_string().replace('{', "").replace('}', "")`
produces
`static const wchar_t HEART = '\u2764';` 
which leads to a warning 
```ignisvulpis@namenlos:~/development/cbindgen$ g++ -c /tmp/x.cpp 
/tmp/x.cpp:9:30: warning: multi-character character constant [-Wmultichar]
 static const wchar_t HEART = '\u2764';
```
Suggestions welcome.

RReverser: I think the correct output should be

```cpp
static const wchar_t HEART = L'\u2764';
```

Also note that `.replace('{', "").replace('}', "")` is not a great way to go about it, since now you are probably corrupting constants like `const c: char = '{';`.

Instead, you should check the character code and, if it's larger than 255, format the escape sequence manually.

src/bindgen/ir/constant.rs
```diff
@@ -151,7 +151,25 @@ impl Literal {
...
+                let v = value.value();
+                if v.len_utf8() == 1 {
+                    Ok(Literal::Expr(format!(
+                        "'{}'",
+                        value.value().escape_default()
+                    )))
+                } else {
+                    Ok(Literal::Expr(format!(
+                        "L'{}'",
+                        value
```
RReverser: As mentioned earlier, it would be easier to avoid `escape_default` for this case altogether and just format manually as a hexadecimal integer with something like:

```
format!(r"L'\u{:X}'", value.value() as u32)
```

src/bindgen/ir/constant.rs
```diff
@@ -151,7 +151,25 @@ impl Literal {
             syn::Expr::Lit(syn::ExprLit {
                 lit: syn::Lit::Char(ref value),
                 ..
-            }) => Ok(Literal::Expr(format!("'{}'", value.value()))),
+            }) => {
+                let v = value.value();
+                if v.len_utf8() == 1 {
+                    Ok(Literal::Expr(format!(
```
RReverser: Note that you can extract common parts out:

```rust
Ok(Literal::Expr(match v as u32 {
  0..=255 => format!("'{}'", v.escape_default()),
  other_code => format!(r"L'\u{:X}'", other_code),
}))
```

AxelNennker: nice!

emilio:
Thanks a lot for the help reviewing @RReverser!

PR #323, opened 2019-04-16
fix #205 Negative literals

### LINKED ISSUE #205: Negative literals seem to be unsupported
I have found that negative const literals are not supported. It seems like they should be.
I made a simple rust crate to demonstrate:
```rust
const POS_ONE: i8 = 1;
const NEG_ONE: i8 = -1;

#[no_mangle]
pub extern fn hello() {
    println!("POS_ONE is {}, NEG_ONE is {}", POS_ONE, NEG_ONE);
}
```

Here are the command lines I used and the warning I got:
```
C:\projects>cargo new cbind_neg_lit --lib
     Created library `cbind_neg_lit` project

C:\projects>code cbind_neg_lit

C:\projects>cd cbindgen

C:\projects\cbindgen>cargo run ..\cbind_neg_lit -o ..\cbind_neg_lit\cbind_neg_lit.h
    Finished dev [unoptimized + debuginfo] target(s) in 0.20s
     Running `target\debug\cbindgen.exe ..\cbind_neg_lit -o ..\cbind_neg_lit\cbind_neg_lit.h`
WARN: Skip cbind_neg_lit::NEG_ONE - (Unsupported literal expression.)

C:\projects\cbindgen>
```
And here is the file that was generated:
```c
#include <cstdint>
#include <cstdlib>

static const int8_t POS_ONE = 1;

extern "C" {

void hello();

} // extern "C"
```

AxelNennker:
Signed-off-by: Axel Nennker <axel.nennker@telekom.de>

Fix https://github.com/eqrion/cbindgen/issues/205

### DIFF
--- a/src/bindgen/ir/constant.rs
+++ b/src/bindgen/ir/constant.rs
@@ -19,6 +19,7 @@ use bindgen::ir::{
 use bindgen::library::Library;
 use bindgen::writer::{Source, SourceWriter};
 use bindgen::Bindings;
+use syn::UnOp;
 
 #[derive(Debug, Clone)]
 pub enum Literal {
@@ -209,7 +210,19 @@ impl Literal {
                     fields: field_pairs,
                 })
             }
-            _ => Err("Unsupported literal expression.".to_owned()),
+            // Unary(ExprUnary { attrs: [], op: Neg(Sub), expr: Lit(ExprLit { attrs: [], lit: Int(LitInt { token: Literal { lit: 1 } }) }) }))
+            syn::Expr::Unary(syn::ExprUnary {
+                attrs: _,
+                ref op,
+                ref expr,
+            }) => match *op {
+                UnOp::Neg(_) => {
+                    let val = Self::load(expr)?;
+                    Ok(Literal::Expr(format!("-{}", val)))
+                }
+                _ => Err(format!("Unsupported Unary expression. {:?}", *op)),
+            },
+            _ => Err(format!("Unsupported literal expression. {:?}", *expr)),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/bindgen/ir/constant.rs
```diff
@@ -209,7 +210,19 @@ impl Literal {
...
                 })
             }
-            _ => Err("Unsupported literal expression.".to_owned()),
+            // Unary(ExprUnary { attrs: [], op: Neg(Sub), expr: Lit(ExprLit { attrs: [], lit: Int(LitInt { token: Literal { lit: 1 } }) }) }))
+            syn::Expr::Unary(syn::ExprUnary {
+                attrs: _,
+                ref op,
+                ref expr,
+            }) => match *op {
+                UnOp::Neg(_) => {
```
RReverser: While at it, maybe worth adding `!` as well? Seems easy enough to add another branch.

AxelNennker: Well, I tried to keep it simple as I did this in my free time on a week end.
Negative integer constants seem to be common while `pub const FALSE: bool = !true;` looks silly.
I would wait for somebody to have this problem.
There are other things I would like cbindgen being able to handle like 
`WARN: Skip constant::BAR - (Unhandled const definition)` from 
`const BAR: &'static str = "hello world";` which is very common, I think.

RReverser: Fair enough :) Sorry for derailing the PR.

src/bindgen/ir/constant.rs
```diff
@@ -209,7 +210,19 @@ impl Literal {
                     fields: field_pairs,
                 })
             }
-            _ => Err("Unsupported literal expression.".to_owned()),
+            // Unary(ExprUnary { attrs: [], op: Neg(Sub), expr: Lit(ExprLit { attrs: [], lit: Int(LitInt { token: Literal { lit: 1 } }) }) }))
```
emilio: I'd remove this comment.

PR #333, opened 2019-05-02
ir: Add the possibility to auto-generate destructors of tagged unions.

emilio:
This trades the ABI-safety (making the struct potentially unsafe to pass by
value), thus the opt-in rather than auto-generating it.

The TL;DR is that I need a way to share types that contain boxed slices from the
style system. That's fine when they're inside regular structs, since C++ does
the right thing, but for unions you need to make it explicit like this.

The actual Gecko setup will hook into the leak logging machinery to avoid silly
leaks and such of course.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,9 @@ derive_gte = false
 [enum]
 # A rule to use to rename enum variants
 rename_variants = "[None|GeckoCase|LowerCase|UpperCase|PascalCase|CamelCase|SnakeCase|ScreamingSnakeCase|QualifiedScreamingSnakeCase]"
+# Whether tagged enums should generate destructors. This makes them dangerous to
+# pass by value.
+derive_tagged_enum_destructor = false
 
 ```
 
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -398,6 +398,8 @@ pub struct EnumConfig {
     pub cast_assert_name: Option<String>,
     /// The way to annotation this enum as #[must_use].
     pub must_use: Option<String>,
+    /// Whether to generate destructors of tagged enums.
+    pub derive_tagged_enum_destructor: bool,
 }
 
 impl EnumConfig {
@@ -424,6 +426,12 @@ impl EnumConfig {
             return x;
         }
         self.derive_mut_casts
+    }
+    pub(crate) fn derive_tagged_enum_destructor(&self, annotations: &AnnotationSet) -> bool {
+        if let Some(x) = annotations.bool("derive-tagged-enum-destructor") {
+            return x;
+        }
+        self.derive_tagged_enum_destructor
     }
 }
 
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -873,7 +873,9 @@ impl Source for Enum {
             }
 
             if config.language == Language::Cxx
-                && self.annotations.bool("destructor").unwrap_or(false)
+                && config
+                    .enumeration
+                    .derive_tagged_enum_destructor(&self.annotations)
             {
                 out.new_line();
                 out.new_line();
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -872,6 +872,33 @@ impl Source for Enum {
                 }
             }
 
+            if config.language == Language::Cxx
+                && self.annotations.bool("destructor").unwrap_or(false)
+            {
+                out.new_line();
+                out.new_line();
+                write!(out, "~{}()", self.export_name);
+                out.open_brace();
+                write!(out, "switch (tag)");
+                out.open_brace();
+                for variant in &self.variants {
+                    if let Some((ref variant_name, ref item)) = variant.body {
+                        write!(
+                            out,
+                            "case {}::{}: {}.~{}(); break;",
+                            self.tag.as_ref().unwrap(),
+                            variant.export_name,
+                            variant_name,
+                            item.export_name(),
+                        );
+                        out.new_line();
+                    }
+                }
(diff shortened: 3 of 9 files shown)


emilio:
r? @eqrion 

I'd prefer an speedy review of this so that @Gankro can update cbindgen in Gecko with this change too (so as to minimize other dev's disruption). But only if this is uncontroversial generally. If not I'm happy to wait.

emilio:
Oh I thought #332 had landed already. Then forget about speediness :)

Gankro:
What's the point of this if we don't ever emit struct destructors? Can't this only ever run no-op (POD) destructors?

Gankro:
The implementation otherwise seems fine, although I'd appreciate a `repr(C, u8)` enum test as well.

emilio:
> What's the point of this if we don't ever emit struct destructors? Can't this only ever run no-op (POD) destructors?

Structs do the right thing by default in C++, it's only when they contain unions when the compiler cannot infer it.

Gankro:
But that's the exact same problem, where are types with real destructors showing up in the system? Template substitutions?

emilio:
> But that's the exact same problem, where are types with real destructors showing up in the system? Template substitutions?

I noted in the example, you can add a destructor manually via either type substitution or arbitrary body contents (`export.body`).

Gankro:
Ok so just to further clarify how we want to use this, will the non-trivial-destructor types live strictly in C++, or will they actually get passed across the FFI boundary (by reference, of course)? If the latter, will the Rust side also have equivalent Drop impls?

emilio:
> Ok so just to further clarify how we want to use this, will the non-trivial-destructor types live strictly in C++, or will they actually get passed across the FFI boundary (by reference, of course)? If the latter, will the Rust side also have equivalent Drop impls?

Rust will "move" the types into C++ (putting them in a struct for example, or passing them by reference), and C++ is the responsible to drop them.

RReverser:
> you can add a destructor manually via either type substitution or arbitrary body contents (export.body).

Why the same can't be done for tagged enums? I'm all for supporting destructors in cbindgen, but it feels wrong to support them only for tagged enums in a special manner, different from regular structs...

Gankro:
This feature is more-so creating greater parity, as C++'s native understanding of structs automagically gets us default destructors that just drop all the fields, but C++ doesn't natively understand tagged unions, so we need to go out of our way to tell it what the default destructor should be.

However, because it is a user-defined destructor, it makes the type non-trivial, and therefore non-pod. It *should* still have standard layout, though (I Am Not A C++ Language Lawyer).

emilio:
> Why the same can't be done for tagged enums? I'm all for supporting destructors in cbindgen, but it feels wrong to support them only for tagged enums in a special manner, different from regular structs...

Right, regular structs already destruct on a per-field basis. You can override this with `export.body`.

RReverser:
> Right, regular structs already destruct on a per-field basis. You can override this with export.body.

Yes, but this works only for POD structs, right? IIRC, cbindgen doesn't expose custom `Drop` for structs?

emilio:
Well, it works if you have a struct with an inner type that has a destructor, so if you have:

```
struct Foo { unsigned something; ~Foo(); };

struct Bar { Foo b; }; // This is cbindgen-generated.
```

Dropping `Bar` will correctly drop `Foo` in C++ But yeah, you can only generate the destructor in `Foo` with `export.body` or such, it's not clear how would you go about automatically translating drop implementations.

RReverser:
> it's not clear how would you go about automatically translating drop implementations

I was actually thinking about this for a while - could cbindgen take similar approach as bindgen does in terms of relying on mangled symbols? (but in the opposite direction)

Currently any `Drop` implementation of a public type is kept in the output binary, e.g.:

```rust
pub struct Foo;

impl Drop for Foo {
    fn drop(&mut self) {}
}
```

emits

```asm
_ZN54_$LT$example..Foo$u20$as$u20$core..ops..drop..Drop$GT$4drop17h05763556a065b101E:
        ret
```

and it should be possible to mangle to the same name and link in C/C++ headers.

Not sure if it's actually feasible or aligns with cbindgen goals though :)

emilio:
> Currently any Drop implementation of a public type is kept in the output binary, e.g.:

Even in optimized builds? That'd be surprising. `bindgen` can only generate non-inline destructors.

> and it should be possible to mangle to the same name and link in C/C++ headers.

Does C++ have something similar to the `#[link_name]` attribute in rust? Even if so, cbindgen right now doesn't have any info from rustc itself, so it'd be quite hard to implement / fragile.

RReverser:
> Even in optimized builds? That'd be surprising. bindgen can only generate non-inline destructors.

Yup, always, just like any public API methods.

> Does C++ have something similar to the #[link_name] attribute in rust?

I'd expect generated destructor to be defined in C++ but also to link the Rust destructor by its mangled name internally.

> so it'd be quite hard to implement / fragile

Yeah this is true.

emilio:
> Yup, always, just like any public API methods.

Just checked and it doesn't do it for `#[inline]` destructors, or generics, for that matter.

> I'd expect generated destructor to be defined in C++ but also to link the Rust destructor by its mangled name internally.

Right, may point is how to do this (`asm("call $symbol")` I guess?)

RReverser:
> or generics, for that matter

Oh yeah, definitely not for generics, because they have to be monomorphised. Surprised about inline marker though :/

> Right, may point is how to do this (asm("call $symbol") I guess?)

Hmm, I would've just used an extern function definition.

emilio:
Symbols like `_ZN54_$LT$example..Foo$u20$as$u20$core..ops..drop..Drop$GT$4drop17h05763556a065b101E` aren't valid C function names AFAIK.

PR #336, opened 2019-05-07
Use placement new for constructing in tagged unions' helper methods.

emilio:
Using operator= is not quite sound in presence of destructors and operator
overloading.

It's perfectly fine to assume that the left-hand-side of an operator= expression
is valid memory, however we're using uninitialized memory here, that may not be
the case.

Use placement new to properly construct tagged unions. I don't need this with
any urgency, but it's the right thing to do in presence of complex types, and
the current code seems a bomb waiting to explode :)

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -700,28 +700,28 @@ impl Source for Enum {
                         for &(ref field_name, ref ty, ..) in body.fields.iter().skip(skip_fields) {
                             out.new_line();
                             match ty {
-                                Type::Array(ref _ty, ref length) => {
+                                Type::Array(ref ty, ref length) => {
                                     // arrays are not assignable in C++ so we
                                     // need to manually copy the elements
+                                    write!(out, "for (int i = 0; i < {}; i++)", length.as_str());
+                                    out.open_brace();
                                     write!(
                                         out,
-                                        "for (int i = 0; i < {}; i++) {{\
-                                         result.{}.{}[i] = {}[i];\
-                                         }}",
-                                        length.as_str(),
-                                        variant_name,
-                                        field_name,
-                                        arg_renamer(field_name)
+                                        "::new (&result.{}.{}[i]) (",
+                                        variant_name, field_name
                                     );
+                                    ty.write(config, out);
+                                    write!(out, ")({}[i]);", arg_renamer(field_name));
+                                    out.close_brace(false);
                                 }
-                                _ => {
+                                ref ty => {
                                     write!(
                                         out,
-                                        "result.{}.{} = {};",
-                                        variant_name,
-                                        field_name,
-                                        arg_renamer(field_name)
+                                        "::new (&result.{}.{}) (",
+                                        variant_name, field_name
                                     );
+                                    ty.write(config, out);
+                                    write!(out, ")({});", arg_renamer(field_name));
                                 }
                             }
                         }
--- a/tests/expectations/annotation.cpp
+++ b/tests/expectations/annotation.cpp
@@ -1,6 +1,7 @@
 #include <cstdarg>
 #include <cstdint>
 #include <cstdlib>
+#include <new>
 
 enum class C : uint32_t {
   X = 2,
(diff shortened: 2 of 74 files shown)


src/bindgen/ir/enumeration.rs
```diff
@@ -700,28 +700,24 @@ impl Source for Enum {
...
-                                         result.{}.{}[i] = {}[i];\
-                                         }}",
-                                        length.as_str(),
-                                        variant_name,
-                                        field_name,
-                                        arg_renamer(field_name)
+                                        "new (result.{}.{} + i) ",
+                                        variant_name, field_name
                                     );
+                                    ty.write(config, out);
```
mystor: Writing out the bare type in a new expression usually works out, but with some more complex types like function pointers, it's generally a better idea to write out the type itself in parentheses:

```c++
::new (&result.foo.x[i]) (int32_t)();
// -or, with a function pointer-
::new (&result.foo.x[i]) (void (*) (int32_t)) (); // max readable
```

src/bindgen/ir/enumeration.rs
```diff
@@ -700,28 +700,24 @@ impl Source for Enum {
...
                                     write!(
                                         out,
-                                        "for (int i = 0; i < {}; i++) {{\
-                                         result.{}.{}[i] = {}[i];\
-                                         }}",
-                                        length.as_str(),
-                                        variant_name,
-                                        field_name,
-                                        arg_renamer(field_name)
+                                        "new (result.{}.{} + i) ",
```
mystor: As we want to be sure to use the global new operator here (provided by `<new>`), it's probably worthwhile to generate the leading `::` to force scope resolution to the global operator.

I also generally prefer avoiding using array->pointer implicit conversions like those here, so imo it might be nicer to write `&result.{}.{}[i]`.

tests/expectations/annotation.cpp
```diff
@@ -53,16 +54,16 @@ union F {
...
     return result;
   }
 
   static F Bar(const uint8_t &aX,
                const int16_t &aY) {
     F result;
-    result.bar.x = aX;
-    result.bar.y = aY;
+    new (&result.bar.x) uint8_t(aX);
+    new (&result.bar.y) int16_t(aY);
```
mystor: As more of a general note, I'm pretty sure you could technically placement new the entire structure at once using direct aggregate initialization. Keeping it this way could be cleaner though, due to the trickiness involved with fixed sized arrays.

```c++
new (&result.bar) Bar_Body{aX, aY};
```

src/bindgen/ir/enumeration.rs
```diff
@@ -700,28 +700,24 @@ impl Source for Enum {
                         for &(ref field_name, ref ty, ..) in body.fields.iter().skip(skip_fields) {
                             out.new_line();
                             match ty {
-                                Type::Array(ref _ty, ref length) => {
+                                Type::Array(ref ty, ref length) => {
                                     // arrays are not assignable in C++ so we
                                     // need to manually copy the elements
```
mystor: I'm a bit curious why you folks use arrays like `int32_t[32]` rather than the c++ `std::array<int32_t, 32>` (which IIRC is required by spec to be standard layout & trivial, so OK to use over FFI). I'm guessing some calling conventions care about the distinction between arrays and arrays in aggregates when passing by value?

The stdlib type has the advantage of being copyable etc. which would mean that this sort of braced-initialization wouldn't be necessary.

src/bindgen/ir/enumeration.rs
emilio: Good point

src/bindgen/ir/enumeration.rs
emilio: Fair :)

tests/expectations/annotation.cpp
emilio: Yeah, I'll keep it like this for now.

src/bindgen/ir/enumeration.rs
emilio: I don't use them :-)

emilio:
```
18:03 <emilio> I addressed your comments in https://github.com/eqrion/cbindgen/pull/336, and will apply the ::new() suggestion over in the other PR
18:03 <emilio> I assume it looks good to you with that?
18:06 <nika> Yeah, looks good.
```

PR #339, opened 2019-05-10
Add the ability to automatically derive copy-constructors for tagged enums.

emilio:
This is useful if some of the variants contain non-trivial types, in which case C++ will automatically delete them unless manually implemented.


emilio:
(Only last commit, the previous is #336)

src/bindgen/ir/enumeration.rs
```diff
@@ -897,6 +897,42 @@ impl Source for Enum {
...
+                            self.tag.as_ref().unwrap(),
+                            variant.export_name,
+                            variant_name,
+                            item.export_name(),
+                            variant_name,
+                        );
+                        out.new_line();
+                    }
+                }
+                write!(out, "default: break;");
```
mystor: You should probably check if you have any non-trivial variants before generating the switch, because IIRC MSVC emits warnings & such for switch statements with only one labeled 'default' case.

src/bindgen/ir/enumeration.rs
```diff
@@ -897,6 +897,42 @@ impl Source for Enum {
...
+                out.close_brace(false);
+                out.close_brace(false);
+            }
```
mystor: Might be worthwhile to, while here, also generate a `T&&` constructor. 

You'd just need to `std::move` from the passed-in values, you don't need to change the enum variant or anything like that.

src/bindgen/ir/enumeration.rs
emilio: We only get here with non-trivial variants, otherwise we generate a plain `enum` instead of a struct / union.

src/bindgen/ir/enumeration.rs
emilio: I think so, though it'd need to be another configuration option (since you may have a non-movable type in the enumeration and you may want a copy-ctor but not a move-ctor, or it may not even compile).

emilio: (So I think I'll defer it for another PR)

emilio:
I'll call this sanity-checked-by-nika :)

PR #349, opened 2019-05-24
Added generation of header files that support both C and C++ compiler

Wodann:
I added the option to generate header files that will successfully compile in both C and C++ compilers. You can opt in to this feature by setting the following setting in the `cbindgen.toml`:

```language = "Both"```

To create this feature, I had to make enums have both a tag and a typedef. As such, my code does not check whether users have opted into these styles.

### DIFF
--- a/tests/expectations/alias.compat.c
+++ b/tests/expectations/alias.compat.c
@@ -12,10 +12,8 @@ enum Status
   Err,
 };
 #ifndef __cplusplus
-
 typedef uint32_t Status;
 #endif // __cplusplus
-
 
 typedef struct {
   int32_t a;
@@ -43,15 +41,11 @@ typedef int32_t Unit;
 typedef Status SpecialStatus;
 
 #ifdef __cplusplus
-
 extern "C" {
-
 #endif // __cplusplus
 
 void root(IntFoo x, DoubleFoo y, Unit z, SpecialStatus w);
 
 #ifdef __cplusplus
-
 } // extern "C"
-
 #endif // __cplusplus
--- a/tests/expectations/alias.compat.c
+++ b/tests/expectations/alias.compat.c
@@ -1,0 +1,57 @@
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+enum Status
+#ifdef __cplusplus
+  : uint32_t
+#endif // __cplusplus
+ {
+  Ok,
+  Err,
+};
+#ifndef __cplusplus
+
+typedef uint32_t Status;
+#endif // __cplusplus
+
+
+typedef struct {
+  int32_t a;
+  float b;
+} Dep;
+
+typedef struct {
+  int32_t a;
+  int32_t b;
+  Dep c;
+} Foo_i32;
+
+typedef Foo_i32 IntFoo;
+
+typedef struct {
+  double a;
+  double b;
+  Dep c;
+} Foo_f64;
+
+typedef Foo_f64 DoubleFoo;
+
+typedef int32_t Unit;
+
+typedef Status SpecialStatus;
+
+#ifdef __cplusplus
+
+extern "C" {
+
+#endif // __cplusplus
+
+void root(IntFoo x, DoubleFoo y, Unit z, SpecialStatus w);
+
+#ifdef __cplusplus
+
+} // extern "C"
+
+#endif // __cplusplus
--- a/tests/expectations/no_includes.cpp
+++ b/tests/expectations/no_includes.cpp
@@ -1,3 +1,4 @@
+
 extern "C" {
 
 void root();
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -10,6 +10,7 @@ fn run_cbindgen(
     path: &Path,
     output: &Path,
     language: Language,
+    cpp_compat: bool,
     style: Option<Style>,
 ) {
     let program = Path::new(cbindgen_path);
@@ -18,6 +19,10 @@ fn run_cbindgen(
         Language::Cxx => {}
         Language::C => {
             command.arg("--lang").arg("c");
+
+            if cpp_compat {
+                command.arg("--cpp-compat");
+            }
         }
     }
 
@@ -86,6 +91,7 @@ fn run_compile_test(
     name: &'static str,
     path: &Path,
     language: Language,
+    cpp_compat: bool,
     style: Option<Style>,
 ) {
     let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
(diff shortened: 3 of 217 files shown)

### CODE REVIEW

**Comment by emilio** [comment]
So main question is: What's the use case for this?

As in, how's this better than generating two files, and doing something like:

```
#ifdef __cplusplus
#include "cbindgen-cpp.h"
#else 
#include "cbindgen-c.h"
#endif 
```

Regarding the typedefs and the `_t` suffix, if I understand correctly that's to make cases like this:

```c
enum Status {
  Ok,
  Err,
};
typedef uint32_t Status;
```

Compile in C++, right?

Would it be better to generate instead:

```
enum Status
#ifdef __cplusplus
  : uint32_t
#endif 
{
  Ok,
  Err,
};

#ifndef __cplusplus 
typedef uint32_t Status;
#endif 
```

?

That could be done even without introducing a third mode, same for the `#ifdef __cplusplus` blocks around `extern "C"`, those could just be generated by default for `lang = C` IMHO.

Wodann:
@emilio Yes, the current C implementation cannot be compiled by a C++ compiler due to renaming of types.

In theory you can generate separate C and C++ headers and combine in a third header file uisng an `#ifdef` statement, but that can quickly grow out of hand as you are now generating 3 files for each generated set of bindings. As such I thought it beneficial to add generation of headers that can be used in both C and C++ - as is.

If you are okay incorporating it in the C generation, I have no objections. I opted to add a new `Both` generation mode, as it leaves people the option to add dedicated C, dedicated C++ or a slightly less verbose combined header. It's up to you :)

I actually really like your proposal to use the `#ifdef` and `#ifndef` statements; and will rewrite my solution to use those.

That leaves the question whether you want me to integrate it into the C implementation or not?

emilio:
> If you are okay incorporating it in the C generation, I have no objections. I opted to add a new Both generation mode, as it leaves people the option to add dedicated C, dedicated C++ or a slightly less verbose combined header. It's up to you :)

I think it's less complexity explosion over-all, so I tend to prefer just changing the C code generation, yeah.

If somebody complains about the `#ifdef`s, I think an option to disable them (`disable_cpp_compat` or something) would still be easier than a third mode, but the ifdefs do what people are most likely to want (let that C header be included everywhere). So I think they are a good default, and it's less complex than a third mode.

WDYT @eqrion?

Wodann:
@emilio I was waiting on any feedback from @eqrion before I started implementing the fixes, but shall I just move ahead with your proposed solution for the time being?

emilio:
Yeah, I think it should not be very controversial, so feel free to go ahead :)

Wodann:
@emilio I didn't want to potentially disrupt previous discussions by force pushing a rebased history. For the time being I reverted the earlier changes, and added new commits that implement the discussed approach. Once you are happy with the implementation, I will overwrite the current branch with a rebased history, so a clean commit history can be merged in.

I added the `cpp_compat` config option so users can opt-in to C++ compatibility for C bindings. I thought it best to disable this feature by default, as I didn't want to break backwards compatibility for existing users.

I also had a couple of minor improvement that I added in the same PR:
- I added the `cbindgen_output`'s `stderr` log in the reported error message, when the bindings generation fails in a test case.
- It was already possible to select a C compiler using the `CC` environment variable. I also added the option to select a C++ compiler using the `CXX` environment variable.

tests/tests.rs
```diff
@@ -106,25 +112,35 @@ fn run_compile_test(
...
+    if language == Language::C && cpp_compat {
+        compile(&output, Language::Cxx)
+    }
 }
 
 fn test_file(profile: &'static str, name: &'static str, filename: &'static str) {
     let test = Path::new(filename);
     for style in &[Style::Type, Style::Tag, Style::Both] {
-        run_compile_test(profile, name, &test, Language::C, Some(*style));
+        run_compile_test(profile, name, &test, Language::C, true, Some(*style));
```
emilio: Maybe worth making the new bool argument an `enum`, so it's clearer what it means from here.

Otherwise at least `/* cpp_compat = */ true`?

tests/expectations/alias.compat.c
```diff
@@ -0,0 +1,57 @@
...
+enum Status
+#ifdef __cplusplus
+  : uint32_t
+#endif // __cplusplus
+ {
+  Ok,
+  Err,
+};
+#ifndef __cplusplus
+
```
emilio: Spacing here in the generated code is a bit weird, may be worth fixing it if it doesn't take much effort, but maybe worth doing it in a separate PR.

tests/expectations/alias.compat.c
```diff
@@ -0,0 +1,57 @@
...
+typedef uint32_t Status;
+#endif // __cplusplus
+
```
emilio: Similarly it seems that this newline is a bit excessive.

tests/tests.rs
Wodann: I decided to add a second for loop: `for cpp_compat in &[true, false]`, to make it clear that we are iterating over the `cpp_compat` option.

I thought that to be a better alternative than introducing an enum, as an enum would mean that the entire codebase becomes riddled with `if config.cpp_compat == Compat::On` statements, as opposed to a cleaner boolean check.

tests/expectations/alias.compat.c
Wodann: I removed the additional newlines, to make it more compact. It did change one edge case in the `no_includes.cpp` file, that now starts with an empty newline.

```rust
        if !self.functions.is_empty() || !self.globals.is_empty() {
            if self.config.language == Language::C && self.config.cpp_compat {
                out.new_line_if_not_start();
                out.write("#ifdef __cplusplus");
            }

            if self.config.language == Language::Cxx || self.config.cpp_compat {
                out.new_line(); // Used to be out.new_line_if_not_start(), but would cause a double newline when creating C bindings with cpp_compat turned on
                out.write("extern \"C\" {");
                out.new_line();
            }

            if self.config.language == Language::C && self.config.cpp_compat {
                out.write("#endif // __cplusplus");
                out.new_line();
            }
```

Wodann:
Removed the initial attempt from the history and forced pushed.

tests/tests.rs
emilio: Oh I totally mean to keep the enum scoped to the tests.rs file, so the only place that would have to do that is the place that adds the argument. But sure, separate loop looks good too

tests/expectations/no_includes.cpp
```diff
@@ -1,3 +1,4 @@
+
```
emilio: I can live with this newline :)

emilio:
Thanks so much for the patch @Wodann!

Wodann:
You're very welcome. Happy to contribute :)

Thank you for your guidance and responsiveness @emilio !

PR #355, opened 2019-06-10
Fix the crate parsing order

### LINKED ISSUE #337: cbindgen 0.8.6 stopped producing opaque structs
Tested with the current https://github.com/lu-zero/crav1e

All the structures in `src/lib.rs` aren't emitted anymore.

Everything works in `0.8.4`.

lu-zero:
Since cbindgen ignores the ident namespaces and the expected behavior is
to have it generate bindings for the idents declared in the root crate,
parse first the root crate and then its dependencies.

Fixes #337.

### DIFF
--- a/src/bindgen/parser.rs
+++ b/src/bindgen/parser.rs
@@ -156,9 +156,27 @@ impl<'a> Parser<'a> {
         assert!(self.lib.is_some());
         self.parsed_crates.insert(pkg.name.clone());
 
-        // Before we do anything with this crate, we must first parse all of its dependencies.
-        // This is guaranteed to terminate because the crate-graph is acyclic (and even if it
-        // wasn't, we've already marked the crate as parsed in the line above).
+        // Check if we should use cargo expand for this crate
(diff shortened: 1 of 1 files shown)


src/bindgen/parser.rs
```diff
@@ -156,6 +156,27 @@ impl<'a> Parser<'a> {
...
+            None => {
+                // This should be an error, but is common enough to just elicit a warning
+                warn!(
+                    "Parsing crate `{}`: can't find lib.rs with `cargo metadata`.",
+                    pkg.name
+                );
+            }
+        }
+
         // Before we do anything with this crate, we must first parse all of its dependencies.
```
emilio: What makes this not be needed then? If it's not really needed, then the comment should go away.

lu-zero: good catch!

emilio:
r? @Gankro (since you introduced that comment?)

I'd need to audit how stateful our parser is. I suspect most of the cross-referencing stuff happens after we're done with all the crates so this could work, but not sure.

PR #361, opened 2019-06-12
Add a docs/ folder and move existing documentation there

eqrion:
We actually have enough random doc files that it probably makes sense to consolidate it to a folder.

@Gankro is this okay, or will it break external references to the user guide?


Gankro:
It would break the currently very active link in this tweet: https://twitter.com/Gankro/status/1138588645768622080

About 150 people have clicked on that link, many more have seen it.

I think this is fine/reasonable after a grace period of a few weeks..?

eqrion:
Yeah that totally makes sense. I'll delay this for a couple weeks.

emilio:
Should we do this now?

Gankro:
yep, let's go for it

PR #362, opened 2019-06-17
Add an option to generate consts / globals / fn from parsed dependencies.

emilio:
This is a limitation that may not be useful if you're using it to parse known
dependencies.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -751,6 +751,12 @@ exclude = ["libc"]
 # default: false
 clean = false
 
+# Whether to generate top-level constants / globals and functions defined in
+# dependent crates.
+#
+# default: false
+top_level_items_outside_of_binding_crate = true
+
 [parse.expand]
 # A list of crate names that should be run through `cargo expand` before
 # parsing to expand any macros. Note that if a crate is named here, it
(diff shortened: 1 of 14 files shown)


lu-zero:
I can give it a go with [crav1e](https://github.com/lu-zero/crav1e) tomorrow since it looks like it would be impacted by this.

emilio:
This is opt-in so it shouldn't be a breaking change right?

lu-zero:
I confirm it does not break anything :)

emilio:
Turned into a whitelist per discussion with @Gankro. Does it look good now?

docs.md
```diff
@@ -751,6 +751,12 @@ exclude = ["libc"]
 # default: false
 clean = false
 
+# Which crates other than the top-level binding crate we should generate
+# top-level constants / globals and functions.
```
Gankro:
```suggestion
# bindings for.
```

template.toml
```diff
@@ -127,6 +127,7 @@ parse_deps = false
 # include = []
 exclude = []
 clean = false
+# deps_with_top_level_items = []
```
Gankro:
```suggestion
deps_with_top_level_items = []
```

I believe this should be uncommented since the default is really the same as `[]`

emilio:
Renamed to extra_bindings per IRC discussion yesterday.

PR #363, opened 2019-06-24
Making clap an optional feature

elichai:
This saves around 10 dependencies for people who uses `cbindgen` in the build.rs only (so they don't need clap)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ repository = "https://github.com/eqrion/cbindgen/"
 travis-ci = { repository = "eqrion/cbindgen" }
 
 [dependencies]
-clap = "2"
+clap = { version = "2", optional = true }
 log = "0.4"
 serde = {version = "1.0", features = ["derive"]}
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -14,11 +14,10 @@ repository = "https://github.com/eqrion/cbindgen/"
 travis-ci = { repository = "eqrion/cbindgen" }
 
 [dependencies]
-clap = "2"
+clap = { version = "2", optional = true }
 log = "0.4"
-serde = {version = "1.0", features = ["derive"]}
+serde = { version = "1.0", default-features = false, features = ["derive"]}
```
emilio: Should this be part of this commit? Or should it be a different commit?

elichai: `serde-derive` is redundant if you use the `derive` feature, and the default features aren't needed here. I can rebase it to a different commit.

elichai: Rebased.

emilio:
Thanks, looks great :)

PR #370, opened 2019-07-23
Preserve empty lines in doc comments.

### LINKED ISSUE #369: Preserve empty lines in comments.
I just updated an older project and realized that the changes introduced in 9248f84d0cc1f87b849262638ec277e030476446 started to skip empty lines in doc comments. As we normally use them to structure longer comments into multiple sections the resulting header is now much harder to read. 

Is their any configuration I might miss to reintroduce the old behavior and preserve them?

0ndorio:
This required two changes in parallel.

As a first and obvious step it removes a check that skipped empty
lines during documentation loading, but it was also necessary to
refactor the `get_comment_lines()` utility method as a second step.

This second refactoring was necessary as each line in a doc comment
is trimmed and transformed in its own doc attribute. The attribute
of an empty line therefore contains an empty string as value. If we
now call `lines()` on such an empty string we end up with an empty
iterator as the method is internally configured to ignore trailing
empty lines.

fixes #369

--- 

Next to this I am unsure about the whole trimming logic inside `get_comment_lines()`. I can provide a second pull request to remove it, if wished. The reasoning would be that rust already removes the leading `///` on each line. It is therefore not possible to use the trimmed symbols at the start of the actual line.  For example the rust-doc bold `**my_parameter**` is trimmed into `my_parameter**` inside the header file which looks at least strange.

### DIFF
--- a/src/bindgen/ir/documentation.rs
+++ b/src/bindgen/ir/documentation.rs
@@ -20,7 +20,7 @@ impl Documentation {
         let doc = attrs
             .get_comment_lines()
             .into_iter()
-            .filter(|x| !x.is_empty() && !x.starts_with("cbindgen:"))
+            .filter(|x| !x.starts_with("cbindgen:"))
             .collect();
 
         Documentation { doc_comment: doc }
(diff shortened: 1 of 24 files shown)


0ndorio:
Thanks :) @emilio. Any opinion regarding the trim logic in `get_comment_lines()`? Should I open a new issue for it?

emilio:
I think it'd be ok to remove, yeah, please do file an issue / PR, thanks!

Issue #347, opened 2019-05-23
cbindgen panicks when parse-expand is run inside of buildscript due to inability to acquire semaphore


Wodann:
I have successfully run `cbingen` on the command-line using the following `cbindgen.toml`:

```toml
include_guard = "SE4_CORE_BINDINGS_H_"
include_version = true
braces = "SameLine"
line_length = 100
tab_width = 4

[export]
prefix = "SE4"

[parse]
parse_deps = false

[parse.expand]
crates = ["se4_capi"]
```
I decided to add it as a build script:

```rust
extern crate cbindgen;

use std::env;

fn main() {
    let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();

    cbindgen::generate(crate_dir)
        .expect("Unable to generate bindings")
        .write_to_file("include/bindings.h");
}
```

Upon execution, of `cargo build` cbindgen instantly crashes. It seems like `cbindgen` tries to acquire a lock on the build directory that was previously acquired by `cargo`, and thus crashes after outputing the following:

```
    Blocking waiting for file lock on build directory
   Compiling se4_capi v0.1.0
error: failed to run custom build command for `se4_capi v0.1.0`

Caused by:
  process didn't exit successfully: `build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'Unable to generate bindings: CargoExpand("se4_capi", Compile("    Blocking waiting for file lock on build directory\nerror: Too many posts were made to a semaphore. (os error 298)\n"))', src\libcore\result.rs:999:5
stack backtrace:
   0: std::sys::windows::backtrace::set_frames
             at /rustc/37ff5d388f8c004ca248adb635f1cc84d347eda0\/src\libstd\sys\windows\backtrace\mod.rs:95
   1: std::sys::windows::backtrace::unwind_backtrace
             at /rustc/37ff5d388f8c004ca248adb635f1cc84d347eda0\/src\libstd\sys\windows\backtrace\mod.rs:82
   2: std::sys_common::backtrace::_print
             at /rustc/37ff5d388f8c004ca248adb635f1cc84d347eda0\/src\libstd\sys_common\backtrace.rs:71
   3: std::sys_common::backtrace::print
             at /rustc/37ff5d388f8c004ca248adb635f1cc84d347eda0\/src\libstd\sys_common\backtrace.rs:59
   4: std::panicking::default_hook::{{closure}}
[...]

PR #371, opened 2019-08-02
Make running cbindgen with own create expansion from build.rs

vmx:
With the current version of cbindgen it's not possible to have macros in the crate cbindgen is triggered from if you run cbindgen programmatically via build.rs. This issue was reported as #347. This PR fixes that problem.

I've create two commits as it is kind of two issues that need a fix.

### Commit 1: Run cargo expand outside the normal target directory 

When crates should get expanded, then cargo is run again from within cbindgen.
When cbindgen is started from a build.rs file, this means that cargo is starting
another cargo run.

Cargo locks the directory it is running on. If two cargos spawn by each other
run with the same target directory, then they both want to accquire a lock and
hence deadlock.

This commit fixes the problem with running the spawned cargo at a different
directory. Please note that this will always be the same directory, hence
any subsequent runs will be faster (as you would exepct it to be).

### Commit 2:  Don't call cbindgen recursively

To expand a crate, cbindgen calls cargo on that crate. cbindgen can be called from a build.rs
file. But if cargo is called again, it will also call cbindgen again and hence end up in an
endless recursion.

This commit makes sure that cbindgen isn't called again if it is already running.

You can verify this fix with this minimal example https://github.com/vmx/cbindgen-expand-bug

### DIFF
--- a/src/bindgen/cargo/cargo_expand.rs
+++ b/src/bindgen/cargo/cargo_expand.rs
@@ -83,6 +83,10 @@ pub fn expand(
         cmd.env("CARGO_TARGET_DIR", PathBuf::from(path).join("expanded"));
     }
 
+    // Set this variable so that we don't call it recursively if we expand a crate that is using
+    // cbindgen
+    cmd.env("_CBINDGEN_IS_RUNNING", "1");
+
     cmd.arg("rustc");
     cmd.arg("--lib");
     cmd.arg("--manifest-path");
--- a/src/bindgen/cargo/cargo_expand.rs
+++ b/src/bindgen/cargo/cargo_expand.rs
@@ -6,7 +6,7 @@ use std::env;
 use std::error;
 use std::fmt;
 use std::io;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::str::{from_utf8, Utf8Error};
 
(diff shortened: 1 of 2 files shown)


vmx:
The build failure seems to be env related and not related to this change. May someone please restart the build?

src/bindgen/cargo/cargo_expand.rs
```diff
@@ -82,6 +82,10 @@ pub fn expand(
        cmd.env("CARGO_TARGET_DIR", out_dir.join("expanded"));
     }
 
+    // Set this variable so that we don't call it recursively if we expand a crate that is using
+    // cbindgen
+    cmd.env("_CBINDGEN_IS_RUNNING", "1");
```
emilio: This is a bit hacky... Though I don't have better ideas.

src/bindgen/cargo/cargo_expand.rs
```diff
@@ -75,6 +75,11 @@ pub fn expand(
         cmd.env("CARGO_TARGET_DIR", _temp_dir.unwrap().path());
     } else if let Ok(ref path) = env::var("CARGO_EXPAND_TARGET_DIR") {
         cmd.env("CARGO_TARGET_DIR", path);
+    } else {
+       // Don't use the default dir, else we could end up in a deadlock when cbindgen is run from
+       // a build.rs file
+       let out_dir = PathBuf::from(&env::var("OUT_DIR").unwrap());
```
emilio: Doesn't this break CLI usage? There's no guarantee that `OUT_DIR` exists when called from the CLI.

Can we instead check the return value of `env::var` and only set it if `OUT_DIR` exists?

Or am I missing something?

src/bindgen/cargo/cargo_expand.rs
vmx: Yes, it's super hacky, but it's also the best idea I had.

src/bindgen/cargo/cargo_expand.rs
vmx: I haven't thought about CLI use, you are right. I'll add this check.

vmx:
The build failure seems again env related. I ran `cargo test` locally and also verified that `cbindgen` running as CLI also still works (with the earlier version of this PR it really didn't).

emilio:
Yeah, this looks good, not sure what's up with Travis...

emilio:
Ok, I retriggered travis and seems much happier now.

Issue #374, opened 2019-08-07
Avoid trimming non white space symbols in comments.


0ndorio:
The current implementation of `get_comment_lines()` in `utilities.rs` removes all kind of special symbols from the start of each line in a comment but rustc already removes any leading `/` symbols. 

I think the original intention was to avoid problems with miss interpreted `/*` or `*/` marker but as it only removes them from the start of the line but nowhere else this isn't the case in here and could only be prevented with they are removed everywhere.

Any additional trimming of leading non white space symbols should therefore not be necessary and it prevents the usage of bold markdown formatting for the first word in a line or the usage of `/` to divide a comment in multiple sections.

Issue #184, opened 2018-08-01
Preserve indentation on comments


dbrgn:
When switching from cbindgen 0.4.x to 0.6.1, the following change was applied to the generated headerfile:

```diff
--- a/ffi/saltyrtc_client_ffi.h
+++ b/ffi/saltyrtc_client_ffi.h
@@ -44,10 +44,10 @@ void salty_event_loop_free_remote(salty_remote_t *ptr);
  * Return a remote handle from an event loop instance.
  *
  * Thread safety:
- *     The `salty_remote_t` instance may be used from any thread.
+ * The `salty_remote_t` instance may be used from any thread.
  * Returns:
- *     A reference to the remote handle.
- *     If the pointer passed in is `null`, an error is logged and `null` is returned.
+ * A reference to the remote handle.
+ * If the pointer passed in is `null`, an error is logged and `null` is returned.
  */
 salty_remote_t *salty_event_loop_get_remote(salty_event_loop_t *ptr);
 
@@ -57,9 +57,9 @@ salty_remote_t *salty_event_loop_get_remote(salty_event_loop_t *ptr);
  * In the background, this will instantiate a Tokio reactor core.
  *
  * Returns:
- *     Either a pointer to the reactor core, or `null`
- *     if creation of the event loop failed.
- *     In the case of a failure, the error will be logged.
+ * Either a pointer to the reactor core, or `null`
+ * if creation of the event loop failed.
+ * In the case of a failure, the error will be logged.
  */
 salty_event_loop_t *salty_event_loop_new(void);
```

Would it be possible to preserve indentation in comments?

eqrion:
Yes it should be. The relevant code is [here](https://github.com/eqrion/cbindgen/blob/master/src/bindgen/ir/documentation.rs#L43).

dbrgn:
If it should be, wouldn't this be a bug instead of an enhancement? :slightly_smiling_face: Or did I misunderstand you?

eqrion:
It's fuzzy whether this was originally intended, but I agree that bug probably makes more sense.

PR #375, opened 2019-08-07
Avoid most trimming and preserve comment structure.

0ndorio:
This fixes two additional documentation related issues: #184 & #374. Both effect nearly the same parts of the source code. It also updates the `documentation` test to verify the new expected behavior.

---

Feel free to nitpick. I will try to come up with all required changes to get this merged.

### DIFF
--- a/src/bindgen/ir/annotation.rs
+++ b/src/bindgen/ir/annotation.rs
@@ -50,10 +50,17 @@ impl AnnotationSet {
     }
 
     pub fn load(attrs: &[syn::Attribute]) -> Result<AnnotationSet, String> {
-        let lines: Vec<String> = attrs
-            .get_comment_lines()
-            .into_iter()
-            .filter(|x| !x.is_empty() && x.starts_with("cbindgen:"))
+        let lines = attrs.get_comment_lines();
+        let lines: Vec<&str> = lines
+            .iter()
+            .filter_map(|line| {
+                let line = line.trim_start();
+                if !line.starts_with("cbindgen:") {
+                    return None;
+                }
+
+                Some(line)
+            })
             .collect();
 
         let must_use = attrs.has_attr_word("must_use");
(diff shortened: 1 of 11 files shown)


src/bindgen/ir/annotation.rs
```diff
@@ -52,8 +52,15 @@ impl AnnotationSet {
     pub fn load(attrs: &[syn::Attribute]) -> Result<AnnotationSet, String> {
         let lines: Vec<String> = attrs
             .get_comment_lines()
-            .into_iter()
-            .filter(|x| !x.is_empty() && x.starts_with("cbindgen:"))
+            .iter()
+            .filter_map(|line| {
+                let line = line.trim_start();
+                if line.is_empty() || !line.starts_with("cbindgen:") {
```
emilio: I think we could avoid a bunch of extra allocations here...

Any reason we can't do:

```
let lines = attrs.get_comment_lines();
let lines: Vec<&str> = lines.iter().filter_map(..).collect();
```

Also, the `is_empty()` check looks redundant to me (was redundant already, but I can't see how an empty line can start with `"cbindgen:"` ;))

src/bindgen/ir/annotation.rs
```diff
@@ -62,12 +69,7 @@ impl AnnotationSet {
 
         // Look at each line for an annotation
         for line in lines {
```
emilio: `debug_assert!(line.starts_with("cbindgen:"))` maybe?

0ndorio: Sounds like a good idea.

src/bindgen/ir/annotation.rs
0ndorio: Absolutely no reason! Going to change this asap.

0ndorio: Included via rebase.

src/bindgen/ir/annotation.rs
0ndorio: Included via rebase

PR #377, opened 2019-08-09
ir: Add an option to generate an empty, default constructor for tagged enums.

emilio:
This allows to clean up a pattern that has been showing up lately, see
occurrences of:

  https://searchfox.org/mozilla-central/rev/9775cca0a10a9b5c5f4e15c8f7b3eff5bf91bbd0/servo/ports/geckolib/cbindgen.toml#329

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -270,6 +270,7 @@ The rest are just local overrides for the same options found in the cbindgen.tom
 * derive-tagged-enum-destructor
 * derive-tagged-enum-copy-constructor
 * prefix-with-name
+* private-default-tagged-enum-constructor
 
 
 
@@ -702,6 +703,14 @@ derive_tagged_enum_destructor = false
 #
 # default: false
 derive_tagged_enum_copy_constructor = false
+
+# Whether enums with fields should generate an empty, private destructor.
+# This allows the auto-generated constructor functions to compile, if there are
+# non-trivially constructible members. This falls in the same family of
+# dangerousness as `derive_tagged_enum_copy_constructor` and co.
+#
+# default: false
+private_default_tagged_enum_constructor = false
 
 
 
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -400,6 +400,9 @@ pub struct EnumConfig {
     pub derive_tagged_enum_destructor: bool,
     /// Whether to generate copy-constructors of tagged enums.
     pub derive_tagged_enum_copy_constructor: bool,
+    /// Whether to generate empty, private default-constructors for tagged
+    /// enums.
+    pub private_default_tagged_enum_constructor: bool,
 }
 
 impl EnumConfig {
@@ -438,6 +441,15 @@ impl EnumConfig {
             return x;
         }
         self.derive_tagged_enum_copy_constructor
+    }
+    pub(crate) fn private_default_tagged_enum_constructor(
+        &self,
+        annotations: &AnnotationSet,
+    ) -> bool {
+        if let Some(x) = annotations.bool("private-default-tagged-enum-constructor") {
+            return x;
+        }
+        self.private_default_tagged_enum_constructor
     }
 }
 
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -895,6 +895,22 @@ impl Source for Enum {
             if config.language == Language::Cxx
                 && config
                     .enumeration
+                    .private_default_tagged_enum_constructor(&self.annotations)
+            {
+                out.new_line();
+                out.new_line();
+                write!(out, "private:");
+                out.new_line();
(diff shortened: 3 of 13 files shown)


docs.md
```diff
@@ -703,6 +703,14 @@ derive_tagged_enum_destructor = false
...
 derive_tagged_enum_copy_constructor = false
 
+# Whether enums with fields should generate an empty, private destructor.
+# This allows the auto-generated constructor functions to compile, if there are
+# non-trivially constructible members. This falls in the same family of
+# dangerousness as `derive_tagged_enum_copy_constructor` and co.
+#
+# default: false
+private_default_tagged_enum_constructor = false
+
```
Gankro: please also add it to the template

src/bindgen/config.rs
```diff
@@ -441,6 +444,12 @@ impl EnumConfig {
         }
         self.derive_tagged_enum_copy_constructor
     }
+    pub(crate) fn private_default_tagged_enum_constructor(&self, annotations: &AnnotationSet) -> bool {
+        if let Some(x) = annotations.bool("private-default-tagged-enum-constructor") {
+            return x;
+        }
```
Gankro: please add this to the list of enum annotations:

https://github.com/eqrion/cbindgen/blob/master/docs.md#enum-annotations

docs.md
```diff
@@ -703,6 +703,14 @@ derive_tagged_enum_destructor = false
 # default: false
 derive_tagged_enum_copy_constructor = false
 
+# Whether enums with fields should generate an empty, private destructor.
+# This allows the auto-generated constructor functions to compile, if there are
+# non-trivially constructible members. This falls in the same family of
+# dangerousness as `derive_tagged_enum_copy_constructor` and co.
```
Gankro: i'm completely unfamiliar with the issue and having trouble making sense of it, can you expand on this? what types don't we support? i'm not familiar with the concept of "trivially constructible".

emilio: So the issue is the following. Simplified.

This builds, as expected:

```c++
#include <cstdint>
#include <new>

struct TypeWithDestructor {
    uint8_t field;
    // ~TypeWithDestructor() {}
};

struct TaggedUnion {
    static TaggedUnion Create(const TypeWithDestructor& a) {
        TaggedUnion ret;
        ret.type = 0;
        new (&ret.foo) TypeWithDestructor(a);
        return ret;
    }

    uint8_t type;
    union {
        TypeWithDestructor foo;
    };
};
```

If you uncomment the destructor, then it [doesn't](https://godbolt.org/z/NekDlQ), because C++ will implicitly delete the default constructor of `TaggedUnion`.

Since all we want is a default constructor that doesn't initialize anything, so that `Create()` does the right thing, a private default constructor serves this purpose. It also avoids people accidentally creating uninitialized structs themselves, which is a win-win.

emilio: The final simplified example for cbindgen would be:

```c++
#include <cstdint>
#include <new>

struct TypeWithDestructor {
    uint8_t field;
    ~TypeWithDestructor() {}
};

struct TaggedUnion {
    static TaggedUnion Create(const TypeWithDestructor& a) {
        TaggedUnion ret;
        ret.type = 0;
        new (&ret.foo) TypeWithDestructor(a);
        return ret;
    }

    private:
    TaggedUnion() {}
    
    public:
    ~TaggedUnion() { /* cbindgen-generated */ }

    uint8_t type;
    union {
        TypeWithDestructor foo;
    };
};
```

With nicer enum types and such of course.

Gankro: Is the claim that this is dangerous just incorrect? Adding a private default constructor doesn't actually change the ABI in any meaningful way, does it? It just enables you to add destructors that *do* mess up the ABI..?

docs.md
emilio: Done

src/bindgen/config.rs
emilio: Done.

docs.md
emilio: Adding constructors changes the ABI of return types in MSVC: https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention?view=vs-2019#return-values

> To return a user-defined type by value in RAX, it must have a length of 1, 2, 4, 8, 16, 32, or 64 bits. It must also have no user-defined constructor, destructor, or copy assignment operator; no private or protected non-static data members; no non-static data members of reference type; no base classes; no virtual functions; and no data members that do not also meet these requirements. (This is essentially the definition of a C++03 POD type. Because the definition has changed in the C++11 standard, we don't recommend using std::is_pod for this test.)

emilio: And yeah I wish I didn't know this :)

emilio:
@Gankra anything blocking this? Thanks!

Gankra:
FRIG

i swear i keep hallucinating that i said yes to this

pls ship it

PR #387, opened 2019-08-23
Add support for adding "using namespace" statements

Pajn:


Pajn:
The Rust crate both consumes a C++ library and is consumed by a C++ binary.

The Rust crates export functions uses C++ types (which are located in namespaces) as parameters and return types

When I'm using cbindgen it only generates the typename without namespace (because how could it otherwise?) which causes the build of C++ binary to fail because it can't find these types.

The library I'm using is full of `using namespace` in its headers so it doesn't seem that crazy to me.

src/bindgen/config.rs
```diff
@@ -609,6 +609,8 @@ pub struct Config {
     pub namespace: Option<String>,
     /// An optional list of namespaces. Only applicable when language="C++"
     pub namespaces: Option<Vec<String>>,
+    /// An optional list of namespaces to ddeclare as using. Only applicable when language="C++"
+    pub using: Option<Vec<String>>,
```
emilio: If not, can we at least call this `using_namespaces` or something? Otherwise the new option name makes little intuitive sense to me.

emilio:
Or is the issue that `using namespace` needs to be inside the `namespace foo {` brackets, in order not to be naked?

Pajn:
Exactly, I need it to be inside the `namespace foo {` brackets. I should have clarified that.

If you are okay with the feature as a principle I can absolutely rename it to `using_namespaces` instead.

tests/rust/using.toml
```diff
@@ -0,0 +1 @@
+using = ["std"]
```
emilio: We need a test for this + namespaces then, since it's the primary use-case.

Pajn:
>That is not an anti-pattern since it's scoped, rather than naked using namespace in a header which is a mess

That's good to know. I don't know C++ and have only followed the patterns of the library I'm using. Do you want some kind of error or warning when using `using_namespaces without `namespace` or `namespaces`?

>I'd prefer to rename it for clarity if you don't mind

No, I agree that `using_namespaces` is clearer. I only used `using` because that what I would call it in C# which is the only place I have encountered that keyword before. I'll make sure to fix that as well as the other requested changes.

Pajn:
One thing I realized when testing with `using_namespaces` + `namespaces` is that you don't write any namespace declarations in ccp_compat. Should I avoid writing any `using namespace` in there as well?

emilio:
> One thing I realized when testing with using_namespaces + namespaces is that you don't write any namespace declarations in ccp_compat. Should I avoid writing any using namespace in there as well?

Probably, yeah.

Pajn:
Now all changes should be done. I completely replaced the old test with the one combined with namespaces as I got the impression `using_namespaces` should never be used alone.

Issue #384, opened 2019-08-17
Macro expansion breaks dependency inclusion


vmx:
There is a bug when using dependencies and macro expansion on the crate that cbindgen is run on.

This bug exposes the same issue as described at https://github.com/eqrion/cbindgen/issues/292,  but only if

```toml
[parse.expand]
crates = ["cbindgen-dep"]
```

is used in cbindgen.toml.

vmx:
I've created a minimal example for the problem: https://github.com/vmx/cbindgen-expand-dep-bug

I will probably have a look at this problem next Friday, though I might not find time and have a look on Friday 2019-09-06.

PR #389, opened 2019-09-05
fix(parser): avoid prematurely returning during expansion

schomatis:
_This may be more of a PoC than a proper solution._

In `parse_crate`, after parsing the first crate, instead of returning early if the `parse.expand` option is set, continue to inspect its dependencies (the same way it's done with a normal crate).

See diff with [`--ignore-all-space`](https://github.com/eqrion/cbindgen/pull/389/files?w=1).

Fixes https://github.com/eqrion/cbindgen/issues/384.


vmx:
I can confirm that it fixes my minimal example and also the actual code where I found this issue.

PS: You can also watch the diff without whitespace changes with this link: https://github.com/eqrion/cbindgen/pull/389/files?w=1

vmx:
@emilio I've added a test. It passes with this branch, but fails on master.

vmx:
The tests pass locally, could someone please retrigger?

emilio:
I've restarted it, though you can also do that by `git commit --amend`, then `git push --force` :)

The test looks great, thank you! I'll merge once green :)

vmx:
@emilio Sorry for the noise, I forgot to check in the most recent version of one of the `tests/expectations` files. It's green now.

PR #401, opened 2019-10-12
Initialize struct literal with list-initializer for C++11 standard

### LINKED ISSUE #397: Initialize struct literals with list-initializer
On cbindgen 0.9.1, Values of bitflags are converted to static const items with designated initializer.

But designated initializer is not available in C++11. Sadly, It is available in C++20. https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers

For C++11, struct literal should be initialized with the list-initializer.

### rust source
```rust
bitflags! {
    #[repr(C)]
    pub struct ResultFlags: u8 {
        const EMPTY = 0b00000000;
        const PROCESSED = 0b00000001;
        const TIME_UPDATED = 0b00000010;
    }
}
```

### current generated - compile error on C++11
```cpp
struct ResultFlags{
  uint8_t bits;

  explicit operator bool() const {
    return !!bits;
  }
  ResultFlags operator|(const ResultFlags& other) const {
    return {static_cast<decltype(bits)>(this->bits | other.bits)};
  }
  ResultFlags& operator|=(const ResultFlags& other) {
    *this = (*this | other);
    return *this;
  }
  ResultFlags operator&(const ResultFlags& other) const {
    return {static_cast<decltype(bits)>(this->bits & other.bits)};
  }
  ResultFlags& operator&=(const ResultFlags& other) {
    *this = (*this & other);
    return *this;
  }
};

static const ResultFlags ResultFlags_EMPTY = { .bits = 0 };
static const ResultFlags ResultFlags_PROCESSED = { .bits = 1 };
static const ResultFlags ResultFlags_TIME_UPDATED = { .bits = 2 };
```

### suggestion - compatible with C++11
```cpp
struct ResultFlags{
  uint8_t bits;

  explicit operator bool() const {
    return !!bits;
  }
  ResultFlags operator|(const ResultFlags& other) const {
    return {static_cast<decltype(bits)>(this->bits | other.bits)};
  }
  ResultFlags& operator|=(const ResultFlags& other) {
    *this = (*this | other);
    return *this;
  }
  ResultFlags operator&(const ResultFlags& other) const {
    return {static_cast<decltype(bits)>(this->bits & other.bits)};
  }
  ResultFlags& operator&=(const ResultFlags& other) {
    *this = (*this & other);
    return *this;
  }
};

static const ResultFlags ResultFlags_EMPTY = { 0 };
static const ResultFlags ResultFlags_PROCESSED = { 1 };
static const ResultFlags ResultFlags_TIME_UPDATED = { 2 };
```

Perlmint:
fix #397 

- Use list-initializer for C++ struct
- Sort struct literal elements by struct definition order

### DIFF
--- a/src/bindgen/bindings.rs
+++ b/src/bindgen/bindings.rs
@@ -2,10 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use std::cell::RefCell;
+use std::collections::HashMap;
 use std::fs;
 use std::fs::File;
 use std::io::{Read, Write};
 use std::path;
+use std::rc::Rc;
 
 use bindgen::config::{Config, Language};
 use bindgen::ir::{
@@ -19,6 +22,7 @@ pub struct Bindings {
     /// The map from path to struct, used to lookup whether a given type is a
     /// transparent struct. This is needed to generate code for constants.
     struct_map: ItemMap<Struct>,
+    struct_fileds_memo: RefCell<HashMap<BindgenPath, Rc<Vec<String>>>>,
     globals: Vec<Static>,
     constants: Vec<Constant>,
     items: Vec<ItemContainer>,
@@ -37,6 +41,7 @@ impl Bindings {
         Bindings {
             config,
             struct_map,
+            struct_fileds_memo: Default::default(),
             globals,
             constants,
             items,
@@ -57,25 +62,28 @@ impl Bindings {
         any
     }
 
-    pub fn struct_field_names(&self, path: &BindgenPath) -> Option<Vec<String>> {
-        let item_container = self.struct_map.get_items(path)?;
-        let mut fields = Vec::<String>::new();
-        for container in item_container {
-            let mut pos: usize = 0;
-            if let ItemContainer::Struct(ref st) = container {
-                for field in &st.fields {
-                    if let Some(found_pos) = fields.iter().position(|v| *v == field.0) {
-                        pos = found_pos + 1;
-                    } else {
-                        fields.insert(pos, field.0.clone());
-                        pos += 1;
-                    }
-                }
-            } else {
-                unreachable!()
-            }
+    pub fn struct_field_names(&self, path: &BindgenPath) -> Rc<Vec<String>> {
+        let mut memos = self.struct_fileds_memo.borrow_mut();
+        if let Some(memo) = memos.get(path) {
+            return memo.clone();
         }
-        Some(fields)
+
+        let mut fields = Vec::<String>::new();
+        self.struct_map.for_items(path, |st| {
+            let mut pos: usize = 0;
+            for field in &st.fields {
+                if let Some(found_pos) = fields.iter().position(|v| *v == field.0) {
+                    pos = found_pos + 1;
+                } else {
+                    fields.insert(pos, field.0.clone());
+                    pos += 1;
+                }
+            }
+        });
+
+        let fields = Rc::new(fields);
+        memos.insert(path.clone(), fields.clone());
+        fields
     }
 
     pub fn write_to_file<P: AsRef<path::Path>>(&self, path: P) -> bool {
--- a/src/bindgen/bindings.rs
+++ b/src/bindgen/bindings.rs
@@ -57,6 +57,27 @@ impl Bindings {
         any
     }
 
+    pub fn struct_field_names(&self, path: &BindgenPath) -> Option<Vec<String>> {
+        let item_container = self.struct_map.get_items(path)?;
+        let mut fields = Vec::<String>::new();
+        for container in item_container {
+            let mut pos: usize = 0;
+            if let ItemContainer::Struct(ref st) = container {
+                for field in &st.fields {
+                    if let Some(found_pos) = fields.iter().position(|v| *v == field.0) {
+                        pos = found_pos + 1;
+                    } else {
+                        fields.insert(pos, field.0.clone());
+                        pos += 1;
+                    }
+                }
+            } else {
+                unreachable!()
+            }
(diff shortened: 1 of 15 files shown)


src/bindgen/ir/constant.rs
```diff
@@ -23,6 +23,10 @@ use syn::UnOp;
 #[derive(Debug, Clone)]
 pub enum Literal {
     Expr(String),
+    PostfixUnaryOp {
+        op: &'static str,
+        value: Box<Literal>,
```
emilio: Could you elaborate on the reasoning behind this change? it seems this could be a separate commit.

src/bindgen/bindings.rs
```diff
@@ -57,6 +57,27 @@ impl Bindings {
...
     }
 
+    pub fn struct_field_names(&self, path: &BindgenPath) -> Option<Vec<String>> {
+        let item_container = self.struct_map.get_items(path)?;
+        let mut fields = Vec::<String>::new();
+        for container in item_container {
+            let mut pos: usize = 0;
+            if let ItemContainer::Struct(ref st) = container {
+                for field in &st.fields {
+                    if let Some(found_pos) = fields.iter().position(|v| *v == field.0) {
```
emilio: This seems it could get pretty slow... I'll try to do some performance measurement...

Off hand, if the only use-case for this so far is bitflags, I'd prefer to just special-case the code generation for single-field structs, which would be much simpler of a patch.

emilio: Alternatively maybe this should be memoized / cached somehow. This does a bunch of work for each constant of a given type, specially if there are a lot of fields.

src/bindgen/ir/constant.rs
Perlmint: Replacing `fmt::Display` implements with implementing `write` method occur error to make `Expr` Literal by stringifying unary expression. So, I introduced a new `PostfixUnqaryOp`.
I thought it is following changes by implementing the `write` method. Is it better making a separate commit?

src/bindgen/bindings.rs
Perlmint: I found this bug during use bitflags. But, already there are other test cases. And, the actual reason is MSVC doesn't support C11 and designated initializer is not a part of C++11 ([It will be added to C++2a](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0329r4.pdf)). IMO, Using list-initializer for every C++ case can resolve this error clearly.

As your suggestion, I'll add some memoizing codes.

Perlmint:
Add memoization to `Bindings::struct_field_names`

src/bindgen/bindings.rs
```diff
@@ -57,25 +62,32 @@ impl Bindings {
...
                     }
+                } else {
+                    unreachable!()
                 }
-            } else {
-                unreachable!()
             }
+
+            memos.insert(path.clone(), Rc::new(fields));
+            memos.get(path).map(|val| val.clone())
```
emilio: This always returns `Some()` right?

This should probably do something like:

```rust
let fields = Rc::new(fields);
memos.insert(path.clone(), fields.clone());
fields
```

And just return `Rc<Vec<..>>`.

src/bindgen/ir/constant.rs
```diff
@@ -174,34 +155,90 @@ impl Literal {
...
+                    write!(out, "({})", export_name);
+                }
+
+                write!(out, "{{ ");
+                let mut is_first_field = true;
+                // In C++, same order as defined is required.
+                let ordered_fields = out
+                    .bindings()
+                    .struct_field_names(path)
+                    .expect("Failed to find struct")
```
emilio: There's no need to clone() again, nor `expect()` if `struct_field_names` never returns `None`.

tests/expectations/bitflags.cpp
```diff
@@ -27,11 +27,11 @@ struct AlignFlags {
     return *this;
   }
 };
-static const AlignFlags AlignFlags_AUTO = (AlignFlags){ .bits = 0 };
-static const AlignFlags AlignFlags_NORMAL = (AlignFlags){ .bits = 1 };
-static const AlignFlags AlignFlags_START = (AlignFlags){ .bits = 1 << 1 };
-static const AlignFlags AlignFlags_END = (AlignFlags){ .bits = 1 << 2 };
-static const AlignFlags AlignFlags_FLEX_START = (AlignFlags){ .bits = 1 << 3 };
```
emilio: Why not keeping the `(StructName)` in C++?

src/bindgen/bindings.rs
```diff
@@ -57,6 +62,34 @@ impl Bindings {
         any
     }
 
+    pub fn struct_field_names(&self, path: &BindgenPath) -> Option<Rc<Vec<String>>> {
+        let mut memos = self.struct_fileds_memo.borrow_mut();
+        if let Some(memo) = memos.get(path) {
+            Some(memo.clone())
```
emilio: nit: `return Some(...);`, then deindent the rest, please.

tests/expectations/bitflags.cpp
Perlmint: It occur error on MSVC. MSVC said it is non-standard.
> struct_literal.cpp(14): error C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax

Perlmint:
Thanks for nice comments. I fixed as your suggestion!

tests/expectations/bitflags.cpp
emilio: Does doing `AlignFlags { ... }` in C++ and `(AlignFlags) { .. }` in C work?

Perlmint: `AlignFlags{ ... }` is working on MSVC! I add new commit doing this.

Issue #391, opened 2019-09-27
cpp_compat option ignores namespace


adamski:
The following config

```toml
language = "C"
cpp_compat = true
include_version = false
namespace = "acme"
```

Results in no namespace being included in the generated header within the ` #ifdef __cplusplus ` sections. 

I would expect something like this to be generated:

```cpp
#ifdef __cplusplus
namespace acme {
extern "C" {
#endif // __cplusplus

void print_hello(void);

#ifdef __cplusplus
} // extern "C"
} // namespace acme
#endif // __cplusplus
```

emilio:
I don't see related discussion in https://github.com/eqrion/cbindgen/pull/349, so thus no reason not to do this I guess. This is probably pretty easy to fix.

PR #407, opened 2019-10-25
Surround namespace with __cplusplus ifdef in cpp_compat mode

adamski:
This implements #391 

Ideally I think some of the outputs should be refactored into seperate functions e.g. `open_cpp_compat` and `close_cpp_compat` to keep things DRY.

### DIFF
--- a/src/bindgen/bindings.rs
+++ b/src/bindgen/bindings.rs
@@ -167,7 +167,7 @@ impl Bindings {
 
         self.write_headers(&mut out);
 
-        if self.config.language == Language::Cxx {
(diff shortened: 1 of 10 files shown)


src/bindgen/bindings.rs
```diff
@@ -311,19 +326,33 @@ impl Bindings {
         if let Some(ref namespaces) = self.config.namespaces {
             wrote_namespace = true;
 
+            if self.config.language == Language::C && self.config.cpp_compat {
```
emilio: This could get a little cleanup, but I can do it after merging.

PR #413, opened 2019-11-17
Use self when possible.

adamnemecek:


emilio:
Not opposed to this, but why?

adamnemecek:
Idiomaticness.

emilio:
Sure, but some of the changes, in particular the ones on enums require the `type_alias_enum_variants` stuff, which became stable only recently.

Issue #373, opened 2019-08-06
Rust char is not wchar_t e.g. on Windows


hsivonen:
Rust's `char` is 32 bits wide. `wchar_t` is not 32 bits wide e.g. on Windows, so mapping Rust's `char` to `wchar_t` is incorrect. (`wchar_t` is non-portable and generally a bad idea.) Rust's `char` should map to C11 `char32_t`.

emilio:
I agree :)

Do you plan to fix this? It should be straight-forward.

adrianwong:
We've run into this same issue, in Prince. Mind if I gave this fix a shot?

adrianwong:
@emilio Thanks for reviewing the PR! Would it be possible to make a new release on crates.io please?

emilio:
Yeah, I'll try to get to that. A bit swamped atm, sorry :/

adrianwong:
Thanks Emilio, no worries.

adrianwong:
Hi, any news on a release date please? :)

adrianwong:
Hi, any news on a release date please? :)

emilio:
Just published 0.10.0, sorry for the lag.

adrianwong:
Not a problem, @emilio. Thanks for your work.

PR #414, opened 2019-11-20
Fix missing C header for char32_t

adrianwong:
My apologies, the change that I introduced in #396 failed to consider the generated C header requiring `uchar.h` for the `char32_t` definition. `char32_t` is defined as a fundamental type in C++, so I don't believe it's necessary to do the same.

As an aside, `char32_t` is C11-specific (as already mentioned in #373), will this cause any issues?

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -360,7 +360,7 @@ includes = ["my_great_lib.h"]
 # imports are included by default because our generated headers tend to require
 # them (e.g. for uint32_t). Currently, the generated imports are:
 #
-# * for C: <stdarg.h>, <stdbool.h>, <stdint.h>, <stdlib.h>
+# * for C: <stdarg.h>, <stdbool.h>, <stdint.h>, <stdlib.h>, <uchar.h>
 #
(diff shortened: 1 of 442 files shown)


adrianwong:
Build's failed due to timeout.

emilio:
Yeah, the travis failure was a caching issue, I restarted the build to hopefully address it.

emilio:
So, we compile all our C/C++ files on CI. Why was this not an issue? Does gcc somehow end up pulling in `<uchar.h>`?

adrianwong:
Turns out it wasn't an issue because there was no existing C test for `char`s.

(The `constant.rs` test which I updated in #396 tests `const char`, which maps to `#define` in the generated C headers.)

I've pushed a new commit that adds a basic `char.rs` test to cover this case.

adrianwong:
No, thank _you_ for the prompts! They were the right questions to ask.

If you could make a new release for this (when you can spare the time!), that would be fantastic.

emilio:
Just published v0.10.1 with this change and #422. Thanks again!

adrianwong:
Legend!

PR #416, opened 2019-11-24
Add support for after_includes

lu-zero:
Let the user add raw C code in the generated header after the include
blocks.

It matches what bindgen provides.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -392,6 +392,8 @@ no_includes = false
 # default: false
 cpp_compat = false
 
+# A list of lines to add verbatim after the includes block
+after_includes = "#define VERSION 1"
 
 
 
--- a/template.toml
+++ b/template.toml
@@ -23,6 +23,7 @@ using_namespaces = []
 sys_includes = []
 includes = []
 no_includes = false
+after_includes = ""
 
 
 
--- a/tests/rust/raw_lines.toml
+++ b/tests/rust/raw_lines.toml
@@ -1,0 +1,3 @@
+include_guard = "INCLUDE_GUARD_H"
+no_includes = false
+after_includes = "#define VERSION 1"
--- a/src/bindgen/bindings.rs
+++ b/src/bindgen/bindings.rs
@@ -151,6 +151,7 @@ impl Bindings {
         if self.config.no_includes
             && self.config.sys_includes.is_empty()
             && self.config.includes.is_empty()
+            && self.config.after_includes.is_none()
         {
             return;
         }
@@ -193,6 +194,11 @@ impl Bindings {
 
         for include in &self.config.includes {
             write!(out, "#include \"{}\"", include);
+            out.new_line();
+        }
+
+        if let Some(ref line) = self.config.after_includes {
+            write!(out, "{}", line);
             out.new_line();
         }
     }
--- a/src/bindgen/builder.rs
+++ b/src/bindgen/builder.rs
@@ -61,6 +61,12 @@ impl Builder {
     }
 
     #[allow(unused)]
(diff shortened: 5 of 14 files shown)


lu-zero:
Somehow travis managed to hang, you'd be fine in switching to github actions?

emilio:
This needs docs and tests, but why is this different from adding an include file with whatever code you want? It seems better than stashing all the C code in the config file.

lu-zero:
Somehow your reply got lost in my mailbox, I'll update it with docs and tests.

My main use-case is to add the crate version information automatically in cargo-c in a cleaner way and probably more people using both bindgen and cbindgen my enjoy having the symmetry :)

ultrasaurus:
this seems very similar to the solution needed for issue https://github.com/eqrion/cbindgen/issues/303, which was worked on a bit in PR https://github.com/eqrion/cbindgen/pull/306 where the proposed syntax was

```
[embed]
before = "file.h"
```

in this PR `raw_lines` is included after include guard but before includes and extern.  It might be good to use a config section, so then this could be expanded upon to include in different parts of the header, like

```
[embed.before]
sys_includes = "define.h"
generated = "mytypes.h"
```

which would leave room for
```
[embed.after]
generated = "myextrastuff.h"
```

Then this PR could implement only:
```
[embed.before]
sys_includes = "define.h"
```

This is mostly a suggestion on config naming/structuring, though I agree with @emilio  that it is nice to keep code in separate files, so I used that in my example config suggestion above.

lu-zero:
The two approaches solve two different problems I'm afraid.

ultrasaurus:
I am curious about how raw_lines vs file inclusion are different problems.  

FYI: I was just chiming in with idea... did not mean to derail anything. I'm just a bystander, learning about cbindgen :)

lu-zero:
`raw lines` does not need I/O and can be used from `build.rs` or similar.

Embedding already existing files is really an half step from just adding them to the distribution and just `#include` them.

lu-zero:
@emilio I rebased it again, can it get in?

tests/rust/raw_lines.toml
```diff
@@ -0,0 +1,3 @@
+include_guard = "INCLUDE_GUARD_H"
+no_includes = true
```
emilio, line 2: Let's test that this works actually after the includes?

lu-zero: Sure, patch updated

lu-zero:
I renamed the option as you suggested now :)

emilio:
It seems you didn't turn it into a string, is there a strong reason for that?

I'd prefer it to be a string for consistency with the other options, so for example instead of:

```
# A list of lines to add verbatim after the includes block
after_includes = ["#define VERSION 1", ""]
```

You'd do:

```
after_includes = """
#define VERSION 1
"""
```

lu-zero:
> It seems you didn't turn it into a string, is there a strong reason for that?

Not a strong reason, patch updated. I just overlooked that part, sorry

docs.md
```diff
@@ -392,6 +392,8 @@ no_includes = false
 # default: false
 cpp_compat = false
 
+# A list of lines to add verbatim after the includes block
+after_includes = ["#define VERSION 1", ""]
```
emilio, line 396: Docs are still using the old syntax.

template.toml
```diff
@@ -23,6 +23,7 @@ using_namespaces = []
 sys_includes = []
 includes = []
 no_includes = false
+after_includes = []
```
emilio, line 26: Template is still using the wrong syntax.

lu-zero:
Now it should be fine :)

PR #419, opened 2019-11-29
Support bulk rename of exported items

### LINKED ISSUE #277: Allow bulk rename of exported items
It should be possible to apply bulk conversions to top-level items (e.g. rename all types to use lowercase snake case).

yshui:
Closes #277

### DIFF
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -238,6 +238,8 @@ pub struct ExportConfig {
     pub rename: HashMap<String, String>,
     /// Table of raw strings to prepend to the body of items.
     pub pre_body: HashMap<String, String>,
+    /// Rename rule for all exported items
+    pub rename_all: crate::bindgen::rename::RenameRule,
     /// Table of raw strings to append to the body of items.
     pub body: HashMap<String, String>,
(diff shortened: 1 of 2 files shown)


yshui:
@emilio I indeed am unsure which function to use here. But I think I saw `apply_to_pascal_case` being called with strings that seem to me are not guaranteed to be pascal case, so I thought it was fine.

yshui:
@emilio Do you have suggestions what I should do here?

emilio:
Sure, you should at the very least write some tests that exercise all this code. You can see a bunch of tests using various `cbindgen.toml` options in the `tests/` directory.

Then based on that we can probably see whether the patch itself needs changes or apply_to_pascal_case is just fine.

yshui:
@emilio In the end I settled for adding some heuristics to guess the case of the input string.

yshui:
@emilio Hi, do you think i took the right approach?

emilio:
Hey, yeah, sorry, need to get to review this... I'm not a fan of heuristics generally, but I'll take a closer look.

yshui:
@emilio Any news?

yshui:
Closing in favor of #703

Issue #412, opened 2019-11-11
cbindgen doesnt understand simple shifts and bitwise OR


jaseemabid:
Constants defined with << and | doesn't seem to be supported. 

For example, this block fails

```rs
pub const FALSE: i64 = (0 << SHIFT) | BOOL;
pub const TRUE: i64 = (1 << SHIFT) | BOOL;
```
with the error

```
WARN: Skip inc::FALSE - (Unsupported literal expression. Paren(ExprParen { attrs: [], paren_token: Paren, expr: Binary(ExprBinary { attrs: [], left: Lit(ExprLit { attrs: [], lit: Int(LitInt { token: 0 }) }), op: Shl(Shl), right: Path(ExprPath { attrs: [], qself: None, path: Path { leading_colon: None, segments: [PathSegment { ident: Ident(SHIFT), arguments: None }] } }) }) }))
WARN: Skip inc::TRUE - (Unsupported literal expression. Paren(ExprParen { attrs: [], paren_token: Paren, expr: Binary(ExprBinary { attrs: [], left: Lit(ExprLit { attrs: [], lit: Int(LitInt { token: 1 }) }), op: Shl(Shl), right: Path(ExprPath { attrs: [], qself: None, path: Path { leading_colon: None, segments: [PathSegment { ident: Ident(SHIFT), 
```

I've no idea how hard it is to support something like this, but it looks like something that can be translated to into C. I'm happy to give this a try if this is good to have feature. 

PS:  I just used cbindgen to rewrite a lot of C in Rust (https://github.com/jaseemabid/inc/commit/314e5fef3dfd540e3d97872bffc84f3a476134eb) and it was a pretty good experience except some minor hiccups. I've also filed #411 and is happy to work on both issues. Thanks a lot for this project! :D

emilio:
This should be relatively straight-forward to fix, I think. We support other operators already.

PR #425, opened 2019-12-04
Add support for expressions like `(0 << SHIFT) | BOOL`

jaseemabid:
The PR adds support for parenthesized expressions, identifiers and bit wise or operators. Fixes https://github.com/eqrion/cbindgen/issues/412. 

See individual commits for details. 

When handling identifiers, I've resorted to a naive impl like this
```rust                
if segments.len() == 1 {                
    Ok(Literal::Expr(format!("{}", segments.last().unwrap().ident)))
```

Let me know if this is not OK, I'm not sure if `ir::Path` is really necessary in this case.


src/bindgen/ir/constant.rs
```diff
@@ -183,7 +207,26 @@ impl Literal {
...
+
+            // Match identifiers, like `5 << SHIFT`
+            syn::Expr::Path(syn::ExprPath {
+                path: syn::Path { ref segments, .. },
+                ..
+            }) => {
+                // Handle only the simplest identifiers and error for anything else.
+                // NOTE: Should this use `ir::Path` instead?
+                if segments.len() == 1 {
+                    Ok(Literal::Expr(format!("{}", segments.last().unwrap().ident)))
```
emilio: This should be a different kind of `Literal`, so that it can work correctly with `export.prefix` and such configuration options. `rename_for_config` should be called on it.

Please add a test for this to the `prefix` tests.

tests/expectations/constant.compat.c
```diff
@@ -21,8 +21,16 @@
...
 
+#define SHIFT 3
+
 #define TAB '\t'
 
+#define XBOOL 1
+
+#define XFALSE (0 << SHIFT) | XBOOL
+
+#define XTRUE (1 << SHIFT) | XBOOL
```
emilio: This is pretty footgunny C code. We should probably parenthesize all binary ops when emmitting to a macro. That would also avoid the need for `Literal::Paren`, I think, which would maybe make the patch a bit simpler.

src/bindgen/ir/constant.rs
jaseemabid: > This should be a different kind of Literal

@emilio  Just to be clear, did you mean I should add a variant to enum `ir::constant::Literal` to handle names instead of reusing Expr? That would indeed be better. What do you think we should call it? Name/Reference/Segment/Path ? 

Makes sense about renaming for config, will do that and add a test :+1:

tests/expectations/constant.compat.c
jaseemabid: @emilio 

> This is pretty footgunny C code. We should probably parenthesize all binary ops when emmitting to a macro. 

I agree, this can go wrong in so many ways - happy to add a parenthesis around the whole expression if there is a binary op present. Should be able to peek into the structure inside the write function do this. 

>  That would also avoid the need for Literal::Paren, I think, which would maybe make the patch a bit simpler.

I agree, adding that variant did make me cringe a bit. I prefer to not have it there. 

Do you think I should just recursively call the value inside and avoid the wrapping all together? Something like this? 

```patch
- syn::Expr::Paren(syn::ExprParen { ref expr, .. }) => Ok(Literal::Paren {
-     inner: Box::new(Self::load(expr)?),
- }),

+ syn::Expr::Paren(syn::ExprParen { ref expr, .. }) => Self::load(expr),
```

src/bindgen/ir/constant.rs
emilio: Yeah. `Literal::Path` would make sense to me, wdyt?

tests/expectations/constant.compat.c
emilio: Yeah, thtat should work, but let me know if it doesn't.

src/bindgen/ir/constant.rs
jaseemabid: Literal::Path(String) sounds good!

tests/expectations/constant.compat.c
jaseemabid: I was slightly sceptical but this seems to have worked fine, working on a few tests now to make sure the config and renaming work correctly.

jaseemabid:
@emilio Thanks for the initial review, v2 looks much better. 

The prefix tests seems to be renaming things alright, there is no Paren anymore and operator precedence around () seems to be working as expected.

jaseemabid:
Thanks @emilio :pray:

Issue #430, opened 2019-12-12
#[repr(align(n), C)] and #[repr(packed, C)] are not represented correctly


cyphar:
`#[repr(align(n), C)]` markers on a struct should be translated to `__attribute__((align(n))`, similarly for `#[repr(packed, C)]` for `__attribute__((packed))`. Right now they are generated as being opaque structs. This is necessary to implement a Rust-based `__attribute__((packed))` (by setting `#[repr(align(1))]`).

Additionally, `#[repr(align(n))]` and `#[repr(C)]` separately applied on the same struct (which AFAIK is equivalent to `#[repr(align(n), C)]`) will result in the structure being generated as a non-opaque struct but with the wrong alignment (resulting in C code passing incorrect structs to Rust code -- which won't end well).

PR #431, opened 2019-12-13
{struct,union}: support #[repr(align(...))] and #[repr(packed)]

cyphar:
As with #[must_use], this new support requires users to specify what
annotations they would like to apply when an alignment-related repr is
encountered.

In order to facilitate the fact that #[repr(align(...))] takes an
argument which will likely need to be included in the final annotation,
the user must specify a function-like macro which takes the alignment
value as an argument.

It's not clear to me how we should handle aligned enums, so I've
punted on those for now.

Fixes: #430 
Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>


src/bindgen/ir/enumeration.rs
```diff
@@ -250,8 +252,12 @@ impl Enum {
 
     pub fn load(item: &syn::ItemEnum, mod_cfg: Option<&Cfg>) -> Result<Enum, String> {
         let repr = Repr::load(&item.attrs)?;
-        if repr == Repr::RUST {
-            return Err("Enum not marked with a valid repr(prim) or repr(C).".to_owned());
+        if repr.style == ReprStyle::Rust && repr.ty.is_none() {
```
emilio: Why this change? I don't think `#[repr(align = ..)] enum { ... }` has a stable ABI.

src/bindgen/ir/union.rs
```diff
@@ -35,9 +35,13 @@ pub struct Union {
 
 impl Union {
     pub fn load(item: &syn::ItemUnion, mod_cfg: Option<&Cfg>) -> Result<Union, String> {
-        if Repr::load(&item.attrs)? != Repr::C {
+        let repr = Repr::load(&item.attrs)?;
+        if repr.style != ReprStyle::C {
             return Err("Union is not marked #[repr(C)].".to_owned());
         }
+        if repr.align.is_some() {
+            return Err("Union is marked with #[repr(align(...))] or #[repr(packed)].".to_owned());
```
emilio: It doesn't seem hard to support this, but if you don't, mind filing a follow-up issue?

tests/rust/aligned_struct.toml
```diff
@@ -0,0 +1,21 @@
+[struct]
+packed = "__attribute__ ((packed))"
+aligned = "__attribute__ ((aligned(<...>)))"
```
emilio: I don't think this is very discoverable... I think we should just take a macro name and assume it's functional, so like:

```
aligned = "CBINDGEN_ALIGNED"
```

And then have in the `header`:

```
#define CBINDGEN_ALIGNED(n) __attribute__ ((aligned(n)))
```

WDYT?

tests/rust/aligned_struct.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+#[repr(align(32), C)]
+pub struct Align32Struct {
+    pub arg1: usize,
+    pub arg2: *mut u8,
+}
+
+// Packed doesn't take an argument.
+#[repr(packed(4), C)]
```
emilio: Actually it can, see https://doc.rust-lang.org/reference/type-layout.html.

We should either `warn!` or add a third `packed_n` annotation.

src/bindgen/ir/enumeration.rs
```diff
@@ -250,8 +252,12 @@ impl Enum {
...
+            return Err("Enum is not marked with a valid #[repr(prim)] or #[repr(C)].".to_owned());
+        }
+        // TODO: Implement translation of aligned enums.
```
emilio: Please file a follow-up issue on this.

This is technically a breaking change too, but before that the layout was wrong so...

tests/rust/aligned_struct.rs
```diff
@@ -0,0 +1,84 @@
...
+pub struct BadPacked4Struct {
+    pub arg1: usize,
+    pub arg2: *mut u8,
+}
+
+// Alignment must be an integer.
+#[repr(align("1"), C)]
```
emilio: Please make sure that rustc compiles this. If it doesn't it's not really worth putting here IMO.

tests/rust/aligned_struct.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub arg1: usize,
+    pub arg2: *mut u8,
+}
+
+// Alignment takes exactly one argument.
+#[repr(align(2, 4), C)]
+pub struct BadAlign2_4Struct {
+    pub arg1: usize,
+    pub arg2: *mut u8,
+}
```
emilio: Let's add also tests for only having repr(align) and such, and not having `repr(C)`. Those shouldn't be generated.

src/bindgen/ir/enumeration.rs
cyphar: `#[repr(align(x))] enum` still results in an opaque type -- the reason for this change was that I dropped the comparisons to `Repr` because a few of them caused tests to fail once I added `Repr.align` (because now `repr == Repr::RUST` fails if the type has alignments set).

I can re-add it back, but I personally find it more confusing to have two different ways of effectively checking `repr.style`.

cyphar: Also, `#[repr(align(x))] enum` was stabilised in Rust 1.37 -- https://github.com/rust-lang/rust/pull/61229.

src/bindgen/ir/union.rs
cyphar: I can add support for it.

src/bindgen/ir/enumeration.rs
emilio: Oh, this was my mistake. What I meant is, why isn't this just `if repr.style == ReprStyle::Rust { return Err(..); }`. But this is wrong as we'd stop generating `#[repr(u8)]` etc.

So please disregard this :)

tests/rust/aligned_struct.toml
cyphar: Sure, as long as it's documented. However, it's unclear to me how we should handle `#[repr(packed(x))]` -- should there be two macros or just one that always takes an argument (and for `#[repr(packed)]` we pass an alignment of `1`)? In C, I'm pretty sure the equivalent of `#[repr(packed(x))]` is `__attribute__((packed, aligned(x)))` which is a problem because `#[repr(packed)]` is `__attribute__((packed))` not `__attribute__((packed, aligned(1)))`. So AFAICS we'd need two different macros?

tests/rust/aligned_struct.rs
cyphar: Yeah, I can implement `packed_n` as well.

emilio: Feel free to do it as a separate PR if more convenient, that way it's easier to review too :)

src/bindgen/ir/enumeration.rs
cyphar: Yup, that was the reason for the removal of direct `repr ==` comparisons.

tests/rust/aligned_struct.toml
cyphar: Should we have any of those settings by default? Or should we keep it purely optional like `must_use`? Arguably it's not really correct to generate a non-opaque version of a structure without including it's alignment (and *especially* if we don't include it's packing).

tests/rust/aligned_struct.rs
cyphar: Several of the bad alignments don't compile (including the non-power-of-2 ones). I'll drop all of those.

tests/rust/aligned_struct.rs
cyphar: It turns out this is slightly more complicated to do correctly, so I will punt on this for now and do it in a follow-up PR.

If I'm reading [this](https://doc.rust-lang.org/nightly/reference/type-layout.html#the-alignment-modifiers) right, it turns out that `#[repr(packed(n))]` does not really have a simple native C representation. In Rust, `#[repr(packed(n))]` indicates that the fields should have the smaller of `n` and `std::mem::align_of<field_type>()`. While in C you can specify individual field alignments like so:

```
struct foo {
  uint64_t foo1 __attribute__((aligned(32)));
  uint32_t foo2 __attribute__((aligned(32)));
  uint16_t foo3 __attribute__((aligned(32)));
  uint8_t foo4 __attribute__((aligned(32)));
  bool foo5 __attribute__((aligned(32)));
} __attribute__((packed));
```

The issue is that, in Rust, the following struct:

```
#[repr(C, packed(32))]
struct Foo {
  foo1: u64;
  foo2: u32;
  foo3: u16;
  foo4: u8;
  foo5: bool;
}
```

Should actually be represented as (depending on your architecture):

```
struct foo {
  uint64_t foo1 __attribute__((aligned(8)));
  uint32_t foo2 __attribute__((aligned(4)));
  uint16_t foo3 __attribute__((aligned(2)));
  uint8_t foo4 __attribute__((aligned(1)));
  bool foo5 __attribute__((aligned(1)));
} __attribute__((packed));
```

And the issue is that it's not clear to me how we can determine what the natural alignment of the constituent types is. And unfortunately, this is critical to making sure that C and Rust code can interoperate with packed structures. It's a bit unfortunate that Rust decided to have different semantics to GCC's alignment extensions, because I'm not really sure how to implement them. And given that `std::mem::align_of::<>()` depends on your architecture, we'd have to do lots of extra work to generate the right alignment for each architecture rust supports.

cyphar: Filed the above as #433.

src/bindgen/ir/enumeration.rs
cyphar: Filed #432.

src/bindgen/ir/union.rs
cyphar: Done.

cyphar:
The changes you requested have been implemented, but I still have one remaining question (which appears to be a bit difficult to implement with the current way the parser is set up -- the parser doesn't know the full configuration of `cbindgen`).

Arguably it's not really correct to generate a non-opaque version of a structure without including it's alignment (and especially if we don't include it's packing). So ideally I think we should treat structs that have alignment as opaque if there isn't a matching alignment macro defined in the configuration. WDYT?

emilio:
And to elaborate: the unfortunate bit that makes `must_use` weirder (I hope) than this, is that C compilers have different attributes for functions / enum / structs. In particular `__attribute__((warn_unused_result))` works for functions, `__attribute__((warn_unused))` works for structs / unions, and at least in the past we needed custom annotations/static analysis for enums.

tests/rust/aligned_struct.rs
emilio: Thanks for filing that! Yeah, I wonder how far can you go with `alignas` and decltype, if you're using C++...

tests/rust/aligned_struct.toml
emilio: I think what you did (dropping if we detect we'd generate bad code) is the right thing to do.

src/bindgen/ir/repr.rs
```diff
@@ -52,33 +59,101 @@ impl Repr {
...
+                        return Err(format!(
+                            "Conflicting #[repr(align(...))] type hints {:?} and {:?}.",
+                            old_align,
+                            ReprAlign::Align(align)
+                        ));
+                    }
+                    repr.align = Some(ReprAlign::Align(align));
+                    continue;
+                }
+                (path, args @ _) => match args {
```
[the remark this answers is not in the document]
cyphar: Yup, you're quite right -- for some reason I keep using `@ _` needlessly.

tests/rust/aligned.toml
```diff
@@ -0,0 +1,38 @@
...
+#define CBINDGEN_STRUCT_PACKED CBINDGEN_PACKED
+
+#define CBINDGEN_ALIGNED(n)     __attribute__ ((aligned(n)))
+#define CBINDGEN_UNION_ALIGNED  CBINDGEN_ALIGNED
+#define CBINDGEN_STRUCT_ALIGNED CBINDGEN_ALIGNED
+"""
+
+[union]
+packed = "CBINDGEN_UNION_PACKED"
+aligned_n = "CBINDGEN_UNION_ALIGNED"
```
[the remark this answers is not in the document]
cyphar: I think with GCC, they consistently use `aligned(...)` and `packed` -- `enum`s might be a bit odd but that's because you need to pass a compiler flag to enable all the alignment features.

src/bindgen/config.rs
```diff
@@ -340,6 +351,14 @@ pub struct StructConfig {
 }
 
 impl StructConfig {
+    pub(crate) fn can_safely_represent(&self, align: &ReprAlign) -> Result<(), String> {
```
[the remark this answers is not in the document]
cyphar: I can make it `ensure_safe_to_represent` or something.

cyphar:
My one follow-up question is are you sure that you want to re-use the `struct` annotations instead of defining them globally? I'd find it a bit odd (as a user) that a field set in `[struct]` will affect `union`s (and eventually `enum`s).

tests/rust/aligned.toml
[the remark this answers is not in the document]
emilio: Ugh, MSVC only supports `#pragma` to pack stuff? That kinda sucks :(

src/bindgen/config.rs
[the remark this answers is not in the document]
emilio: Works for me either way, really.

emilio:
> My one follow-up question is are you sure that you want to re-use the struct annotations instead of defining them globally? I'd find it a bit odd (as a user) that a field set in [struct] will affect unions (and eventually enums).

I don't know, I think it'd be a bit unfortunate to have to define the same thing three times to get packed / aligned structs to show up right...

On the other hand I see your point. I guess it depends on whether you see `struct` as "thing that generates a struct" (in which case rust ADTs will fall into it too, for example) or just as "rust struct".

Maybe there should be an `[attributes]` (or `[layout]`) section for these? Anyhow given the precedent, probably it is fine to just use `struct` for now, unless you feel strongly against it. I'd prefer if stuff just worked instead of having to duplicate the same configuration 3 times.

cyphar:
Sorry, I should clarify -- what I meant was that I'm on-board to not having the same config option multiple times. My question was whether there was a reason to not put it in the main `Config` struct (or as you mentioned, a separate `[layout]` section)?

tests/rust/aligned.toml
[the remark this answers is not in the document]
cyphar: [Yeah](https://docs.microsoft.com/en-us/cpp/preprocessor/pack?view=vs-2019). Luckily MSVC does have attributes for alignment ([`alignas`](https://docs.microsoft.com/en-us/cpp/cpp/alignof-and-alignas-cpp?view=vs-2019) and [the older `__declspec(align(n))`](https://docs.microsoft.com/en-us/cpp/cpp/align-cpp?view=vs-2019)). So this is only an issue for packing.

But we could actually implement MSVC-style packing (as a separate option), because [GCC does actually support `#pragma pack`](https://gcc.gnu.org/onlinedocs/gcc-4.4.4/gcc/Structure_002dPacking-Pragmas.html). It would look something like:

```
#pragma pack(push, 1)
typedef struct foo {
  // ...
};
#pragma pack(pop)
```

emilio:
I'm totally fine making it a separate section of the configuration, it's your call :)

tests/rust/aligned.toml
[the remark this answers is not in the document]
emilio: Yup, I concur. If somebody needs it it should be pretty easy to add support for it.

cyphar:
There, all the relevant changes have been made.

emilio:
Let me know if you need a release with these changes and I can try to prepare one soon-ish.

PR #438, opened 2019-12-20
Fix dependency resolution with lockfile v2.

### LINKED ISSUE #436: Fails with new `Cargo.lock` format
`cargo` has enabled the new format by default[1] and this causes the `cbindgen` to fail to parse the dependencies.

1. https://github.com/rust-lang/cargo/pull/7579

It fails at https://github.com/eqrion/cbindgen/blob/master/src/bindgen/cargo/cargo.rs#L125 as the new format does not has the version annotated on the same line.

emilio:
Lockfile v2 only includes version numbers if they're conflicting.

Fixes #436.


otavio:
This works just fine.

In long term, it'd be good to change `cbindgen` to use same code of `cargo` and avoid code duplication. However, some work at `cargo` side need to be done to allow that to happen.

kvark:
Not familiar with this code enough to review it properly...

emilio:
I think this is trivial enough to get merged. Though a sanity check from @Gankra when they have the time would be awesome of course.

baszalmstra:
Our beta and nightly pipelines seem to fail without this fix. Would it be possible to create a release?

emilio:
https://github.com/eqrion/cbindgen/commit/f5d76c44c466b47d1c776acd9974df838f30d431

PR #440, opened 2019-12-20
Update to rust 2018.

emilio:
This is just `cargo fix --edition` and the change in Cargo.toml.


emilio:
I don't think there's any reason not to do this, so I'm going to go ahead and merge this.

PR #441, opened 2019-12-22
ir: support #[repr(*64)] enums

cyphar:
Since there is a native C type for 64-bit integers (*int64_t), we can
just add support for them. Support for larger enums (u128) are currently
unstable and don't have a reasonable (standard-enough) C representation
to support at the moment.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

### DIFF
--- a/tests/rust/enum.rs
+++ b/tests/rust/enum.rs
@@ -1,9 +1,9 @@
 enum Opaque {
     Foo(i32),
-    Bar
+    Bar,
 }
 
-#[repr(u32)]
+#[repr(u64)]
 enum A {
     a1 = 0,
     a2 = 2,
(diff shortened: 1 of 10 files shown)


tests/rust/enum.rs
```diff
@@ -1,98 +1,106 @@
 enum Opaque {
     Foo(i32),
-    Bar
+    Bar,
 }
 
-#[repr(u32)]
+#[repr(u64)]
```
emilio: Would've been easier to just add a new enum instead, but nbd.

emilio:
Yeah, u128 and co have big issues (see https://github.com/rust-lang/rust/issues/54341).

cyphar:
Wow, that was fast. :D

tests/rust/enum.rs
cyphar: I figured you might want me to keep the same ordering in the test (then again, it doesn't really matter).

PR #446, opened 2020-01-03
Fix for finding dependency version in lockfile v2

baszalmstra:
With lockfile v2 a version is not required for dependencies if the lock
file only contains a single entry for a specified crate.  However, without
a version present, at a later stage, a dependant crate will not be found
in the metadata cache which causes the dependency to be ignored completely.
This commit tries to infer the version of dependant crates when the
version specifier is missing from the contents of the lockfile.


emilio:
Thanks for the patch! Could you elaborate on why is this needed?

Also, this needs a test, as it presumably changes behavior.

baszalmstra:
@emilio No problem, I've added the test case that fails without this patch and works with it.

The problem is as following: In the new lock file format, if there is only one version of a dependency then everywhere else in the lockfile the version will be omitted:

```
# This file is automatically @generated by Cargo.
# It is not intended for manual editing.
[[package]]
name = "dep"
version = "0.1.0"

[[package]]
name = "expand-dep"
version = "0.1.0"
dependencies = [
 "dep",
]
```

Here the package `expand-dep` depends on `dep` but as you might notice not on any particular version. Thats because there is only one version of `dep` in the lock file anyway. 

However, if cbindgen parses this lockfile the parsed version of the dependency `dep` will be `None` because it simply cannot parse it (its not there which was already fixed in #438 ). In `cargo.rs` (`fn dependencies`) it will also create a `PackageRef` with a `None` version. Later, when looking up the dependencies it cannot be found in the metadata because such a package (one without a version) doesn't exist.

This PR patches the building of dependent `PackageRef`s by building a `PackageRef` for dependencies that does include the correct version number. 

I hope that clears it up. :)

emilio:
The test doesn't seem to be passing on automation?

src/bindgen/cargo/cargo.rs
```diff
@@ -135,6 +135,28 @@ impl Cargo {
...
+                // If a version was not specified find the only package with the name of the dependency
+                let dep_version = dep_version.or_else(|| {
+                    let mut package_iter = self.metadata.packages
+                        .iter()
+                        .filter(|p| p.name_and_version.name == dep_name && p.name_and_version.version.is_some());
+
+                    // Get the first element from the iterator as the result.
+                    let result = package_iter
+                        .next()
+                        .and_then(|p| p.name_and_version.version.as_ref().map(|f| f.as_str()));
```
emilio: You can do `.map(|p| p.name_and_version.version.as_ref().unwrap().as_str())` if you want, as you've checked it's some...

But this may be cleaner with something like:

```rust
let versions = self.metadata.packages.iter().filter_map(|package| {
    if package.name_and_version.name != dep_name {
        return None;
    }
    p.name_and_version.version.as_ref()
});

let version = versions.next();
if versions.next().is_none() {
    version
} else {
    warn!("...");
    None
}
```

Wdyt?

baszalmstra: Thanks, thats much cleaner. I applied it.

baszalmstra:
Thanks @emilio, I've applied your suggestion, fixed the tests and rebased.

baszalmstra:
Thanks! Will you create a new release containing this patch?

Issue #420, opened 2019-11-30
Swift bindgen 


adamnemecek:
I'm considering adding a language option to cbindgen for Swift. 

The Swift compiler lets you include a C header which, when annotated with the CF_SWIFT_NAME macro exposes the C code to Swift code as Swift (so that one doesn't need to for example explicitly pass `self`/`this` to functions, similarly as with C++ cbindgen). 

Here's more info
https://developer.apple.com/documentation/swift/objective-c_and_c_code_customization/customizing_your_c_code_for_swift

Is there any interest in this feature?

emilio:
If you have a use case for this and it is not very invasive, I'd say go for it :)

This needs to have tests, which may be the trickiest part... Would this be testable in non-apple platforms?

PR #447, opened 2020-01-07
Add support for the export_name function attribute

### LINKED ISSUE #283: Should respect #[export_name]
Rust supports `#[export_name = "..."]` as an alternative to `#[no_mangle]` that allows to customise exported name of various items.

alaroldai:
Fixes #283 

Added a helper trait / methods on `syn::ItemFn` to handle attribute lookup, including:

- `export_name_attr` to lookup the attribute
- `exported_name` to select the correct name (`export_name` if it was set, otherwise the function name if it's `no_mangle`, otherwise `None`)
- `attr_name_value_lookup` added to `SynAttributeHelpers` to factor out attribute lookup

### DIFF
--- a/src/bindgen/parser.rs
+++ b/src/bindgen/parser.rs
@@ -17,7 +17,7 @@ use crate::bindgen::ir::{
     AnnotationSet, Cfg, Constant, Documentation, Enum, Function, GenericParams, ItemMap,
     OpaqueItem, Path, Static, Struct, Type, Typedef, Union,
 };
-use crate::bindgen::utilities::{SynAbiHelpers, SynItemHelpers};
+use crate::bindgen::utilities::{SynAbiHelpers, SynItemFnHelpers, SynItemHelpers};
 
 const STD_CRATES: &[&str] = &[
     "std",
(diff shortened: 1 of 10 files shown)


src/bindgen/utilities.rs
```diff
@@ -238,6 +261,25 @@ impl SynAttributeHelpers for [syn::Attribute] {
         })
     }
 
+    fn attr_name_value_lookup(&self, name: &str) -> Option<String> {
```
emilio: Can we rewrite `has_attr_name_value` to share code with this somehow? Maybe making this return an `impl Iterator<Item = String>`, making the caller's duty to call `.next()` to reduce the lookup to one result?

src/bindgen/utilities.rs
```diff
@@ -39,6 +39,28 @@ pub fn find_first_some<T>(slice: &[Option<T>]) -> Option<&T> {
...
 }
 
+pub trait SynItemFnHelpers: SynItemHelpers {
+    fn export_name_attr(&self) -> Option<String>;
+
+    fn exported_name(&self) -> Option<String>;
+}
+
+impl SynItemFnHelpers for syn::ItemFn {
+    fn export_name_attr(&self) -> Option<String> {
```
emilio: Given this only has one caller, probably worth inlining it instead of adding the function? It's a one liner anyway.

src/bindgen/parser.rs
```diff
@@ -585,12 +587,21 @@ impl Parse {
         } else {
             warn!("Skip {}::{} - (not `pub`).", crate_name, &item.sig.ident);
         }
-        if (item.sig.abi.is_omitted() || item.sig.abi.is_c()) && !item.is_no_mangle() {
+
+        if !(item.sig.abi.is_omitted() || item.sig.abi.is_c()) {
```
emilio: we should probably move these warnings above and making them early return instead, but this is fine if you don't want to do that refactoring for now.

src/bindgen/utilities.rs
alaroldai: Done

src/bindgen/utilities.rs
alaroldai: Done

src/bindgen/parser.rs
alaroldai: I'm doing this in a separate change (having a go at #420 - see [here](https://github.com/alaroldai/cbindgen/blob/swift-name/src/bindgen/parser.rs#L618))
I think it makes more sense in that branch because I'm also refactoring this whole function to share functionality with the new `load_syn_method` implementation.

src/bindgen/utilities.rs
```diff
@@ -225,17 +244,28 @@ impl SynAttributeHelpers for [syn::Attribute] {
     }
 
     fn has_attr_name_value(&self, name: &str, value: &str) -> bool {
-        self.iter().filter_map(|x| x.parse_meta().ok()).any(|attr| {
-            if let syn::Meta::NameValue(syn::MetaNameValue { path, lit, .. }) = attr {
-                if let syn::Lit::Str(lit) = lit {
-                    path.is_ident(name) && (&lit.value() == value)
-                } else {
-                    false
+        self.attr_name_value_lookup(name)
```
emilio: This doesn't seem to be equivalent to the old code, which has `any()` semantics... But it seems `has_attr_name_value` is unused, so let's remove it?

emilio:
Thanks!

PR #449, opened 2020-01-08
Support 'swift_name' attributes on generated functions

### LINKED ISSUE #420: Swift bindgen

alaroldai:
Fixes #420

- Added a parameter to `bindgen::ir::Function` to track attributes on generated functions. This is in addition to the existing `postfix` annotation, which seems to be used exclusively for user-provided strings, and so seems like a bad fit for auto-generated code.
- Added a config parameter under `FunctionConfig` to control whether swift function names are generated

No support for documentation annotations, but adding support for something like the following may be a good follow-up task:

```rust
impl FooRef {
  // cbindgen:swift-name=FooRef.doTheThing(with arg:)
  fn do_thing(arg: u8) {
  }
}
```

### DIFF
--- a/src/bindgen/ir/function.rs
+++ b/src/bindgen/ir/function.rs
@@ -23,6 +23,8 @@ use crate::bindgen::writer::{Source, SourceWriter};
 #[derive(Debug, Clone)]
 pub struct Function {
     pub path: Path,
+    /// Path to the self-type of the function
+    /// If the function is a method, this will contain the path of the type in the impl block
     pub self_type_path: Option<Path>,
     pub ret: Type,
     pub args: Vec<(String, Type)>,
@@ -30,7 +32,6 @@ pub struct Function {
     pub cfg: Option<Cfg>,
     pub annotations: AnnotationSet,
     pub documentation: Documentation,
-    pub attributes: Vec<String>,
 }
 
 impl Function {
@@ -42,7 +43,10 @@ impl Function {
         attrs: &[syn::Attribute],
         mod_cfg: Option<&Cfg>,
     ) -> Result<Function, String> {
-        let args = sig.inputs.iter().try_skip_map(|x| x.as_ident_and_type())?;
+        let args = sig
+            .inputs
+            .iter()
+            .try_skip_map(|x| x.as_ident_and_type(self_type_path.as_ref()))?;
         let ret = match sig.output {
             syn::ReturnType::Default => Type::Primitive(PrimitiveType::Void),
             syn::ReturnType::Type(_, ref ty) => {
@@ -63,7 +67,6 @@ impl Function {
             cfg: Cfg::append(mod_cfg, Cfg::load(attrs)),
             annotations: AnnotationSet::load(attrs)?,
             documentation: Documentation::load(attrs),
-            attributes: Vec::new(),
         })
     }
 
@@ -80,7 +83,8 @@ impl Function {
             ("".to_string(), "".to_string())
         };
 
-        let item_name = self.path
+        let item_name = self
+            .path
             .name()
             .trim_start_matches(type_name)
             .trim_start_matches('_');
@@ -88,14 +92,7 @@ impl Function {
         let item_args = {
             let mut items = vec![];
             for (arg, _) in self.args.iter() {
-                match arg.as_str() {
-                    "self" => {
-                        items.push("self:".to_string());
-                    }
-                    other => {
-                        items.push(format!("{}:", other));
-                    }
-                }
+                items.push(format!("{}:", arg.as_str()));
             }
             items.join("")
         };
@@ -203,17 +200,12 @@ impl Source for Function {
 
             if !func.extern_decl {
                 if let Some(ref postfix) = postfix {
-                    out.write(" ");
-                    write!(out, "{}", postfix);
+                    write!(out, " {}", postfix);
                 }
             }
 
             if let Some(ref swift_name_macro) = config.function.swift_name_macro {
-                let swift_name = func.swift_name();
-                if !swift_name.is_empty() {
-                    out.write(" ");
-                    write!(out, "{}({})", swift_name_macro, func.swift_name());
-                }
+                write!(out, " {}({})", swift_name_macro, func.swift_name());
             }
 
             out.write(";");
@@ -255,11 +247,7 @@ impl Source for Function {
             }
 
             if let Some(ref swift_name_macro) = config.function.swift_name_macro {
-                let swift_name = func.swift_name();
-                if !swift_name.is_empty() {
-                    out.write(" ");
-                    write!(out, "{}({})", swift_name_macro, func.swift_name());
-                }
+                write!(out, " {}({})", swift_name_macro, func.swift_name());
             }
 
             out.write(";");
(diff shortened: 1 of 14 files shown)


src/bindgen/ir/function.rs
```diff
@@ -166,6 +168,15 @@ impl Source for Function {
                     write!(out, "{}", postfix);
                 }
             }
+
+            let attributes = func
+                .attributes
+                .iter()
+                .map(|s| format!(" __attribute__(({}))", s))
```
emilio: We've generally avoided hardcoding attribute names and `__attribute__` declarations here. I think a slightly more consistent approach would be something like an `[fn] swift_name_macro = "FOO"` config. This would get the raw content of the string just like `CF_SWIFT_NAME` does. So that to use this feature the only thing you need to do would be `[fn] swift_name_macro = "CF_SWIFT_NAME"`, for example.

src/bindgen/utilities.rs
```diff
@@ -41,6 +41,53 @@ pub fn find_first_some<T>(slice: &[Option<T>]) -> Option<&T> {
 
 pub trait SynItemFnHelpers: SynItemHelpers {
     fn exported_name(&self) -> Option<String>;
+
+    fn swift_name(&self, type_name: Option<&syn::Ident>) -> String;
```
emilio: Seems like this should belong to our IR instead of syn... Why is this here?

alaroldai: You're right, this doesn't need to be here. Moved it into `bindgen::ir::Function`.

src/bindgen/ir/function.rs
alaroldai: Yes, that's a much better solution. It doesn't work quite so well for the `__attribute__` version of this behaviour because of the double-parentheses needed, but most users will probably prefer `CF_SWIFT_NAME` or similar instead (and a macro can be defined and included separately if necessary).

src/bindgen/ir/function.rs
```diff
@@ -53,15 +56,52 @@ impl Function {
...
+
+        let item_name = self.path
+            .name()
+            .trim_start_matches(type_name)
+            .trim_start_matches('_');
+
+        let item_args = {
+            let mut items = vec![];
+            for (arg, _) in self.args.iter() {
+                match arg.as_str() {
```
emilio: This is `match` is just `items.push(format!("{}:", arg.as_str())` right?

src/bindgen/ir/function.rs
```diff
@@ -160,12 +200,22 @@ impl Source for Function {
...
             if !func.extern_decl {
                 if let Some(ref postfix) = postfix {
                     out.write(" ");
                     write!(out, "{}", postfix);
                 }
             }
+
+            if let Some(ref swift_name_macro) = config.function.swift_name_macro {
+                let swift_name = func.swift_name();
+                if !swift_name.is_empty() {
```
emilio: this can never be empty, right?

src/bindgen/ir/function.rs
```diff
@@ -160,12 +200,22 @@ impl Source for Function {
...
+                    out.write(" ");
```
emilio: Instead of this, just put the space in the string literal below? These two comments apply to the following block as well.

src/bindgen/utilities.rs
```diff
@@ -57,6 +57,20 @@ impl SynItemFnHelpers for syn::ItemFn {
     }
 }
 
+impl SynItemFnHelpers for syn::ImplItemMethod {
+    fn exported_name(&self) -> Option<String> {
+        self.attrs
+            .attr_name_value_lookup("export_name")
+            .or_else(|| {
+                if self.is_no_mangle() {
+                    Some(self.sig.ident.to_string())
```
emilio: Does this really work for impl items? Like, does `impl Foo { extern fn bar(&self) { ... } }` really produce a `bar` symbol?

tests/rust/swift_name.rs
```diff
@@ -0,0 +1,46 @@
...
+    if let Some(nonnull) = std::ptr::NonNull::new(self.ptr) {
+      nonnull.as_ref().bar
+    }
+  }
+
+  /// cbindgen:postfix=/*a comment!*/
+  #[export_name="FooRef_doThing"]
+  pub extern fn do_thing(self: FooRef) -> i32 {
+    if let Some(nonnull) = std::ptr::NonNull::new(self.ptr) {
+      nonnull.as_ref().bar
```
emilio: Does this really compile as rust? I'm ~sure there's at least an `unsafe` block missing around.

emilio: Also four-space indent please?

tests/rust/swift_name.rs
```diff
@@ -0,0 +1,46 @@
...
+
+  #[export_name="FooRef_setBar"]
+  pub extern fn set_bar(self: FooRef, bar: i32) {
+    if let Some(nonnull) = std::ptr::NonNull::new(self.ptr) {
+      nonnull.as_mut().bar = bar;
+    }
+  }
+
+  #[export_name="FooRef_getBar"]
+  pub extern fn get_bar(self: FooRef) -> i32 {
```
emilio: So what happens when you use `&self` / `&mut self` / etc? What about `self: Box<Self>` or such?

Can we have tests for that? We should at least not generate broken code.

src/bindgen/ir/function.rs
```diff
@@ -23,17 +23,20 @@ use crate::bindgen::writer::{Source, SourceWriter};
...
 pub struct Function {
     pub path: Path,
+    pub self_type_path: Option<Path>,
     pub ret: Type,
     pub args: Vec<(String, Type)>,
     pub extern_decl: bool,
     pub cfg: Option<Cfg>,
     pub annotations: AnnotationSet,
     pub documentation: Documentation,
+    pub attributes: Vec<String>,
```
emilio: This can be removed now.

src/bindgen/ir/function.rs
```diff
@@ -23,17 +23,20 @@ use crate::bindgen::writer::{Source, SourceWriter};
 #[derive(Debug, Clone)]
 pub struct Function {
     pub path: Path,
+    pub self_type_path: Option<Path>,
```
emilio: Please add a comment saying something like "If we're a method, this will be `Some`, and the path will be the path of the type in the `impl` block" or such.

It'd be nice to make something more similar to what associated constants do, but it may be overkill for this.

src/bindgen/parser.rs
```diff
@@ -626,33 +626,40 @@ impl Parse {
             return;
         }
 
+        let loggable_item_name = {
```
emilio: This is a bit unfortunate in the sense that it runs this code unconditionally... But can improve in a follow-up.

emilio:
Fixed by:
 * https://github.com/eqrion/cbindgen/commit/cf406e964062549531b9b928f601be341a079c09 (this PR squashed)
 * https://github.com/eqrion/cbindgen/commit/688407c7c99e62d64276bf71b0af4380d0c46c5c
 * https://github.com/eqrion/cbindgen/commit/b74034313414f11cede8e22694d82aebcd70a112 (something I thought was worth doing, as the logging should happen in the uncommon case)

PR #452, opened 2020-01-12
Add [export.body_prepend] to config

### LINKED ISSUE #448: Mark non-pub fields as private in c++ headers
If I have the following Rust struct:

```rust
#[repr(C)]
pub struct MyStruct {
    value1: i32,
    pub value2: i32
}
```
It would be very useful for the generated struct to reflect the field's visibility:

```c++
struct MyStruct {
private:
    int32_t value1;
public:
    int32_t value2;
};
```

My main motivation for this is to create exported structs where **all fields are private**. An example is a pointer wrapper: If I allocate a pointer for an opaque struct with Box and then return the raw pointer, there's nothing stopping the caller from modifying it before passing it back for destruction. If instead I make a struct to wrap it, and give it private fields, all the caller can do is copy the struct around and pass it back unchanged.

A change like this would enable APIs that were more type-safe and more in the spirit of Rust.

An alternate, lower-tech approach would be an annotation to just unconditionally mark all fields private. I imagine most use-cases will want all-private or all-public fields, so an annotation to set them all private might be a smarter solution, and save field-specific visibility for a use case that needs it.

What do you think? Is there interest in this? If so I'd be happy to tackle it.

ejmahler:
This change allows users to prepend text to the body of structs etc. Previously, users could only append, and this change allows both.

While working on this, I noticed that appending for untagged enums doesn't seem to work:

It's in the test toml file:
```
"MyCLikeEnum" = """
  BogusVariantForSerializationForExample,
"""
```

But I looked through the history and `BogusVariantForSerializationForExample` has never appeared in any of the expected outputs. If this is a bug, I can open another pull request to fix it.

Fixes #448

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -488,7 +488,16 @@ renaming_overrides_prefixing = true
 "MyType" = "my_cool_type"
 "my_function" = "BetterFunctionName"
 
-# Table of things to add to the body of any struct, union, or enum that has the
+# Table of things to prepend to the body of any struct, union, or enum that has the
+# given name. This can be used to add things like methods which don't change ABI,
+# mark fields private, etc
+[export.body_prepend]
+"MyType" = """
+  MyType() = delete;
+private:
+"""
+
+# Table of things to append to the body of any struct, union, or enum that has the
 # given name. This can be used to add things like methods which don't change ABI.
 [export.body]
 "MyType" = """
--- a/docs.md
+++ b/docs.md
@@ -491,7 +491,7 @@ renaming_overrides_prefixing = true
 # Table of things to prepend to the body of any struct, union, or enum that has the
 # given name. This can be used to add things like methods which don't change ABI,
 # mark fields private, etc
-[export.body_prepend]
+[export.pre_body]
 "MyType" = """
   MyType() = delete;
 private:
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -232,9 +232,7 @@ impl ExportConfig {
     }
 
     pub(crate) fn pre_body(&self, path: &Path) -> Option<&str> {
-        self.pre_body
-            .get(path.name())
-            .map(|s| s.trim_matches('\n'))
+        self.pre_body.get(path.name()).map(|s| s.trim_matches('\n'))
     }
 
     pub(crate) fn post_body(&self, path: &Path) -> Option<&str> {
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -214,6 +214,8 @@ pub struct ExportConfig {
     pub exclude: Vec<String>,
     /// Table of name conversions to apply to item names
     pub rename: HashMap<String, String>,
+    /// Table of raw strings to prepend to the body of items.
+    pub body_prepend: HashMap<String, String>,
     /// Table of raw strings to append to the body of items.
     pub body: HashMap<String, String>,
     /// A prefix to add before the name of every item
@@ -229,7 +231,13 @@ impl ExportConfig {
         self.item_types.is_empty() || self.item_types.contains(&item_type)
     }
 
-    pub(crate) fn extra_body(&self, path: &Path) -> Option<&str> {
+    pub(crate) fn body_prepend(&self, path: &Path) -> Option<&str> {
+        self.body_prepend
+            .get(path.name())
+            .map(|s| s.trim_matches('\n'))
+    }
+
+    pub(crate) fn body_append(&self, path: &Path) -> Option<&str> {
         self.body.get(path.name()).map(|s| s.trim_matches('\n'))
     }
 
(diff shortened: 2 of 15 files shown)


ejmahler:
Looks like `test_body` failed, but I'm nor sure why. It passes locally, and the error message isn't helpful:

```
---- test_body stdout ----

Running: "/home/travis/build/eqrion/cbindgen/target/debug/cbindgen" "--lang" "c" "--cpp-compat" "--style" "type" "-o" "/home/travis/build/eqrion/cbindgen/tests/expectations/body.compat.c" "--verify" "--config" "/home/travis/build/eqrion/cbindgen/tests/rust/body.toml" "/home/travis/build/eqrion/cbindgen/tests/rust/body.rs"

thread 'test_body' panicked at 'cbindgen failed: "/home/travis/build/eqrion/cbindgen/tests/expectations/body.compat.c" with error: ', tests/tests.rs:54:5

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

tests/rust/body.toml
```diff
@@ -18,3 +18,25 @@
...
+"""
+
+"MyFancyEnum_Prepended" = """
+  #ifdef __cplusplus
+    inline void wohoo();
+  #endif
+"""
+
+"MyCLikeEnum_Prepended" = """
+  BogusVariantForSerializationForExample,
```
emilio: This doesn't work because it wouldn't be legal C++, so yeah, it is intended. I guess you could add more variants, but that didn't seem useful. Could be fixed if you have some use case for it but...

docs.md
```diff
@@ -488,7 +488,16 @@ renaming_overrides_prefixing = true
 "MyType" = "my_cool_type"
 "my_function" = "BetterFunctionName"
 
-# Table of things to add to the body of any struct, union, or enum that has the
+# Table of things to prepend to the body of any struct, union, or enum that has the
+# given name. This can be used to add things like methods which don't change ABI,
+# mark fields private, etc
+[export.body_prepend]
```
emilio: Could we call this `pre_body`? I think it reads better but... your call, if you feel strongly about this feel free to leave it as-is.

src/bindgen/config.rs
```diff
@@ -229,7 +231,13 @@ impl ExportConfig {
         self.item_types.is_empty() || self.item_types.contains(&item_type)
     }
 
-    pub(crate) fn extra_body(&self, path: &Path) -> Option<&str> {
+    pub(crate) fn body_prepend(&self, path: &Path) -> Option<&str> {
```
emilio: I think `pre_body` / `post_body` would read a bit nicer.

tests/expectations/body.cpp
```diff
@@ -47,8 +53,53 @@ union MyUnion {
...
+  struct Bar_Prepended_Body {
+    int32_t _0;
+  };
+
+  struct Baz_Prepended_Body {
+    int32_t _0;
+  };
+
+  #ifdef __cplusplus
+    inline void wohoo();
```
emilio: May be worth to put this at the top of everything? That allows to make everything private which can be nice. wdyt?

tests/rust/body.toml
ejmahler: I brought it up because it was in the documentation + test of the existing body example. If it's intended to not work, should we remove the existing example? It's also mentioned in the documentation.

docs.md
ejmahler: Will do.

src/bindgen/config.rs
ejmahler: Will do.

tests/expectations/body.cpp
ejmahler: Whoops. I most have overlooked this.

src/bindgen/ir/enumeration.rs
```diff
@@ -642,12 +650,14 @@ impl Source for Enum {
                 }
 
                 out.open_brace();
-            }
 
-            // Emit the body_prepend section, if relevant
-            if let Some(body) = config.export.pre_body(&self.path) {
-                out.write_raw_block(body);
-                out.new_line();
+                // Emit the pre_body section, if relevant
```
emilio: We should probably remove this C-specific path. Adding stuff to C is not so useful, and this is adding the body to the struct that has the tag instead of the enum, right?

ejmahler:
```
typedef struct MyFancyEnum_Prepended {
  #ifdef __cplusplus
    inline void wohoo();
  #endif
  MyFancyEnum_Prepended_Tag tag;
  union {
    Bar_Prepended_Body bar_prepended;
    Baz_Prepended_Body baz_prepended;
  };
} MyFancyEnum_Prepended;
```

This is how it looks for every C output. This is the correct place, right? Post_body puts it in the same place.

IMO we should try to keep consistency between the pre_body and post_body segments wherever possible, and the C tagged structs get post_body, so it would make sense forthem to get pre_body too.

emilio: Yeah, good point, I was mistaken here. Forgot about the anonymous union thing.

PR #453, opened 2020-01-12
Add a test for generic pointer parameter

kvark:
The test is a reduced version of https://github.com/gfx-rs/wgpu/pull/440#issuecomment-573284991 . It currently fails. cc @emilio

### DIFF
--- a/tests/rust/generic-pointer.rs
+++ b/tests/rust/generic-pointer.rs
@@ -1,0 +1,11 @@
+#[repr(C)]
+struct Foo<T> {
+    a: T,
+}
+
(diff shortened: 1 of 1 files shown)


emilio:
Ah, so that panic is basically an `unimplemented!`. Someone needs to come up with a mangling scheme for that, but it shouldn't be too hard.

TheKK:
Is there any name mangling guideline we can follow?
For this PR we need to implement mangling for `Type::ConstPtr(Box<Type>)` and it's a little tricky to handle nested type if we want to present them in flat type name.

PR #454, opened 2020-01-14
Handle new line in doc attribute

### LINKED ISSUE #442: `doc` attribute doesn't behave like rustc does
Inputs
---

```rust
#[no_mangle]
#[doc = "a \n b"]
pub extern "C" fn example_a() {}
```

and

```rust
#[no_mangle]
#[doc = "a \n\n b"]
pub extern "C" fn example_b() {}
```

Rendered rustdoc
---
[rustdoc/the doc attribute](https://doc.rust-lang.org/rustdoc/the-doc-attribute.html)

圖片

圖片

Actual generated header
---

```cpp
///a
 b
void example_a();
```

```cpp
///a

 b
void example_b();
```

Expected generated header
---

```cpp
///a b
void example_a();
```

```cpp
///a
///
///b
void example_b();
```

This happens when I'm trying to generate multi-line comments with macro. Looks like ([code](https://github.com/eqrion/cbindgen/blob/16fe3ec142653277d5405d9a6d25914d925c9c3c/src/bindgen/utilities.rs#L252)) we simply use value in single `doc` attribute directly without any modification (like rustdoc does).

BTW, I'm happy to help this out :)

TheKK:
This should fix #442

Please tell me if there anything I can do better :)

### DIFF
--- a/src/bindgen/utilities.rs
+++ b/src/bindgen/utilities.rs
@@ -272,8 +272,7 @@ impl SynAttributeHelpers for [syn::Attribute] {
                 })) = attr.parse_meta()
                 {
                     if path.is_ident("doc") {
-                        let text = content.value().trim_end().to_owned();
-                        comment.push(text);
+                        comment.extend(split_doc_attr(&content.value()));
                     }
                 }
             }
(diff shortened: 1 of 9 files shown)


src/bindgen/utilities.rs
```diff
@@ -261,24 +261,31 @@ impl SynAttributeHelpers for [syn::Attribute] {
     }
 
     fn get_comment_lines(&self) -> Vec<String> {
-        let mut comment = Vec::new();
-
-        for attr in self {
```
emilio: Is there any reason to rewrite the for loop instead of just changing `comment.push(text)` for `comment.extend(split_doc_attr(content.value())`?

It's not particularly easier to understand IMHO.

TheKK: Agree with you. I'll revert it back to the for-loop version and use Vec::extend instead. Thanks for suggestions!

TheKK:
Thanks for reviewing :)

PR #460, opened 2020-01-24
Enable most warnings, and fix findings.

sw17ch:
This PR enables `-Wall -Werror -Wno-attributes` when compiling C++ and C. With these warnings in place, several warnings are generated.

# C++ universal character name warning

```
/cbindgen/tests/expectations/constant.cpp:8:31: error: character constant too long for its type [-Werror]
    8 | static const uint32_t EQUID = L'\u10083';
      |                               ^~~~~~~~~~
cc1plus: all warnings being treated as errors
```

This is fixed by replacing the `L'\u10083'` syntax with `L'\U00010083'`. The `U` variant supports 8 digits, while `\u` only supports 4.

# C++17 inline variables warning

```
/cbindgen/tests/expectations/associated_in_body.cpp:45:1: error: inline variables are only available with ‘-std=c++17’ or ‘-std=gnu++17’ [-Werror]
   45 | inline const StyleAlignFlags StyleAlignFlags::AUTO = StyleAlignFlags{ /* .bits = */ 0 };
      | ^~~~~~
/Dev/cbindgen/tests/expectations/associated_in_body.cpp:46:1: error: inline variables are only available with ‘-std=c++17’ or ‘-std=gnu++17’ [-Werror]
   46 | inline const StyleAlignFlags StyleAlignFlags::NORMAL = StyleAlignFlags{ /* .bits = */ 1 };
      | ^~~~~~
/Dev/cbindgen/tests/expectations/associated_in_body.cpp:47:1: error: inline variables are only available with ‘-std=c++17’ or ‘-std=gnu++17’ [-Werror]
   47 | inline const StyleAlignFlags StyleAlignFlags::START = StyleAlignFlags{ /* .bits = */ (1 << 1) };
      | ^~~~~~
/Dev/cbindgen/tests/expectations/associated_in_body.cpp:48:1: error: inline variables are only available with ‘-std=c++17’ or ‘-std=gnu++17’ [-Werror]
   48 | inline const StyleAlignFlags StyleAlignFlags::END = StyleAlignFlags{ /* .bits = */ (1 << 2) };
      | ^~~~~~
/Dev/cbindgen/tests/expectations/associated_in_body.cpp:49:1: error: inline variables are only available with ‘-std=c++17’ or ‘-std=gnu++17’ [-Werror]
   49 | inline const StyleAlignFlags StyleAlignFlags::FLEX_START = StyleAlignFlags{ /* .bits = */ (1 << 3) };
      | ^~~~~~
```

Inline variables are only supported in C++17. This PR changes the `--std=c++11` C++ compilation option to `--std=c++17`.

# C++ left shift count >= width of type warning

```
/Dev/cbindgen/tests/expectations/prefix.cpp:8:40: error: left shift count >= width of type [-Werror=shift-count-overflow]
    8 | static const int64_t PREFIX_X = (42 << 42);
      | 
```

In the `prefix` test, a "random" value of 42 appears to have been used. It doesn't seem to have meaning to the `prefix` test. When left shifting 42 by 42 in C and C++, the left hand side of the shift is treated as an integer unless a suffix is specified. Left-shifting a 32 bit integer by 42 results in a value of 0. The compiler points out the programmer doesn't intend this, and, indeed, the surrounding code seems to indicate we expected the left hand side to be treated as a 64 bit integer.

### DIFF
--- a/tests/rust/prefix.rs
+++ b/tests/rust/prefix.rs
@@ -1,7 +1,7 @@
-pub const LEN: i32 = 42;
+pub const LEN: i32 = 22;
 
 pub type NamedLenArray = [i32; LEN];
-pub type ValuedLenArray = [i32; 42];
+pub type ValuedLenArray = [i32; 22];
 
 #[repr(u8)]
 pub enum AbsoluteFontWeight {
@@ -14,7 +14,7 @@ pub enum AbsoluteFontWeight {
 pub extern "C" fn root(x: NamedLenArray, y: ValuedLenArray, z: AbsoluteFontWeight) { }
 
 #[no_mangle]
-pub const X: i64 = 42 << 42;
+pub const X: i64 = 22 << 22;
 
 #[no_mangle]
 pub const Y: i64 = X + X;
--- a/src/bindgen/ir/constant.rs
+++ b/src/bindgen/ir/constant.rs
@@ -160,7 +160,7 @@ impl Literal {
                     syn::Lit::Byte(ref value) => Ok(Literal::Expr(format!("{}", value.value()))),
                     syn::Lit::Char(ref value) => Ok(Literal::Expr(match value.value() as u32 {
                         0..=255 => format!("'{}'", value.value().escape_default()),
-                        other_code => format!(r"L'\u{:X}'", other_code),
+                        other_code => format!(r"L'\U{:08X}'", other_code),
                     })),
                     syn::Lit::Int(ref value) => {
                         Ok(Literal::Expr(value.base10_digits().to_string()))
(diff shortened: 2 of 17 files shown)


tests/rust/prefix.rs
```diff
@@ -14,7 +14,7 @@ pub enum AbsoluteFontWeight {
 pub extern "C" fn root(x: NamedLenArray, y: ValuedLenArray, z: AbsoluteFontWeight) { }
 
 #[no_mangle]
-pub const X: i64 = 42 << 42;
+pub const X: i64 = 22 << 22;
```
emilio: We should probably just use larger types for the intermediate representations. This mostly works around the issue, but it's fine.

emilio:
Thanks!

sw17ch:
Thank you very much for merging this PR. I have opened #464 as requested.

PR #463, opened 2020-01-25
Fix a non-C-like-enumeration size mismatch between Rust and C when using `Style::Tag`.

sw17ch:
When using `Style::Tag` with non-C-like-enumerations and discriminant types that do not match the size of enumerations in C (such as those with `#[repr(C, u8)]`), cbindgen emits types that are incompatible between Rust and C.

I have found two problems:

# Inconsistent Padding

Here's the input type:

```rust
#[repr(C, u8)]
enum Tu64 {
    None,
    Some(u64),
}
```

Here's the generated C code:

```c
enum Tu64_Tag {
  None,
  Some,
};
typedef uint8_t Tu64_Tag;

struct Some_Body {
  uint64_t _0;
};

struct Tu64 {
  enum Tu64_Tag tag;
  union {
    struct Some_Body some;
  };
};
```

In this case, both C and Rust agree that `Tu64` is sized as 16 bytes. The `u64` forces its alignment to 8 bytes, so Rust designates the first byte as the discriminant, and the following 7 bytes as padding.

However, C has the tag defined as an `enum Tu64_Tag`. This means that the tag is sized the same way enumerations are on the host system. In our case, that's likely going to be 4 bytes. In C, the first 4 bytes are discriminant, and the following 4 are padding.

This means the interface between Rust and C is unsafe since the discriminant C reads may be influenced by noise in padding bytes that Rust may not initialize. For example, allocating a `Tu64` in C, setting it all to `0xFF`, then passing it by reference into Rust for initialization will result in a corrupted discriminant when C looks at the initialization result.

# Inconsistent Sizes

We'll use an example similar to the one above, but replace the `u64` with a `u8`.

```rust
#[repr(C, u8)]
enum Tu8 {
    None,
    Some(u8),
}
```

Here's the generated code:

```c
enum Tu8_Tag {
  None,
  Some,
};
typedef uint8_t Tu8_Tag;

struct Some_Body {
  uint8_t _0;
};

struct Tu8 {
  enum Tu8_Tag tag;
  union {
    struct Some_Body some;
  };
};
```

In this case, Rust sees that the value we can hold in the enum and the discriminant are both of size 1, so we can safely pack them in 2 bytes.

However, in C, we can see that we're still using `enum Tu8_Tag` for the discriminant tag type. This means that the size in C is going to be at least the size of an enumeration + 1 byte, and will have an alignment other than the one used by Rust.

Passing a `Tu8` from C to Rust results in Rust reading a corrupted body value. Passing a `Tu8` from Rust to C will almost certainly result in corruption of the tag. Furthermore, any mutation C does to the `Tu8` allocated by Rust can corrupt unrelated structures in Rust's memory by overflowing the bounds of Rust's `Tu8`.

# Solution

In both cases, the solution is to use the `typedef` that's generated as the tag type in C. That is, using `Tu64_Tag tag` instead of `enum Tu64_Tag tag` and `Tu8_Tag tag` instead of `enum Tu8_Tag tag`.

# Testing

This PR includes some extra infrastructure so that tests to verify this sort of thing can be written. The details of this test system are documented in the commit message of ea03db1. The test I added is at `tests/checkers/tag/enum.c`. The body of this test `#include`s the cbindgen generated file, and then verifies the expected size. If bac063e is checked out, this test is seen to fail in `cargo test`.

# Fix

I have created another commit that appears to fix this issue (by not emitting the `enum` before the tag type name when the representation expresses a size), and does not cause any other tests to fail. I'm not convinced this is the best way to fix this, nor is the importance of the change in df0167b  obvious. I would appreciate feedback about alternative ways to fix this that might be more clear to future readers.

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -682,7 +682,8 @@ impl Source for Enum {
                 out.open_brace();
             }
 
-            if config.language == Language::C && !config.style.generate_typedef() {
+            if config.language == Language::C && !size.is_some() && !config.style.generate_typedef()
+            {
                 out.write("enum ");
             }
 
--- a/tests/expectations/both/enum.c
+++ b/tests/expectations/both/enum.c
@@ -157,6 +157,30 @@ typedef struct I {
   };
 } I;
 
+enum P_Tag {
+  P0,
+  P1,
+};
+typedef uint8_t P_Tag;
+
+typedef struct P0_Body {
+  uint8_t _0;
+} P0_Body;
+
+typedef struct P1_Body {
+  uint8_t _0;
+  uint8_t _1;
+  uint8_t _2;
+} P1_Body;
+
+typedef struct P {
+  P_Tag tag;
+  union {
+    P0_Body p0;
+    P1_Body p1;
+  };
+} P;
+
 void root(Opaque *opaque,
           A a,
           B b,
@@ -172,4 +196,12 @@ void root(Opaque *opaque,
           L l,
           M m,
           N n,
-          O o);
+          O o,
+          P p);
+
+#include <stddef.h>
+#include "testing-helpers.h"
+static_assert(offsetof(CBINDGEN_STRUCT(P), tag) == 0, "unexpected offset for tag");
+static_assert(offsetof(CBINDGEN_STRUCT(P), p0) == 1, "unexpected offset for p0");
+static_assert(offsetof(CBINDGEN_STRUCT(P), p0) == 1, "unexpected offset for p1");
+static_assert(sizeof(CBINDGEN_STRUCT(P)) == 4, "unexpected size for P");
--- a/tests/expectations/both/enum.compat.c
+++ b/tests/expectations/both/enum.compat.c
@@ -217,6 +217,36 @@ typedef struct I {
   };
 } I;
 
+enum P_Tag
+#ifdef __cplusplus
+  : uint8_t
+#endif // __cplusplus
+ {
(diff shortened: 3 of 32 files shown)


tests/checkers/tag/enum.c
```diff
@@ -0,0 +1,11 @@
+#include "expectations/tag/enum.c"
+
+#include <assert.h>
+
+static_assert(sizeof(struct P) == 4, "unexpected size for P");
```
emilio: Is `static_assert` a C thing nowadays? since when?

emilio: (Oh, I found this btw, TIL that C11 won static_assert, that's great)

emilio:
I opened https://github.com/eqrion/cbindgen/pull/465 with a proposal with a slightly less invasive change which allows to test this as well, and has the side benefit of testing it on all the styles / configurations. WDYT?

sw17ch:
This sounds great. I'm happy to add the trailers (I didn't realize trailers were a supported feature). If you merge #465 with the helper, I'll replace my test infrastructure in this test with trailers in the toml file.

Thank you!

emilio:
Alright, merged that to master, as it was green. Thanks for bearing with me!

sw17ch:
Force-pushed from 9d2b8fa  to 228f7df. This drops the `checkers` tests in favor of the footers static assertions. I also added some uses of `offsetof` to demonstrate where the tags and bodies are expected. I've only done this for the single struct.

Using footers was a much better idea. Thank you for the assistance!

PR #466, opened 2020-01-26
Add option for sorting of function names

### LINKED ISSUE #461: Keep order of exported functions
Hi,

In my project all `extern "C" pub fn`'s are declared in a single file / module.
Is it possible to ensure that the order of the generated exported functions in the header keeps same as in the module?

To me it seems that the functions are ordered by the functions name. Am I correct about this? Is there a way of changing / workaround this?

systec-ad:
Closes #461

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -589,6 +589,13 @@ swift_name_macro = "CF_SWIFT_NAME"
 # default: "None"
 rename_args = "PascalCase"
 
+# This rule specifies if the order of functions will be sorted in some way.
+#
+# "Name": sort by the name of the function
+# "None": keep order in which the functions have been parsed
+#
+# default: "Name"
+sort_by = "None"
 
 [struct]
 # A rule to use to rename struct field names. The renaming assumes the input is
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -201,6 +201,28 @@ impl FromStr for ItemType {
 
 deserialize_enum_str!(ItemType);
 
+/// Type which specifies the sort order of functions
+#[derive(Debug, Clone, PartialEq)]
+pub enum SortKey {
+    Name,
+    None,
+}
+
+impl FromStr for SortKey {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        use self::SortKey::*;
+        Ok(match &*s.to_lowercase() {
+            "name" => Name,
+            "none" => None,
+            _ => return Err(format!("Unrecognized sort option: '{}'.", s)),
+        })
+    }
+}
+
+deserialize_enum_str!(SortKey);
+
 /// Settings to apply when exporting items.
 #[derive(Debug, Clone, Deserialize, Default)]
 #[serde(rename_all = "snake_case")]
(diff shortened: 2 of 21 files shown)


src/bindgen/config.rs
```diff
@@ -293,6 +315,8 @@ pub struct FunctionConfig {
     pub rename_args: Option<RenameRule>,
     /// An optional macro to use when generating Swift function name attributes
     pub swift_name_macro: Option<String>,
+    /// Select sort stradegy for function names
+    pub sort_by: SortStradegy,
```
emilio: You mean `strategy`, right? Not stradegy.

src/bindgen/library.rs
```diff
@@ -55,10 +55,16 @@ impl Library {
 
     pub fn generate(mut self) -> Result<Bindings, Error> {
         self.remove_excluded();
-        self.functions.sort_by(|x, y| x.path.cmp(&y.path));
         self.transfer_annotations();
         self.simplify_standard_types();
 
+        match self.config.function.sort_by {
+            SortStradegy::ByName => {
+                self.functions.sort_by(|x, y| x.path.cmp(&y.path));
```
emilio: I think this can just be `sort_by_key(|x| &x.path)`, while you're at it.

docs.md
```diff
@@ -589,6 +589,13 @@ swift_name_macro = "CF_SWIFT_NAME"
 # default: "None"
 rename_args = "PascalCase"
 
+# This rule specifies if the order of functions will be sorted in some way.
+#
+# "ByName": sort by the name of the function
+# "Keep": keep order in which the functions have been parsed
+#
+# default: "ByName"
```
emilio: A bit bikesheddy, but I think this would be a bit better if the variant was just `Name`. At least to me `sort_by = "Name"` reads better than `sort_by = "ByName"`.

Similarly, mostly for consistency with renaming options, probably the "keep" variant should be `None`.

src/bindgen/config.rs
```diff
@@ -201,6 +201,28 @@ impl FromStr for ItemType {
 
 deserialize_enum_str!(ItemType);
 
+/// Type which specifies the sort order of functions
+#[derive(Debug, Clone, PartialEq)]
+pub enum SortStradegy {
```
emilio: Probably if you agree with the naming suggestion above, this could be:

```
pub enum SortKey {
    Name,
    None,
}
```

Or such.

src/bindgen/config.rs
systec-ad: oh dear! :see_no_evil:

docs.md
systec-ad: You are right. This feels a lot more natural.

src/bindgen/library.rs
systec-ad: Thanks. Learnt something new here.

systec-ad:
I tried to use `sort_by_key()` and ran into lifetime errors.
> error[E0495]: cannot infer an appropriate lifetime for borrow expression due to conflicting requirements
Afaik it would only work with cloning the path. What would you prefer?

emilio:
sort_by seems fine then :)

PR #471, opened 2020-02-11
Emit documentation on constants

### LINKED ISSUE #403: Constant doc strings are not placed into the C header output
Given the following input:

```rust
/// A reasonable approximation of PI
pub const PI: f32 = 3.14159;
```

The following output is generated

```c
#define PI 3.14159
```

I would instead expect the following to be generated:

```c
/**
 * A reasonable approximation of PI
 */
#define PI 3.14159
```

So far I have successfully generated docs for other types, but constants do not seem to have their documentation brought along.

arsing:
Fixes #403

### DIFF
--- a/src/bindgen/ir/constant.rs
+++ b/src/bindgen/ir/constant.rs
@@ -489,6 +489,8 @@ impl Constant {
             _ => &self.value,
         };
 
+        self.documentation.write(config, out);
+
         if config.constant.allow_static_const && config.language == Language::Cxx {
             out.write(if in_body { "inline " } else { "static " });
             if let Type::ConstPtr(..) = self.ty {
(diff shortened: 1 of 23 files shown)


emilio:
Thanks! Can you add a test for this?

emilio:
(just adding docs to any of the constants in `tests/rust/constant.rs` would do)

arsing:
Done. I also missed two existing tests that already had doc comments that are now being emitted.

emilio:
Will wait for CI before merging, but hopefully it'll be green :-)

emilio:
Seems like rustfmt failed to download on CI, but I just ran the tests locally and this is alright. Not worth waiting for a new nightly with rustfmt imho, so I'm going to just merge this.

Thanks again!

arsing:
@emilio Any chance you could release a 0.13.2 with this? It's the only reason we still have to `cargo install --git` for $dayjob. (Also, what's up with that `"` tag and release?)

emilio:
@arsing sure, will do. I don't know about the `''` tag, I noticed when I published the last version and that day github was being flaky so I suspect it was just a GH fluke.

PR #473, opened 2020-02-15
bindgen: enable "void prototype" while writing struct field

### LINKED ISSUE #470: Fix generated typedefs for function types without arguments
For type 
`pub type Function = fn() -> usize;`
generated C typedef is
`typedef uintptr_t (*Function)();`
but expected according to C standard is
`typedef uintptr_t (*Function)(void);`
It implies warning when compiling C code:
`warning: function declaration isn't a prototype`

TheKK:
Fix #470 

This `void` in some C++ files should be fine according to standard ([ref][stackoverflow ref])

Let's me know if there's anything I can improve :)

[stackoverflow ref]: https://softwareengineering.stackexchange.com/a/287002

### DIFF
--- a/src/bindgen/cdecl.rs
+++ b/src/bindgen/cdecl.rs
@@ -286,7 +286,7 @@ pub fn write_func<F: Write>(
 }
 
 pub fn write_field<F: Write>(out: &mut SourceWriter<F>, t: &Type, ident: &str) {
-    CDecl::from_type(t).write(out, Some(ident), false);
+    CDecl::from_type(t).write(out, Some(ident), true);
 }
 
 pub fn write_type<F: Write>(out: &mut SourceWriter<F>, t: &Type) {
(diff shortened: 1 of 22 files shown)


src/bindgen/cdecl.rs
```diff
@@ -286,7 +286,7 @@ pub fn write_func<F: Write>(
 }
 
 pub fn write_field<F: Write>(out: &mut SourceWriter<F>, t: &Type, ident: &str) {
-    CDecl::from_type(t).write(out, Some(ident), false);
+    CDecl::from_type(t).write(out, Some(ident), true);
```
emilio: In other places we intentionally _don't_ do this for C++, so I think we should be consistent here.

Most of the callers if not all should have a `config` object. I think we should pass it through and always write the `void` when the language is "C", wdyt?

TheKK: That's a good idea, I'll apply these change later.

TheKK:
Looks like CI failed with some irrelevant issues (it didn't even start running tests), could we rerun it again?

emilio:
I cleared caches (builds just time out when they get too big) and restarted the build. Thanks!

emilio:
Thanks a lot for the fix!

TheKK:
It's my pleasure as well!

PR #475, opened 2020-02-22
Only write `default` cases if the switch is not exhaustive.

jdashg:
This fixes C++ -Wcovered-switch-default warnings.


jdashg:
This still needs tests, but is the general idea acceptable?

emilio:
LGTM, but could you add a test?

emilio:
(I'll just add them myself now, it is easy enough :))

PR #477, opened 2020-02-24
Omit or display more detailed warning message while trying to generate declaration from Rust functions

### LINKED ISSUE #467: Don't warn if both #[no_mangle] and extern "C" are missing
I have a number of functions that I don't want exported through a C ABI but I'm getting warnings. This clutters the output and makes it more difficult to see real issues. I suggest that if a function is missing both #[no_mangle] and export "C" then the code author intends that the function should not be exported and there should be no warning. If just one or the other is missing then that is an actual issue.

TheKK:
fix #467

I'd like to hear your opinion about my implementation.

- are these warning messages okay in each case?
- is the "match in if-else statement" okay? or should I write it in a single match statement?

Please let me know what's the suitable way to go :)

### DIFF
--- a/src/bindgen/parser.rs
+++ b/src/bindgen/parser.rs
@@ -637,8 +637,11 @@ impl Parse {
         };
 
         if let syn::Visibility::Public(_) = vis {
-            if sig.abi.is_omitted() || sig.abi.is_c() {
-                if let Some(exported_name) = named_symbol.exported_name() {
+            let is_extern_c = sig.abi.is_omitted() || sig.abi.is_c();
+            let exported_name = named_symbol.exported_name();
+
+            if is_extern_c {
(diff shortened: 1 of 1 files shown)


src/bindgen/parser.rs
```diff
@@ -659,9 +659,10 @@ impl Parse {
                         loggable_item_name()
                     );
                 }
-                (false, _) => {
+                (false, Some(_exported_name)) => {
```
emilio: I would just use `(_)` or `(..)`, fwiw. But this works for me as well

Barugon:
Thanks for doing this @TheKK.

src/bindgen/parser.rs
TheKK: I can change it to `_` next time :)

PR #479, opened 2020-02-25
ir: write declaration of global variable like struct fields

### LINKED ISSUE #476: Wrong output for static arrays
```rust
#[no_mangle]
pub static mut MCRT_ERROR_TEXT: [c_char; 128] = [0; 128];
```
outputs this
```c
extern char[128] MCRT_ERROR_TEXT;
```
when it should be this:
```c
extern char MCRT_ERROR_TEXT[128];
```

TheKK:
fix #476 

Just wondering, should I use `Source::write of (String, Type)` or `cdecl::write_field` directly?

### DIFF
--- a/src/bindgen/ir/global.rs
+++ b/src/bindgen/ir/global.rs
@@ -6,6 +6,7 @@ use std::io::Write;
 
 use syn;
 
+use crate::bindgen::cdecl;
 use crate::bindgen::config::Config;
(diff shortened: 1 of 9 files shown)


emilio:
Please use `write_field` directly, that avoids having to allocate the new string, plus it's a bit nicer on its own IMHO.

TheKK:
Fixed!

PR #482, opened 2020-03-03
constant: Fix the interaction of allow_constexpr and inline consts.

emilio:


emilio:
err, that was somewhat unintended, actually... Oh well.

emilio:
/me thought he was on a branch, but laptop was on master.

AlaskanEmily:
It seems a little weird to me that Foo::CONSTANT_I64_BODY isn't just totally defined inside the body of the struct to me. I know there are different rules about symbols being generated for some constexpr types (like strings/char arrays), but even then shouldn't it have constexpr on the initial declaration?

emilio:
That would be lovely, but it doesn't work for structs of the same type, which is the primary use of this feature (for bitfields)

PR #483, opened 2020-03-05
Escape type name and field name of generated tagged union

### LINKED ISSUE #368: Escape `Double`
```irc
> @emilio> gandalf: ah, `double` is a C++ keyword, use `Double_` and file a cbindgen issue to escape that?
```

TheKK:
Should fix  #368
I escape the `enum Tag` as well since they are quite relevant.

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -17,6 +17,7 @@ use crate::bindgen::library::Library;
 use crate::bindgen::mangle;
 use crate::bindgen::monomorph::Monomorphs;
 use crate::bindgen::rename::{IdentifierType, RenameRule};
+use crate::bindgen::reserved;
 use crate::bindgen::utilities::find_first_some;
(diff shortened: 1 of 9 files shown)


src/bindgen/ir/enumeration.rs
```diff
@@ -208,9 +209,15 @@ impl EnumVariant {
 impl Source for EnumVariant {
     fn write<F: Write>(&self, config: &Config, out: &mut SourceWriter<F>) {
         let condition = self.cfg.to_condition(config);
+        let escaped_export_name = {
+            let mut name = self.export_name.to_owned();
+            reserved::escape(&mut name);
```
emilio: `write()` is not the right place to do this. This should be done in `rename_for_config` for consistency with the other callers.

TheKK: Oh I see, never know about the `Itme` trait. I'll change it!

TheKK:
I've committed changes, please take a look :)

PR #490, opened 2020-03-10
support large integer constants

devietti:
I was using cbindgen for a project recently where I needed to export a `const` from Rust that was larger than i64, which would cause `gcc` to complain that `warning: integer constant is so large that it is unsigned`. This small patch includes a `ULL` suffix on the constant in the generated code to avoid this issue.


src/bindgen/ir/constant.rs
```diff
@@ -163,7 +163,11 @@ impl Literal {
                         other_code => format!(r"L'\U{:08X}'", other_code),
                     })),
                     syn::Lit::Int(ref value) => {
-                        Ok(Literal::Expr(value.base10_digits().to_string()))
+                        if value.base10_parse::<i64>().is_err() {
+                            Ok(Literal::Expr(format!("{}ULL", value.base10_digits())))
```
emilio: Hmm, what if it is negative? Do we need just `LL` instead?

devietti: Good question - per the C/C++ spec (https://en.cppreference.com/w/cpp/language/integer_literal, the _Notes_ section):
> There are no negative integer literals. Expressions such as -1 apply the unary minus operator to the value represented by the literal, which may involve implicit type conversions.

So to export a constant that has the value of `i64::min_value()`, the ULL suffix is needed because the magnitude of the value is too large to fit into a _signed_ 64-bit representation. I added this to the `constant_big.rs` test case.

devietti:
This also got me wondering about exporting `i128`/`u128` constants, but those don't seem to be exported at the moment, perhaps because of issues like https://github.com/rust-lang/rust/issues/54341.

src/bindgen/ir/constant.rs
emilio: Cool, that makes sense. Thanks for the extra test!

PR #491, opened 2020-03-11
parser: escape Rust function name from "r#" (raw identifier)

### LINKED ISSUE #410: Raw identifier is not generated correctly
The following Rust code uses [raw identifier](https://doc.rust-lang.org/stable/edition-guide/rust-2018/module-system/raw-identifiers.html), and the generated code preserved the raw identifier syntax which isn't valid syntax for C/C++:

```rust
#[no_mangle]
pub extern "C" fn r#match() {}
```

With the example [cbindgen.toml](https://raw.githubusercontent.com/eqrion/cbindgen/master/template.toml), the following C++ code is generated:

```cpp
#include <cstdarg>
#include <cstdint>
#include <cstdlib>
#include <new>


extern "C" {

void r#match();

} // extern "C"
```

Desired output:

```cpp
#include <cstdarg>
#include <cstdint>
#include <cstdlib>
#include <new>


extern "C" {

void match();

} // extern "C"
```

TheKK:
Should fix #410.

Wondering if there's any better word than "escape"? "trim" perhaps?

### DIFF
--- a/src/bindgen/parser.rs
+++ b/src/bindgen/parser.rs
@@ -541,7 +541,7 @@ impl Parse {
                     );
                     return;
                 }
-                let path = Path::new(function.sig.ident.to_string());
+                let path = Path::new(escape_raw_identifier(function.sig.ident.to_string()));
                 match Function::load(path, None, &function.sig, true, &function.attrs, mod_cfg) {
                     Ok(func) => {
                         info!("Take {}::{}.", crate_name, &function.sig.ident);
@@ -643,7 +643,7 @@ impl Parse {
         if let syn::Visibility::Public(_) = vis {
             match (is_extern_c, exported_name) {
                 (true, Some(exported_name)) => {
-                    let path = Path::new(exported_name);
+                    let path = Path::new(escape_raw_identifier(exported_name));
                     match Function::load(path, self_type, &sig, false, &attrs, mod_cfg) {
                         Ok(func) => {
                             info!("Take {}.", loggable_item_name());
@@ -965,5 +965,14 @@ impl Parse {
         // fine to just do it here instead of deferring it like we do with the
         // other calls to this function.
         self.load_syn_assoc_consts_from_impl(crate_name, mod_cfg, &impl_);
+    }
+}
+
+/// Remove leading "r#" if there's one since "r#" is invalid in generated language.
+fn escape_raw_identifier(i: String) -> String {
+    if i.starts_with("r#") {
+        i[2..].to_owned()
+    } else {
+        i
     }
 }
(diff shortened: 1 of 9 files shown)


src/bindgen/parser.rs
```diff
@@ -541,7 +541,7 @@ impl Parse {
                     );
                     return;
                 }
-                let path = Path::new(function.sig.ident.to_string());
+                let path = Path::new(escape_raw_identifier(function.sig.ident.to_string()));
```
emilio: Maybe this should be done in `Path::new`? Maybe somewhere else?

This is an issue for fields and everything though...

TheKK: I believe we can put them into `Function::load` in this context at least.
Would you mind telling me more about `Path`? It's not very clear to me what it represent :)

emilio: Hmm, that's still not the right thing I fear. We kinda need it for all the identifiers we read out from syn...

`Path` is basically a name and a set of generics that represent a struct. I think the idea why it's a `Path` is in case we eventually supported modules or such (so `foo::bar::Baz` would be a path for example). I think right now they're only identifiers.

TheKK: > ... I think right now they're only identifiers.

In that case, putting raw identifier stripping into `Path::new` sounds reasonable. I'll write more tests case since now this PR affect more than function identifier.

emilio: Right, Path are identifiers at the moment, but there's a lot of identifiers that aren't paths (such as field names).

I think we should try to do it at the syn level. Maye add an `to_unscaped_string` to `syn::Identifier` in an extension trait, and use it across the codebase?

TheKK: Just found that `syn` has already done that for us, [syn::ext::IdentExt][link].

I'll check usages of `syn::Ident::to_string` and see what's the better strategy to apply `unwraw`to them.

[link]: https://docs.rs/syn/1.0.16/syn/ext/trait.IdentExt.html#tymethod.unraw

emilio: Yeah, using unraw sounds great!

TheKK: What's your opinion about creating a new constructor for `Path` which like `Path::new_from_ident(&syn::Ident) -> Path`? It could make the (current) relation between `Path` and `syn::Ident` more clear, also reduce the change which forgetting to use `unraw`.

emilio: Ah, sorry for missing this. That looks ok though probably it's not needed. Just calling `unraw()` at the right places should do.

TheKK: ok，let me modify them

Hywan:
Any news from this PR :-)?

TheKK:
Sorry that I kinda busy these days :(
Should I close this PR?

emilio:
No worries! Let's close for now, someone (or you, when you are less busy) can resubmit with comments addressed or what not. Thanks a lot @TheKK!

PR #496, opened 2020-03-22
Add missing cpp_compat documentation.

RazrFalcon:


docs.md
```diff
@@ -376,6 +376,21 @@ includes = ["my_great_lib.h"]
 # default: false
 no_includes = false
 
+# Whether to make a C header C++ compatible.
+# These will wrap generated functions into a `extern "C"` block, e.g.
```
emilio: It also has effects on enums / enum classes and such. But this is a good start too.

Issue #499, opened 2020-03-28
Tests should not generate files in the tests directory


sylvestre:
TMPDIR is probably a better place than
/usr/share/cargo/registry/cbindgen-0.13.2/tests/expectations/typedef.compat.o (as example

Note that this is failing in Debian pacakges test CI:
https://ci.debian.net/packages/r/rust-cbindgen/

Example:
```
---- test_typedef stdout ----
Running: "/tmp/tmp.DS1uKzckhx/target/x86_64-unknown-linux-gnu/debug/cbindgen" "--lang" "c" "--cpp-compat" "--style" "type" "-o" "/usr/share/cargo/registry/cbindgen-0.13.2/tests/expectations/typedef.compat.c" "/usr/share/cargo/registry/cbindgen-0.13.2/tests/rust/typedef.rs"
Running: "gcc" "-D" "DEFINED" "-c" "/usr/share/cargo/registry/cbindgen-0.13.2/tests/expectations/typedef.compat.c" "-o" "/usr/share/cargo/registry/cbindgen-0.13.2/tests/expectations/typedef.compat.o" "-I" "/usr/share/cargo/registry/cbindgen-0.13.2/tests" "-Wall" "-Werror" "-Wno-attributes" "-g" "-O2" "-fdebug-prefix-map=/usr/share/cargo/registry/cbindgen-0.13.2=." "-fstack-protector-strong" "-Wformat" "-Werror=format-security" "-D" "CBINDGEN_STYLE_TYPE"
thread 'test_typedef' panicked at 'Output failed to compile: Output { status: ExitStatus(ExitStatus(256)), stdout: "", stderr: "Assembler messages:\nFatal error: can\'t create /usr/share/cargo/registry/cbindgen-0.13.2/tests/expectations/typedef.compat.o: Permission denied\n" }', tests/tests.rs:106:5
stack backtrace:
   0: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
   1: core::fmt::write
   2: std::io::Write::write_fmt
   3: std::io::impls::<impl std::io::Write for alloc::boxed::Box<W>>::write_fmt
   4: std::panicking::default_hook::{{closure}}
   5: std::panicking::default_hook
   6: std::panicking::rust_panic_with_hook
   7: std::panicking::continue_panic_fmt
   8: std::panicking::begin_panic_fmt
   9: tests::compile
             at tests/tests.rs:106
  10: tests::run_compile_test
[...]

Issue #437, opened 2019-12-20
Possible to remove the `_` in mangle_path for C?


adamski:
We have a generic type being exposed to C. Mangle path adds an underscore between the types. Is it possible to configure this, as it messes with the project's code style :)

emilio:
Well, I guess we could add options to configure it, but off-hand you need _some_ sort of mangling scheme.

emilio:
I guess a minimal approach to do this would be to add a new configuration option like `mangling_separator` or something, which is a string. That'd allow you to use the empty string to remove them all.

If that seems reasonable it shouldn't be hard to implement...

adamski:
That sounds good to me! I'm up for having a go at a PR for that.

emilio:
Sure, please go for it :)

emilio:
This was fixed in #502.

PR #502, opened 2020-04-01
Add option for `mangling_separator` in ExportConfig

adamski:
Addresses #437 

I'm not sure if `mangling_separator` should live in ExportConfig or elsewhere. 
Also wondering if there is a neater way to pass `mangling_separator` down the call chain.

Tested with a `cpp_compat` config, without the new option and with. 

Not able to run tests yet due to #499

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -227,7 +227,7 @@ impl EnumVariant {
 
     fn specialize(&self, generic_values: &[Type], mappings: &[(&Path, &Type)], config: &Config) -> Self {
         Self::new(
-            mangle::mangle_name(&self.name, generic_values, config.export.mangling_separator.as_deref()),
+            mangle::mangle_name(&self.name, generic_values, config.export.mangle_separator.as_deref()),
             self.discriminant,
             self.body.specialize(generic_values, mappings, config),
             self.cfg.clone(),
@@ -547,7 +547,7 @@ impl Item for Enum {
         let mangled_path = mangle::mangle_path(
             &self.path,
             generic_values,
-            library.get_config().export.mangling_separator.as_deref());
+            library.get_config().export.mangle_separator.as_deref());
 
         let monomorph = Enum::new(
             mangled_path,
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -51,7 +51,12 @@ impl VariantBody {
         }
     }
 
-    fn specialize(&self, generic_values: &[Type], mappings: &[(&Path, &Type)], config: &Config) -> Self {
+    fn specialize(
+        &self,
+        generic_values: &[Type],
+        mappings: &[(&Path, &Type)],
+        config: &Config,
+    ) -> Self {
         match *self {
             Self::Empty(ref annos) => Self::Empty(annos.clone()),
             Self::Body { ref name, ref body } => Self::Body {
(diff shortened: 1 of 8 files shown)


src/bindgen/ir/enumeration.rs
```diff
@@ -181,13 +181,13 @@ impl EnumVariant {
         }
     }
 
-    fn specialize(&self, generic_values: &[Type], mappings: &[(&Path, &Type)]) -> Self {
+    fn specialize(&self, generic_values: &[Type], mappings: &[(&Path, &Type)], mangling_separator: &Option<String>) -> Self {
```
emilio, line 184: Passing an `Option<&str>` is a bit more idiomatic, is it much more difficult?

Alternatively if passing the whole `&Config` down here is sensible I think that'd be slightly preferable, but no big deal.

src/bindgen/mangle.rs
```diff
@@ -4,31 +4,35 @@
...
 
-fn internal_mangle_name(name: &str, generic_values: &[Type], last_in_parent: bool) -> String {
+fn internal_mangle_name(name: &str, generic_values: &[Type], last_in_parent: bool, mangle_separator: &Option<String>) -> String {
     if generic_values.is_empty() {
         return name.to_owned();
     }
 
     let mut mangled = name.to_owned();
+    let separator = match mangle_separator {
+        Some(s) => s.to_owned(),
```
emilio, line 28: If this was an `Option<&str>`, then you don't need to heap-allocate here.

You don't need here anyway though, this could be:

```
match mangle_separator {
    Some(s) => &**s,
    None => "_",
}
```

src/bindgen/mangle.rs
```diff
@@ -4,31 +4,35 @@
...
+        None => "_".to_string()
+    };
 
-    mangled.push_str("_"); // <
+    mangled.push_str(separator.as_str()); // <
     for (i, ty) in generic_values.iter().enumerate() {
         if i != 0 {
-            mangled.push_str("__"); // ,
+            mangled.push_str(&concat_separators(&separator, 2)); // ,
```
emilio, line 35: Having a temporary string here is a bit unfortunate, just:

```
for _ in 0..2 {
    mangled.push_str(separator);
}
```

Or alternatively, something like:

```
fn extend_n(mangled: &mut String, separator: &str, n: usize) {
    // ...
}
```

adamski:
@eqrion Thanks for the review. I passed Config instead, seems to make more sense also to be able to address #503 
I still can't test due to #505

src/bindgen/mangle.rs
```diff
@@ -43,16 +49,15 @@ impl<'a> Mangler<'a> {
...
-        self.output.extend(std::iter::repeat(separator).take(count));
+        self.output.extend(std::iter::repeat(self.mangle_separator).take(count));
     }
 
     fn append_mangled_type(&mut self, ty: &Type, last: bool) {
         match *ty {
             Type::Path(ref generic) => {
                 let sub_path =
-                    Mangler::new(generic.export_name(), generic.generics(), last).mangle();
+                    Mangler::new(generic.export_name(), generic.generics(), last, &Some(self.mangle_separator.to_owned())).mangle();
```
emilio, line 60: This should not need to allocate, The function should take an `Option<&'a str>` instead, and this should just be `Some(self.mangle_separator)`.

src/bindgen/mangle.rs
```diff
@@ -98,8 +103,8 @@ impl<'a> Mangler<'a> {
         if !self.last {
             self.push(Separator::ClosingAngleBracket)
         }
+        }
```
emilio, line 106: This change looks off, please run rustfmt?

src/bindgen/mangle.rs
adamski: Agreed, I tried to make it all `Option<&str>` but it got tricky since `Config::export` stores `Option<String>`. I now see that that we can convert to `Option<&str>` with `Option::as_deref` so I'll get that sorted.

src/bindgen/mangle.rs
adamski: Sorted 0f73c70

src/bindgen/mangle.rs
adamski: Fixed with 7f5b606 (plus some tidy ups)

adamski:
> Thanks, this looks good. This needs to get rustfmt'd to pass CI, and also, can we get a test for this? Let me know if you need help adding one.

I'll run rustfmt and push. I already added a couple of tests for it. Please see `mangle.rs`

emilio (approved):
I meant more on the line of integration tests. Those are usually preferred, but fair enough :)

emilio:
https://travis-ci.org/github/eqrion/cbindgen/builds/719002633 looks green :)

PR #507, opened 2020-04-08
Fix dependencies spec and add CI testing with -Zminimal-versions

asomers:


Cargo.toml
```diff
@@ -17,15 +17,15 @@ travis-ci = { repository = "eqrion/cbindgen" }
 [dependencies]
 clap = { version = "2", optional = true }
 log = "0.4"
-serde = { version = "1.0", default-features = false, features = ["derive"]}
+serde = { version = "1.0.103", default-features = false, features = ["derive"]}
```
emilio, line 20: Can you elaborate? Why do we need this change?

asomers: As-is, the dependencies are incorrectly specified.  If you do `cargo +nightly update -Zminimal-versions && cargo +nightly build`, the build will fail.  Worse, any crate that depends on cbindgen will also fail to build if it tries building with its minimal dependencies.

emilio:
Thanks!

emilio:
https://travis-ci.org/github/eqrion/cbindgen/builds/672741687 is green (not sure what's up with GitHub's integration).

Issue #510, opened 2020-04-11
Add `#pragma once` guard


igor-sadchenko:
it would be nice to be able to use `#pragma once` guard.
This guard is a non-standard, but supported in most compilers.
https://en.wikipedia.org/wiki/Pragma_once

igor-sadchenko:
Pull-request #511 with `#pragma once` guard was merged.

PR #511, opened 2020-04-11
Added pragma once option. Issue #510

igor-sadchenko:
Added `#pragma once`guard:
- Added option `pragma_once` to config
- Added description to [docs.md](docs.md)
- Added tests
- Fix tests to compile header files


tests/tests.rs
```diff
@@ -154,9 +155,44 @@ fn run_compile_test(
...
+        if is_header {
+            &name[0..header_position.unwrap()]
+        } else {
+            name
+        },
+        &ext
+    );
+    output.push(&source_file);
+
+    if is_header {
```
emilio, line 173: Why are all these changes needed?

tests/rust/pragma_once.header.toml
```diff
@@ -0,0 +1,2 @@
+pragma_once = true
+no_includes = true
```
emilio, line 2: Why is `no_includes` needed? It'd be nice to test the relative ordering of `#pragma once` and `#include`s.

tests/tests.rs
```diff
@@ -154,9 +155,44 @@ fn run_compile_test(
...
+        let header_ext = match output.extension().unwrap().to_str().unwrap().as_ref() {
+            "cpp" => "hpp",
+            "c" => "h",
+            _ => "",
```
emilio, line 177: May be better to panic here or something, or just default to `.h`.

tests/tests.rs
emilio: If they _are_ needed, it may be better to generate the `.c` / `.cpp` file in the tmpdir, rather than checking it in the repo...

igor-sadchenko: File with `#pragma once` compile with warning that is treated an error
```
error: #pragma once in main file [-Werror]
    1 | #pragma once
      |         ^~~~
cc1: all warnings being treated as errors
```

tests/rust/pragma_once.header.toml
igor-sadchenko: I did this following the include_guard option.
Removed `no_includes`

tests/tests.rs
igor-sadchenko: ok

igor-sadchenko:
> So you can achieve this trivially with:
> 
> ```
> header = "#pragma once"
> ```
> 
> in the config, right? Anyhow not really opposed to this, just a bit concerned / confused about all the testing changes and why are they needed.

Yes, you can use `header`, but I think more convenient to have separate option for `#pragma once`.
I explain why change tests in this comment https://github.com/eqrion/cbindgen/pull/511#discussion_r407263651

tests/tests.rs
emilio: Ok, instead of these complicated changes, can we just avoid passing `-Werror` for this test?

We can either have a whitelist in the tests.rs file, or some kind of hint in the file name / contents, not sure what's better.

emilio: (Ideally we'd just pass [`-Wno-pragma-outside-header`](https://clang.llvm.org/docs/DiagnosticsReference.html#wpragma-once-outside-header) and forget about this, but it seems gcc doesn't support this)

igor-sadchenko: Ok. I changed this.

PR #512, opened 2020-04-15
Fix #254

### LINKED ISSUE #254: Parsing crate `foo`: can't find mod `bar` with edition 2018 submodules.
```
$ rustup default beta
$ cargo new --lib foo
$ cd foo
$ echo "mod qux;" > src/lib.rs
$ echo "mod bar;" > src/qux.rs
$ mkdir src/qux
$ touch src/qux/bar.rs
$ cargo build
   Compiling foo v0.1.0 (/tmp/foo)                                                                                                                                       
    Finished dev [unoptimized + debuginfo] target(s) in 0.33s
$ cbindgen
WARN: Parsing crate `foo`: can't find mod bar`.
#include <cstdint>
#include <cstdlib>
```

vtavernier:
By checking the recursion depth in the crate we are currently parsing, we can decide on the right module directory to find submodules in. The added `mod_2018` test fails without the changes to `parser.rs`.


src/bindgen/parser.rs
```diff
@@ -238,6 +238,10 @@ impl<'a> Parser<'a> {
     }
 
     fn parse_mod(&mut self, pkg: &PackageRef, mod_path: &FilePath) -> Result<(), Error> {
+        self.parse_mod_depth(pkg, mod_path, 0)
```
emilio, line 241: Can we just pass zero on the relevant callers and keep `parse_mod` as the name for all functions?

src/bindgen/parser.rs
```diff
@@ -265,16 +269,25 @@ impl<'a> Parser<'a> {
...
         };
 
-        let mod_dir = mod_path.parent().unwrap();
+        // Compute module directory according to Rust 2018 rules
+        let mod_dir_2018;
+
+        let mod_dir = if depth == 0 {
+            mod_path.parent().unwrap()
+        } else {
+            mod_dir_2018 = mod_path.parent().unwrap().join(mod_path.file_stem().unwrap());
```
emilio, line 278: Does this break non-2018 users?

emilio:
Thanks so much for working on this btw!

src/bindgen/parser.rs
vtavernier: I actually hesitated between the two options, fixed in 0d32d99

src/bindgen/parser.rs
vtavernier: It seems to break non-2018 unfortunately, I added a test for that in 0d89c9e and I'll investigate this in more detail.

vtavernier: Should be fixed by 835229b

src/bindgen/parser.rs
```diff
@@ -265,16 +265,25 @@ impl<'a> Parser<'a> {
             self.cache_src.get(&owned_mod_path).unwrap().clone()
         };
 
-        let mod_dir = mod_path.parent().unwrap();
+        // Compute module directory according to Rust 2018 rules
+        let mod_dir_2018;
 
-        self.process_mod(pkg, mod_dir, &mod_parsed)
+        let mod_dir = if depth == 0 || mod_path.ends_with("mod.rs") {
```
emilio, line 271: Hmm, I still don't think this is quite sound is it? At the very least we should check the file name for full equality for `mod.rs`, not just whether the path ends with `mod.rs`

vtavernier: Apparently `Path::ends_with` only checks entire path components so it is checking if the (entire) file name is `mod.rs`. This wouldn't be the case with a `&str`.

https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.ends_with

emilio:
Sorry for the lag here, taking this patch for a spin now :)

PR #514, opened 2020-04-17
Fix #418: Use heck for case conversion

### LINKED ISSUE #418: Use heck for case conversion
https://github.com/withoutboats/heck

vtavernier:
All case conversion cases seem to be covered by this crate so this greatly simplifies `rename.rs`.


emilio:
Thank you!

Issue #521, opened 2020-05-05
char literal as u8 not supported in bitflags


Delamare2112:
The following fails and gives the following warning:
```rust
bitflags! {
    pub struct Thing: u8 {
        const READY = 'A' as u8;
    }
}
```
Gives: `WARN: Skip dutara_types::READY - (Unsupported expression. Cast(ExprCast { attrs: [], expr: Lit(ExprLit { attrs: [], lit: Char(LitChar { token: 'A' }) }), as_token: As, ty: Path(TypePath { qself: None, path: Path { leading_colon: None, segments: [PathSegment { ident: Ident(u8), arguments: None }] } }) }))`  

Replacing `'A' as u8` with `65` works fine.  

I think it would be really neat if this was supported :)

emilio:
You can probably use byte literals instead, so `const READY = b'A'`, which is the right thing to do anyway. Supporting casts may be doable though. I don't plan to work on it right away, but it should be relatively straight-forward. See `src/bindgen/ir/constant.rs`.

This is a problem for all constants, the following won't generate `C`:

```rust
pub const A: u8 = 65;

pub const B: u8 = b'A';

pub const C: u8 = 'A' as u8;
```

PR #526, opened 2020-05-13
Support casts in constants

neurosie:
Addresses #521.

### DIFF
--- a/src/bindgen/ir/constant.rs
+++ b/src/bindgen/ir/constant.rs
@@ -37,6 +37,10 @@ pub enum Literal {
         export_name: String,
         fields: HashMap<String, Literal>,
     },
+    Cast {
+        ty: Type,
+        value: Box<Literal>,
+    },
 }
 
 impl Literal {
@@ -45,7 +49,8 @@ impl Literal {
             Literal::PostfixUnaryOp { .. }
             | Literal::BinOp { .. }
             | Literal::Expr(..)
-            | Literal::Path(..) => {}
+            | Literal::Path(..)
+            | Literal::Cast { .. } => {}
             Literal::Struct {
                 ref mut path,
                 ref mut export_name,
@@ -72,6 +77,7 @@ impl Literal {
                 ..
             } => left.is_valid(bindings) && right.is_valid(bindings),
             Literal::Struct { ref path, .. } => bindings.struct_exists(path),
+            Literal::Cast { ref value, .. } => value.is_valid(bindings),
         }
     }
 }
@@ -104,6 +110,9 @@ impl Literal {
                 right.rename_for_config(config);
             }
             Literal::Expr(_) => {}
+            Literal::Cast { ref mut value, .. } => {
+                value.rename_for_config(config);
+            }
         }
     }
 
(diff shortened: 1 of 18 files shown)


src/bindgen/ir/constant.rs
```diff
@@ -45,7 +49,8 @@ impl Literal {
             Literal::PostfixUnaryOp { .. }
             | Literal::BinOp { .. }
             | Literal::Expr(..)
-            | Literal::Path(..) => {}
+            | Literal::Path(..)
+            | Literal::Cast { .. } => {}
```
emilio, line 53: This needs to call `replace_self_with` in the subexpression, but this is also an issue with other literals.

src/bindgen/ir/constant.rs
```diff
@@ -104,6 +110,9 @@ impl Literal {
                 right.rename_for_config(config);
             }
             Literal::Expr(_) => {}
+            Literal::Cast { ref mut value, .. } => {
```
emilio, line 113: The type should also be renamed to account for the configuration. This should be trivial to test.

src/bindgen/ir/constant.rs
```diff
@@ -252,6 +274,12 @@ impl Literal {
                 right.write(config, out);
                 write!(out, ")");
             }
+            Literal::Cast { ref ty, ref value } => {
```
emilio, line 277: We should probably wrap the whole expression in parens.

neurosie:
Something like `const FOO: *mut Struct = 0 as usize as *mut Struct;` is not currently handled, because it's not a primitive or pointer to one (decided in code [here](https://github.com/eqrion/cbindgen/blob/5c7a7fca4233c7cb73038c58599cb1938c8219ad/src/bindgen/ir/constant.rs#L303)).

However, I did find a problem with something like this:
```
struct A;
pub const B: i32 = 0 as *const A as i32;
```

This gets reordered to
```
static const int32_t B = (int32_t)(const A*)0;
struct A;
```
which fails to compile. It gets reordered because [primitive or pointer to primitive consts are always printed first](https://github.com/eqrion/cbindgen/blob/5c7a7fca4233c7cb73038c58599cb1938c8219ad/src/bindgen/bindings.rs#L218-L224). This is kind of a weird edge case, do you think it's worth handling?

emilio:
Yes, we should handle it somehow, either bailing out or handling it correctly. Seems like "is a primitive or pointer to it" is not enough for these. Probably we really want something like "uses non-primitive types" (where pointer is considered a "primitive" type), and check that.

src/bindgen/ir/constant.rs
neurosie: Casts have pretty high precedence (https://en.cppreference.com/w/c/language/operator_precedence), so I don't think that's necessary. I can add it if you really want though.

src/bindgen/ir/constant.rs
```diff
@@ -72,6 +91,24 @@ impl Literal {
...
+    pub fn uses_only_primitive_types(&self) -> bool {
+        match self {
+            Literal::Expr(..) => true,
+            Literal::Path(..) => true,
+            Literal::PostfixUnaryOp { ref value, .. } => value.uses_only_primitive_types(),
+            Literal::BinOp {
+                ref left,
+                ref right,
+                ..
+            } => left.uses_only_primitive_types() & right.uses_only_primitive_types(),
```
emilio, line 107: Any reason this is using `&` instead of `&&`?

src/bindgen/ir/constant.rs
emilio: Macros that expand to complex stuff without parens always make me a bit nervous, but you're probably right, we can always change this if needed.

src/bindgen/ir/constant.rs
emilio: I assume not, will fix this as a follow-up.

PR #534, opened 2020-06-11
README: add etesync-rs to the example usages

tasn:


tasn:
P.S, I created etesync-rs, thanks a lot for creating cbindgen and thus making etesync-rs possible! :)

PR #536, opened 2020-06-15
Fixes to support Rust 1.32

jasonish:
- travis-ci: add Rust 1.32 to the build matrix
- Don't use enum variants on type aliases

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,8 @@ addons:
 script:
   - export CC=gcc-8
   - export CXX=g++-8
-  - cargo fmt --all -- --check
+  - echo $TRAVIS_RUST_VERSION
+  - test "$TRAVIS_RUST_VERSION" = "nightly" && cargo fmt --all -- --check || true
   - cargo build --verbose
   - cargo build --verbose --no-default-features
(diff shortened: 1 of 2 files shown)


emilio:
Ah, the new build is failing, we should probably only run `cargo fmt --check` on nightly: https://travis-ci.org/github/eqrion/cbindgen/jobs/698689169

jasonish:
> Ah, the new build is failing, we should probably only run `cargo fmt --check` on nightly: https://travis-ci.org/github/eqrion/cbindgen/jobs/698689169

Made this fix. However the tests still fail on the older compiler.  Not sure if/when I'll get around to fixing that.

emilio:
Ugh, so...

 * The tests with Cargo.toml v2 format obviously fail. Not sure how we'd mark that as expected.
 * The tests for expand require nightly, see #38

petrochenkov:
@emilio 
1.32 is old at this point and dependencies (`remove_dir_all-0.5.3`) require 1.40 already.

emilio:
Yeah given there are conflicts with this patch and the above, let's close this for now.

PR #540, opened 2020-07-07
Do not accept array as function arguments

### LINKED ISSUE #528: cbindgen should not map [type; N] function args
Current cbindgen has the faulty map:

``` rust
f(a: [t; 3]) -> ();
```
``` rust
void f(t a[3]);
```

I made a tiny [repo](https://github.com/lu-zero/capi-testcase) to see the problem.

Luni-4:
This PR fixes #528 


tests/rust/function_args.toml
```diff
@@ -0,0 +1 @@
+language = "C"
```
emilio, line 1: This should work regardless of the language, right? I think this should be removed, that way we test both the C and C++ codepath.

Luni-4: Right!

PR #545, opened 2020-07-17
Add an attribute to tell cbindgen how to present pointers

### LINKED ISSUE #542: Add an attribute to tell cbindgen how to present pointers
In C you may represent pointer arguments using `Type *arg`, `Type arg[]` and `Type arg[N]`. 

Would be nice having something along the lines of
``` rust
fn fun(arg: *const Type /* cbindgen:as [3] */)
```
and produce
``` c
void fun(Type arg[3])
```

Luni-4:
This PR fixes #542 

### DIFF
--- a/src/bindgen/cdecl.rs
+++ b/src/bindgen/cdecl.rs
@@ -5,7 +5,7 @@
 use std::io::Write;
 
 use crate::bindgen::declarationtyperesolver::DeclarationType;
-use crate::bindgen::ir::{Function, Type};
+use crate::bindgen::ir::{ArrayLength, Function, Type};
 use crate::bindgen::writer::{ListType, SourceWriter};
 use crate::bindgen::{Config, Language};
 
@@ -56,6 +56,25 @@ impl CDecl {
         cdecl.build_type(t, false);
         cdecl
     }
+
+    fn from_func_arg(t: &Type, array_length: Option<&str>) -> CDecl {
+        let mut cdecl = CDecl::new();
+        let length = match array_length {
+            Some(l) => l,
+            None => return CDecl::from_type(t),
+        };
+        let (ty, is_const) = match t {
+            Type::Ptr { ty, is_const, .. } => (ty, is_const),
+            _ => unreachable!(
+                "Should never have an array length for a non pointer type {:?}",
+                t
+            ),
+        };
+        let ptr_as_array = Type::Array(ty.clone(), ArrayLength::Value(length.to_string()));
+        cdecl.build_type(&ptr_as_array, *is_const);
+        cdecl
+    }
+
     fn from_func(f: &Function, layout_vertical: bool) -> CDecl {
         let mut cdecl = CDecl::new();
         cdecl.build_func(f, layout_vertical);
@@ -66,7 +85,12 @@ impl CDecl {
         let args = f
             .args
             .iter()
-            .map(|arg| (arg.name.clone(), CDecl::from_type(&arg.ty)))
+            .map(|arg| {
+                (
+                    arg.name.clone(),
+                    CDecl::from_func_arg(&arg.ty, arg.array_length.as_deref()),
+                )
+            })
             .collect();
         self.declarators
             .push(CDeclarator::Func(args, layout_vertical));
--- a/src/bindgen/ir/function.rs
+++ b/src/bindgen/ir/function.rs
@@ -2,6 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use std::collections::HashMap;
 use std::io::Write;
 
 use crate::bindgen::cdecl;
@@ -23,6 +24,7 @@ use crate::bindgen::writer::{Source, SourceWriter};
 pub struct FunctionArgument {
     pub name: Option<String>,
     pub ty: Type,
+    pub array_length: Option<String>,
 }
 
 #[derive(Debug, Clone)]
@@ -170,7 +172,11 @@ impl Function {
                 .into_iter()
                 .map(|arg| {
                     let name = arg.name.map(|n| r.apply(&n, IdentifierType::FunctionArg));
-                    FunctionArgument { name, ty: arg.ty }
+                    FunctionArgument {
+                        name,
+                        ty: arg.ty,
+                        array_length: None,
+                    }
                 })
                 .collect()
         }
@@ -181,6 +187,37 @@ impl Function {
             arg.ty.rename_for_config(config, &generic_params);
             if let Some(ref mut name) = arg.name {
                 reserved::escape(name);
+            }
+        }
+
+        // Save the array length of the pointer arguments which need to use
+        // the C-array notation
+        if let Some(tuples) = self.annotations.list("ptrs-as-arrays") {
+            let mut ptrs_as_arrays: HashMap<String, String> = HashMap::new();
+            for str_tuple in tuples {
+                let parts: Vec<&str> = str_tuple[1..str_tuple.len() - 1]
+                    .split(';')
+                    .map(|x| x.trim())
+                    .collect();
+                if parts.len() != 2 {
+                    warn!(
+                        "{:?} does not follow the correct syntax, so the annotation is being ignored",
+                        parts
+                    );
+                    continue;
+                }
+                ptrs_as_arrays.insert(parts[0].to_string(), parts[1].to_string());
+            }
+
+            for arg in &mut self.args {
+                if !matches!(arg.ty, Type::Ptr { .. }) {
+                    continue;
+                }
+                let name = match arg.name {
+                    Some(ref name) => name,
+                    None => continue,
+                };
+                arg.array_length = ptrs_as_arrays.get(name).cloned();
             }
         }
     }
@@ -337,11 +374,16 @@ impl SynFnArgHelpers for syn::FnArg {
                 if let Type::Array(..) = ty {
                     return Err("Array as function arguments are not supported".to_owned());
                 }
-                Ok(Some(FunctionArgument { name, ty }))
+                Ok(Some(FunctionArgument {
+                    name,
+                    ty,
+                    array_length: None,
+                }))
             }
             syn::FnArg::Receiver(ref receiver) => Ok(Some(FunctionArgument {
                 name: Some("self".to_string()),
                 ty: gen_self_type(receiver),
+                array_length: None,
             })),
         }
     }
--- a/tests/rust/ptrs_as_arrays.rs
+++ b/tests/rust/ptrs_as_arrays.rs
@@ -1,0 +1,19 @@
+/// cbindgen:ptrs-as-arrays=[[arg;3]]
+#[no_mangle]
+pub unsafe extern "C" fn ptr_as_array(n: u32, arg: *const u32, v: *const u64) {}
+
+/// cbindgen:ptrs-as-arrays=[[arg;3], [v; 4]]
+#[no_mangle]
+pub unsafe extern "C" fn ptr_as_array1(n: u32, arg: *const u32, v: *mut u64) {}
+
+/// cbindgen:ptrs-as-arrays=[[n;2], [arg; ], [v;], [k; 3]]
+#[no_mangle]
+pub unsafe extern "C" fn ptr_as_array2(n: u32, arg: *mut u32, v: *const u64) {}
+
+/// cbindgen:ptrs-as-arrays=[[a;2;3]]
+#[no_mangle]
+pub unsafe extern "C" fn ptr_as_array_wrong_syntax(arg: *mut u32, v: *const u32, _: *const u32) {}
+
+/// cbindgen:ptrs-as-arrays=[[_;2], [_;3]]
+#[no_mangle]
+pub unsafe extern "C" fn ptr_as_array_unnamed(_: *mut u32, _: *const u32) {}
--- a/docs.md
+++ b/docs.md
@@ -339,6 +339,19 @@ All function attributes are just local overrides for the same options found in t
 * rename-all=RenameRule
 * prefix
 * postfix
+* ptrs-as-arrays=\[[ptr\_name1; array\_length1], [ptr\_name2; array\_length2], ...\] -- represents the pointer arguments of a function as arrays. Below how the mappings are performed:
+
+```
+arg: *const T --> const T arg[array_length]
+arg: *mut T ---> T arg[array_length]
+```
+
+If `array_length` is not specified:
+
+```
+arg: *const T --> const T arg[]
+arg: *mut T --> T arg[]
+```
 
 ## Generating Swift Bindings
 
--- a/tests/expectations/both/ptrs_as_arrays.c
+++ b/tests/expectations/both/ptrs_as_arrays.c
@@ -1,0 +1,14 @@
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+void ptr_as_array(uint32_t n, const uint32_t arg[3], const uint64_t *v);
+
+void ptr_as_array1(uint32_t n, const uint32_t arg[3], uint64_t v[4]);
+
+void ptr_as_array2(uint32_t n, uint32_t arg[], const uint64_t v[]);
+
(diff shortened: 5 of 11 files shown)


src/bindgen/cdecl.rs
```diff
@@ -131,7 +131,14 @@ impl CDecl {
                 self.build_type(t, false);
             }
             Type::Array(ref t, ref constant) => {
-                let len = constant.as_str().to_owned();
+                let mut len = constant.as_str().to_owned();
+                // Checks if a pointer represented as an array is const
+                let is_const = if let Some('c') = len.chars().last() {
```
emilio, line 136: This looks a bit hacky :/

src/bindgen/ir/function.rs
```diff
@@ -167,6 +168,33 @@ impl Function {
         for args in &mut self.args {
             reserved::escape(&mut args.0);
         }
+
+        // Rename pointer arguments which need to use the C-array notation
```
emilio, line 172: It feels bad to change _types_ in a function that is about changing _names_, I wonder if we have precedents for that... If not we should change rename_for_config to something more generic I guess.

Luni-4:
> Thanks for the patch, this looks generally reasonable. The syntax seems very complex for what it really is.

Actually, I've chosen this syntax because it is similar to that of an array. Initially I thought to change bracket squares to parentheses, but that syntax would have been much more similar to tuples. What do you suggest as an alternative syntax instead?

> (I wonder the value of this, more generally, as to the compiler they're all the same. This kind of constraints are better documented in the function declaration IMHO, but...)

Could you elaborate more on this, please?

src/bindgen/cdecl.rs
Luni-4: Yep, it is :)

I didn't want to change a lot of code just to insert the `const` identifier, so since the length of an array is a `String`, I thought to implement this simple trick. The best solution would be adding a new type for `arrays-as-pointer`, but is it worth just for some annotations? It isn't in my opinion. If you have better ideas, I'm happy to listen to them

src/bindgen/ir/function.rs
Luni-4: format_for_config? IMHO, I don't think the name of this function is a problem because, from the `Rust` perspective, we are always dealing with pointers.

lu-zero:
> Thanks for the patch, this looks generally reasonable. The syntax seems very complex for what it really is.
> @lu-zero does this PR solve your use-case? would you use this syntax?

It does :)
 
> (I wonder the value of this, more generally, as to the compiler they're all the same. This kind of constraints are better documented in the function declaration IMHO, but...)

Sadly this fringe feature is apparently liked enough by C developers and clang should be able to use the information to do additional input validation.

emilio:
> Actually, I've chosen this syntax because it is similar to that of an array. Initially I thought to change bracket squares to parentheses, but that syntax would have been much more similar to tuples. What do you suggest as an alternative syntax instead?

Oh to be clear I don't have much better suggestions. It just looks odd that's function-global rather than per-argument, and it's not particularly easy to read I guess, but...

> Could you elaborate more on this, please?

I just meant that what cbindgen does now is correct. This is not a correctness fix, just an stylistic feature.

src/bindgen/cdecl.rs
emilio: I think this is not quite acceptable. There's nothing preventing the user to insert a `c` as an array length, right?

Your patch breaks this, for example:

```rust
pub const c: usize = 4;
pub type C = *const [i16; c];

#[no_mangle]
pub extern "C" fn foo(foo: C) {}
```

Luni-4:
> > Actually, I've chosen this syntax because it is similar to that of an array. Initially I thought to change bracket squares to parentheses, but that syntax would have been much more similar to tuples. What do you suggest as an alternative syntax instead?
> 
> Oh to be clear I don't have much better suggestions. It just looks odd that's function-global rather than per-argument, and it's not particularly easy to read I guess, but...

You're perfectly right, that would be the best solution. As @lu-zero stated in his issue, this was my initial syntax:

```rust
fn fun(arg: *const Type /* cbindgen:as [3] */)
```

But I wasn't able to get comments, it seems they are not parsed, so I had to abandon that approach.

src/bindgen/ir/function.rs
Luni-4: I can:
- rename the function name `rename_for_config` --> `format_for_config`
- create a new function specifically for this annotation feature

Which one do you prefer?

emilio: I think if we do this we should rename it, rather than adding an extra pass just to deal with this. But in hindsight we're already messing with the types in this pass, so maybe don't bother.

emilio:
Do they get parsed if they're doc comments?

emilio:
They do. For something like:

```rust
#[no_mangle]
pub extern "C" fn foo(
    /// FOO BAR
    a: C,
) {}
```

`FOO BAR` is in the "attrs" of the `PatType`, for example.

emilio:
While reviewing this I noticed some pretty bad error messages which I'm fixing in https://github.com/eqrion/cbindgen/pull/550. Also, the last commit should help here if we go for the per-argument approach, as it should make storing annotations in the arguments trivial.

Luni-4:
> While reviewing this I noticed some pretty bad error messages which I'm fixing in #550. Also, the last commit should help here if we go for the per-argument approach, as it should make storing annotations in the arguments trivial.

Thanks a lot! :)

 Now we can add a boolean field to the struct called `is_const_annotation` to check if the pointer is `const` or not. 

About doc comments instead, aren't they out-of-context when used for function arguments?

emilio:
> About doc comments instead, aren't they out-of-context when used for function arguments?

What do you mean with out-of-context here? (Sorry, English is not my first language, so bear with me :P)

Luni-4:
> > About doc comments instead, aren't they out-of-context when used for function arguments?
> 
> What do you mean with out-of-context here? (Sorry, English is not my first language, so bear with me :P)



> > About doc comments instead, aren't they out-of-context when used for function arguments?
> 
> What do you mean with out-of-context here? (Sorry, English is not my first language, so bear with me :P)

It's not my first language either, so sorry if I'm writing down my sentences in a bad way :P

With out-of-context, I meant that doc comments are usually used just above the function signature and not among its arguments. Or can they be used in that case too? 
I know we are dealing with annotations, so we can do whatever we want, but perhaps it would be better not to confuse things so much.

emilio:
As far as I can tell it's fine to use doc comments there, but I agree it's not a very common place to put them. I don't have too strong of an opinion there... @lu-zero?

lu-zero:
I'm not sure how they would mix with `cargo doc`, but beside going further down the dependency stack and figure out how to make syn give us the content of comments, I guess either doc comments or the current syntax seem fine for me.

In the end the user looks at the generated headers.

src/bindgen/ir/function.rs
Luni-4: Perfect, I'm going to leave the name as-is

Luni-4:
@lu-zero @emilio 

I propose to maintain the current syntax for now and change that eventually. What do you think about this idea?

emilio:
That's not great because we'd end up having to maintain two different syntaxes, which is painful for such a tiny feature with no impact on correctness.

Luni-4:
Ok, then I'm just leaving the " change that eventually" part :)

Luni-4:
@emilio 

I tried the following syntax:
- block_comment -> `fn fun(arg: *const Type /* cbindgen:as [3] */)`
- inner_block_doc -> `fn fun(arg: *const Type /*! cbindgen:as [3] */)`
- outer_block_doc -> `fn fun(arg: *const Type /** cbindgen:as [3] */)`

But all of them can't be used because they generate errors in `cargo doc`

lu-zero:
If is possible to land it and have release soon would be great :)

tests/rust/ptrs_as_arrays.rs
```diff
@@ -0,0 +1,11 @@
+/// cbindgen:ptrs-as-arrays=[[arg;3]]
```
lu-zero, line 1: does it work with the normal `//` comment as well?

Luni-4: Unfortunately no, because comments aren't parsed.

lu-zero: I see, I tried it on rav1e and it seems working well for the purpose :)

Luni-4:
@emilio 

Can we merge this one?

src/bindgen/ir/function.rs
```diff
@@ -185,6 +191,33 @@ impl Function {
...
+                let parts: Vec<&str> = str_tuple[1..str_tuple.len() - 1]
+                    .split(';')
+                    .map(|x| x.trim())
+                    .collect();
+                debug_assert!(parts.len() == 2);
+                ptrs_as_arrays.insert(parts[0].to_string(), parts[1].to_string());
+            }
+
+            for args in &mut self.args {
+                if let Some(value) = ptrs_as_arrays.get(args.name.as_ref().unwrap()) {
```
emilio, line 208: `unwrap()` here can crash pretty easily.

src/bindgen/ir/function.rs
```diff
@@ -185,6 +191,33 @@ impl Function {
...
+                    let (ptr_as_array, is_const) = match &args.ty {
```
emilio, line 209: This seems like it may not be the best place to do this conversion... the constness shenanigans look wrong, constness is only about the C declarator and it's needed because of the automatic decay to pointer... But perhaps it's fine.

Still the `is_const_annotation` field should be properly documented. The name reads like the type has some sort of extra `const` annotation somewhere, but this is just about whether the pointer this array represents should be qualified as const.

src/bindgen/ir/function.rs
```diff
@@ -185,6 +191,33 @@ impl Function {
...
+
+        // Rename pointer arguments which need to use the C-array notation
+        if let Some(tuples) = self.annotations.list("ptrs-as-arrays") {
+            let mut ptrs_as_arrays: HashMap<String, String> = HashMap::new();
+            for str_tuple in tuples {
+                let parts: Vec<&str> = str_tuple[1..str_tuple.len() - 1]
+                    .split(';')
+                    .map(|x| x.trim())
+                    .collect();
+                debug_assert!(parts.len() == 2);
```
emilio, line 203: What guarantees that the user actually writes two parts?

This should probably be a proper `assert_eq!` with an error message, or a warning (but then it should be handled gracefully).

emilio (changes requested):
I'm still not a fan of either the syntax or the place where this is implemented, specially since this is just syntax sugar...

But I guess I'm ok living with this if stuff is properly named/documented. And a test for the annotation on a function with unnamed arguments, which right now just crashes.

Luni-4:
> I'm still not a fan of either the syntax or the place where this is implemented, specially since this is just syntax sugar...
> 
> But I guess I'm ok living with this if stuff is properly named/documented. And a test for the annotation on a function with unnamed arguments, which right now just crashes.

Thanks for your review! :)

Sorry, but I can't come up with an alternative syntax or solution for this kind of problem. Pointers-as-arrays notation can be used in C and I think `cbindgen` should at least try to support that, even if I'm aware is not so good as syntax

src/bindgen/ir/function.rs
Luni-4: > This seems like it may not be the best place to do this conversion... the constness shenanigans look wrong, constness is only about the C declarator and it's needed because of the automatic decay to pointer... But perhaps it's fine.

I think it's the best place to do this conversion, otherwise we need to move around the length of each array and that would be confusing in my opinion

> Still the `is_const_annotation` field should be properly documented. The name reads like the type has some sort of extra `const` annotation somewhere, but this is just about whether the pointer this array represents should be qualified as const.

I'm going to change that name, effectively it is not so good as it seems.

Luni-4: Nevermind, perhaps I found another way

Luni-4:
Note: I need to squash the commits of this PR

src/bindgen/cdecl.rs
```diff
@@ -56,6 +56,25 @@ impl CDecl {
         cdecl.build_type(t, false);
         cdecl
     }
+
+    fn from_func_type(t: &Type, length_array: &Option<String>) -> CDecl {
```
emilio, line 60: Let's make this an `Option<&str>` and use as_deref() on the caller.

emilio, line 60: The name of this function makes it seem like `t` is a function type of sort. Should probably be `from_function_arg` or such.

emilio: `length_array` should be named `array_length`, I think. Otherwise it seems like an array of some sort of length value.

src/bindgen/cdecl.rs
```diff
@@ -56,6 +56,25 @@ impl CDecl {
...
+        let mut cdecl = CDecl::new();
+        if length_array.is_some() {
```
emilio, line 62: I think I'd prefer:

```
let length = match array_length {
    Some(l) => l,
    None => return CDecl::from_type(ty),
};

let (ty, is_const) = match t {
    Type::Ptr { ty, is_const, ... } => (ty, is_const),
    _ => unreachable!("Should never have an array length for a non pointer type {:?}", t),
};

let ptr_as_array = Type::Array(ty, ArrayLength::Value(length));
// ...

File: src/bindgen/ir/function.rs [hunk 9]
```diff
@@ -183,6 +189,51 @@ impl Function {
...
+                    .collect();
+                if parts.len() != 2 {
+                    warn!(
+                        "{:?} does not follow the correct syntax, so the annotation is being ignored",
+                        parts
+                    );
+                    continue;
+                }
+                if parts[0] == "_" {
+                    unnamed_ptrs.push(parts[1].to_string());
```
emilio, line 211: It seems weird that the unnamed argument order is positional and gets removed even if the argument is not a pointer... I'd probably just not handle unnamed arguments altogether, just not crashing is fine :)

src/bindgen/ir/function.rs
```diff
@@ -183,6 +189,51 @@ impl Function {
...
+                } else {
+                    ptrs_as_arrays.insert(parts[0].to_string(), parts[1].to_string());
+                }
+            }
+
+            for arg in &mut self.args {
+                let (is_annotation, value) = if arg.name.is_none() && unnamed_ptrs.is_empty() {
```
emilio, line 218: Instead of having a hardcoded boolean that we check below and minting an empty value, let's continue? I think this whole loop should be much simpler, something like:

```
if !matches!(arg.ty,Type::Ptr { .. }) {
    continue;
}

let name = match arg.name {
    Some(ref name) => name,
    None => continue,
};

arg.array_length = ptr_as_arrays.get(name).cloned();
```

src/bindgen/ir/function.rs
Luni-4: How would you treat them? Since they don't have a proper name, I thought it would have been better to use a positional approach. I think the unnamed arguments should be considered, perhaps someone could be interested in this kind of conversion `_: *const T --> const T[array_length]` in a code

Luni-4: I can add a check in order to push only arguments that are pointers

emilio: Well, they can name the arguments in that case, IMO.

Luni-4: I think it could be a constraint for some developers, that's why I have discarded that approach initially, but effectively you're right, loss boilerplate for us :smile:

lu-zero: People that care about describing the intent of an argument would probably also name it, so I do not see it as something somebody would miss.

Luni-4:
There is an error, let me fix it :)

Luni-4:
@emilio 

Ok, now it's ready!

emilio (approved):
Ok, let's go with this, thanks for bearing with me :)

Luni-4:
Thanks to you for helping me in making this PR much clear, very appreciated! :)

Luni-4:
@emilio 

Can we bump a new version, please?

emilio:
Yes, I'll try to do a release this week, please nag if that doesn't happen :)

Luni-4:
@emilio 

Nagging you gently :)

Issue #548, opened 2020-07-22
Support for `!` return type


mexus:
Currently, when run against a function that never returns, like `fn abort() -> ! { .. }`, cbindgen skips it with an error message like `... Unsupported type: Never(TypeNever { bang_token: Bang })`.

Could we possibly get support for it? For C and C++ prior to C++11, I guess all major compilers support the `noreturn` attribute as `__attribute__ ((noreturn))`, while a `[[ noreturn ]]` attribute has been added to C++11.

mexus:
Resolved with #549

PR #549, opened 2020-07-22
Support "never" return type

mexus:
This is my attempt to add support for "never" return type in function declarations.

Solves #548

### DIFF
--- a/src/bindgen/ir/function.rs
+++ b/src/bindgen/ir/function.rs
@@ -33,6 +33,7 @@ pub struct Function {
     pub cfg: Option<Cfg>,
     pub annotations: AnnotationSet,
     pub documentation: Documentation,
+    pub never_return: bool,
 }
 
 impl Function {
@@ -46,10 +47,16 @@ impl Function {
     ) -> Result<Function, String> {
         let mut args = sig.inputs.iter().try_skip_map(|x| x.as_ident_and_type())?;
 
+        let mut never_return = false;
         let mut ret = match sig.output {
             syn::ReturnType::Default => Type::Primitive(PrimitiveType::Void),
             syn::ReturnType::Type(_, ref ty) => {
-                Type::load(ty)?.unwrap_or_else(|| Type::Primitive(PrimitiveType::Void))
+                if let syn::Type::Never(_) = ty.as_ref() {
+                    never_return = true;
+                    Type::Primitive(PrimitiveType::Void)
+                } else {
+                    Type::load(ty)?.unwrap_or_else(|| Type::Primitive(PrimitiveType::Void))
+                }
             }
         };
 
@@ -69,6 +76,7 @@ impl Function {
             cfg: Cfg::append(mod_cfg, Cfg::load(attrs)),
             annotations: AnnotationSet::load(attrs)?,
             documentation: Documentation::load(attrs),
+            never_return,
         })
     }
 
@@ -205,6 +213,12 @@ impl Source for Function {
                 write!(out, " {}({})", swift_name_macro, func.swift_name());
             }
 
+            if func.never_return {
+                if let Some(ref no_return_attr) = config.function.no_return {
+                    out.write_fmt(format_args!(" {}", no_return_attr));
+                }
+            }
+
             out.write(";");
 
             condition.write_after(config, out);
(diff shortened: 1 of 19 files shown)


src/bindgen/bindings.rs
```diff
@@ -215,6 +215,27 @@ impl Bindings {
 
         self.open_namespaces(&mut out);
 
+        let noreturn_required = self.functions.iter().any(|x| x.never_return);
```
emilio, line 218: It'd be good to not introduce a walk for every function here.

src/bindgen/ir/function.rs
```diff
@@ -46,10 +47,16 @@ impl Function {
     ) -> Result<Function, String> {
         let mut args = sig.inputs.iter().try_skip_map(|x| x.as_ident_and_type())?;
 
+        let mut never_return = false;
         let mut ret = match sig.output {
             syn::ReturnType::Default => Type::Primitive(PrimitiveType::Void),
             syn::ReturnType::Type(_, ref ty) => {
-                Type::load(ty)?.unwrap_or_else(|| Type::Primitive(PrimitiveType::Void))
+                if let syn::Type::Never(_) = ty as &syn::Type {
```
emilio, line 54: Do you really need the `as &syn::Type`?

src/bindgen/ir/function.rs
```diff
@@ -246,6 +258,10 @@ impl Source for Function {
                 write!(out, " {}({})", swift_name_macro, func.swift_name());
             }
 
+            if func.never_return {
+                out.write(" CBINDGEN_NO_RETURN_ATTR ");
```
emilio, line 262: here instead of hardcoding this we'd do something like:

```rust
if func.never_return {
    if let Some(ref no_return_attr) = config.function.no_return {
        out.write(" ");
        out.write(no_return_attr);
        out.write(" ");
    }
}
```

src/bindgen/bindings.rs
```diff
@@ -215,6 +215,27 @@ impl Bindings {
...
+        if noreturn_required {
```
emilio, line 219: I think the way we've gone about similar C / C++ annotations is a bit different. Instead of hardcoding the macro here, we just add a configuration option so that the user of the library can use whatever macro they want / need. See for example how `must_use` works.

I think by default converting `Never` to `Void` and _then_ adding the attribute if specified in the configuration would be better.

It also avoids the walk over every function, and prevents hardcoding compiler macros here. Otherwise for users that compile with MSVC you'd need to use `__declspec(noreturn)`, etc.

Wdyt? Would such a thing work for you?

mexus:
Thanks for such a fast review :) All the points you've mentioned are absolutely clear; hope I'll be able to make the amends within a day or two.

src/bindgen/ir/function.rs
mexus: I guess the best thing I can do is to replace `ty as &syn::Type` with `ty.as_ref()`; will do that

mexus:
I've marked all the "conversations" as resolved, but please have a look at the updated commit :)

src/bindgen/ir/function.rs
```diff
@@ -205,6 +213,10 @@ impl Source for Function {
                 write!(out, " {}({})", swift_name_macro, func.swift_name());
             }
 
+            if let Some(no_return_attr) = &config.function.no_return {
+                out.write_fmt(format_args!(" {} ", no_return_attr));
```
emilio, line 217: This needs to somehow be conditional on `never_return`, right? Please add a non-never-returning function to the new test.

src/bindgen/ir/function.rs
```diff
@@ -246,6 +258,10 @@ impl Source for Function {
...
```
emilio, line 262: Same here.

Also, for consistency, use `if let Some(ref ..) = config... {` rather than `if let Some(..) = &config... {`?

emilio: (Also do we really need the trailing space?)

mexus:
Hope this time I've managed not to forget anything :smile:

mexus:
Thanks a lot @emilio !!

PR #556, opened 2020-07-30
Allow 1<<31 as a bitflags value

### LINKED ISSUE #555: Running cbindgen on a u32 bitflag with a 1<<31 entry produces C++ code that doesn't compile
Assuming you have a setup where `cargo +nightly test` passes everything, apply this patch:

```
diff --git a/tests/rust/bitflags.rs b/tests/rust/bitflags.rs
index 6c3fe4e..017104a 100644
--- a/tests/rust/bitflags.rs
+++ b/tests/rust/bitflags.rs
@@ -13,10 +13,18 @@ bitflags! {
         const START = 1 << 1;
         /// 'end'
         const END = 1 << 2;
         /// 'flex-start'
         const FLEX_START = 1 << 3;
     }
 }
 
+bitflags! {
+    #[repr(C)]
+    pub struct DebugFlags: u32 {
+        /// Flag with the topmost bit set of the u32
+        const BIGGEST_ALLOWED = 1 << 31;
+    }
+}
+
 #[no_mangle]
-pub extern "C" fn root(flags: AlignFlags) {}
+pub extern "C" fn root(flags: AlignFlags, bigger_flags: DebugFlags) {}
```

and run the tests. For me there's this failure:

```
Running: "/home/kats/.mozbuild/clang/bin/clang++" "-D" "DEFINED" "-c" "/home/kats/zspace/cbindgen/tests/expectations/bitflags.cpp" "-o" "/home/kats/tmp/cbindgen-test-outputWsIxRN/bitflags.o" "-I" "/home/kats/zspace/cbindgen/tests" "-Wall" "-Werror" "-Wno-attributes" "-std=c++17" "-Wno-deprecated" "-Wno-unused-const-variable" "-Wno-return-type-c-linkage"
thread 'test_bitflags' panicked at 'Output failed to compile: Output { status: ExitStatus(ExitStatus(256)), stdout: "", stderr: "/home/kats/zspace/cbindgen/tests/expectations/bitflags.cpp:83:80: error: constant expression evaluates to -2147483648 which cannot be narrowed to type \'uint32_t\' (aka \'unsigned int\') [-Wc++11-narrowing]\nstatic const DebugFlags DebugFlags_BIGGEST_ALLOWED = DebugFlags{ /* .bits = */ (1 << 31) };\n                                                                               ^~~~~~~~~\n/home/kats/zspace/cbindgen/tests/expectations/bitflags.cpp:83:80: note: insert an explicit cast to silence this issue\nstatic const DebugFlags DebugFlags_BIGGEST_ALLOWED = DebugFlags{ /* .bits = */ (1 << 31) };\n                                                                               ^~~~~~~~~\n                                                                               static_cast<uint32_t>( )\n1 error generated.\n" }', tests/tests.rs:118:5
```

staktrace:
Fixes #555

### DIFF
--- a/src/bindgen/bitflags.rs
+++ b/src/bindgen/bitflags.rs
@@ -43,7 +43,7 @@ impl Bitflags {
             }
         };
 
-        let consts = flags.expand(name);
+        let consts = flags.expand(name, repr);
         let impl_ = parse_quote! {
             impl #name {
                 #consts
@@ -81,7 +81,7 @@ struct Flag {
 }
 
 impl Flag {
-    fn expand(&self, struct_name: &syn::Ident) -> TokenStream {
+    fn expand(&self, struct_name: &syn::Ident, repr: &syn::Type) -> TokenStream {
         let Flag {
             ref attrs,
             ref name,
@@ -90,7 +90,7 @@ impl Flag {
         } = *self;
         quote! {
             #(#attrs)*
-            pub const #name : #struct_name = #struct_name { bits: #value };
+            pub const #name : #struct_name = #struct_name { bits: (#value) as #repr };
         }
     }
 }
@@ -124,10 +124,10 @@ impl Parse for Flags {
 }
 
 impl Flags {
-    fn expand(&self, struct_name: &syn::Ident) -> TokenStream {
+    fn expand(&self, struct_name: &syn::Ident, repr: &syn::Type) -> TokenStream {
         let mut ts = quote! {};
         for flag in &self.0 {
-            ts.extend(flag.expand(struct_name));
+            ts.extend(flag.expand(struct_name, repr));
         }
         ts
     }
--- a/tests/expectations/associated_in_body.c
+++ b/tests/expectations/associated_in_body.c
@@ -14,22 +14,22 @@ typedef struct {
 /**
  * 'auto'
  */
-#define StyleAlignFlags_AUTO (StyleAlignFlags){ .bits = 0 }
+#define StyleAlignFlags_AUTO (StyleAlignFlags){ .bits = (uint8_t)0 }
 /**
  * 'normal'
  */
-#define StyleAlignFlags_NORMAL (StyleAlignFlags){ .bits = 1 }
+#define StyleAlignFlags_NORMAL (StyleAlignFlags){ .bits = (uint8_t)1 }
 /**
  * 'start'
  */
-#define StyleAlignFlags_START (StyleAlignFlags){ .bits = (1 << 1) }
+#define StyleAlignFlags_START (StyleAlignFlags){ .bits = (uint8_t)(1 << 1) }
 /**
  * 'end'
  */
-#define StyleAlignFlags_END (StyleAlignFlags){ .bits = (1 << 2) }
+#define StyleAlignFlags_END (StyleAlignFlags){ .bits = (uint8_t)(1 << 2) }
 /**
  * 'flex-start'
  */
-#define StyleAlignFlags_FLEX_START (StyleAlignFlags){ .bits = (1 << 3) }
+#define StyleAlignFlags_FLEX_START (StyleAlignFlags){ .bits = (uint8_t)(1 << 3) }
 
 void root(StyleAlignFlags flags);
--- a/tests/expectations/associated_in_body.compat.c
+++ b/tests/expectations/associated_in_body.compat.c
@@ -14,23 +14,23 @@ typedef struct {
 /**
  * 'auto'
  */
-#define StyleAlignFlags_AUTO (StyleAlignFlags){ .bits = 0 }
+#define StyleAlignFlags_AUTO (StyleAlignFlags){ .bits = (uint8_t)0 }
 /**
  * 'normal'
  */
-#define StyleAlignFlags_NORMAL (StyleAlignFlags){ .bits = 1 }
+#define StyleAlignFlags_NORMAL (StyleAlignFlags){ .bits = (uint8_t)1 }
 /**
  * 'start'
  */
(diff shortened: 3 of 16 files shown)


froydnj:
Don't you have to cast the value being shifted, not the result of the shift?  `1 << 31` is undefined behavior because it's shifting different values through the sign bit; `uint32_t(1) << 31` is not.

staktrace:
Hm, good point. I'm not sure how easy it is to plumb that into cbindgen. It would require breaking apart the binary operation and "knowing" that "1" is going to be the same type as the field in the struct, or something along those lines.

emilio:
I need to take a closer look, that being said, it baffles me a bit that `static const uint32_t CONST = (1 << 31);` doesn't warn, but `static const Foo FOO = Foo { (1 << 31) };` does.

emilio:
Hmm, so... Do we need this cast in C mode?

If so, we may want something like this instead (it also generates sligthly nicer code IMO, wdyt @staktrace?):

```diff
diff --git a/src/bindgen/bitflags.rs b/src/bindgen/bitflags.rs
index 6c89622..cc1a4cb 100644
--- a/src/bindgen/bitflags.rs
+++ b/src/bindgen/bitflags.rs
@@ -43,7 +43,7 @@ impl Bitflags {
             }
         };
 
-        let consts = flags.expand(name);
+        let consts = flags.expand(name, repr);
         let impl_ = parse_quote! {
             impl #name {
                 #consts
@@ -81,7 +81,7 @@ struct Flag {
 }
 
 impl Flag {
-    fn expand(&self, struct_name: &syn::Ident) -> TokenStream {
+    fn expand(&self, struct_name: &syn::Ident, repr: &syn::Type) -> TokenStream {
         let Flag {
             ref attrs,
             ref name,
@@ -90,7 +90,7 @@ impl Flag {
         } = *self;
         quote! {
             #(#attrs)*
-            pub const #name : #struct_name = #struct_name { bits: #value };
+            pub const #name : #struct_name = #struct_name { bits: (#value) as #repr };
         }
     }
 }
@@ -124,10 +124,10 @@ impl Parse for Flags {
 }
 
 impl Flags {
-    fn expand(&self, struct_name: &syn::Ident) -> TokenStream {
+    fn expand(&self, struct_name: &syn::Ident, repr: &syn::Type) -> TokenStream {
         let mut ts = quote! {};
         for flag in &self.0 {
-            ts.extend(flag.expand(struct_name));
+            ts.extend(flag.expand(struct_name, repr));
         }
         ts
     }
```

Also, question for @froydnj: Is it really UB? Or only "implementation defined" (which is well defined if you assume two's complement?). I'm surprised something so simple wouldn't warn with Wpedantic if it was UB, but...

A funny consequence of this patch is that now nested struct expressions look really funny:

```rust
#[repr(C)]
pub struct Bar {
    baz: u32,
}

#[repr(C)]
pub struct Foo {
    bar: Bar,
}

pub const FOO: Foo = Foo { bar: Bar { baz: 3 } };
```

generates:

```c++
static const Foo FOO = Foo{ /* .bar = */ static_cast<decltype(Foo::bar)>(Bar{ /* .baz = */ static_cast<decltype(Bar::baz)>(3) }) };
```

Which is ugly but compiles, turns out (wasn't sure whether you could static-cast structs). Maybe the bitflags-scoped fix is slightly less invasive.

emilio:
Something that _is_ UB would be:

```c++
uint64_t FOO = (1 << 32);
```

Which I guess is also an issue for rust bitflags, as the equivalent rust code does compile just fine. You need to insert a rust cast for this to behave correctly which is unfortunate: `pub const FOO: u64 = (1 as u64) << 33;`

But well that does warn loudly, so there's that...

staktrace:
> Hmm, so... Do we need this cast in C mode?

It doesn't hurt to have it in C mode too, I guess. Your patch is definitely cleaner than mine and generates better code, I agree. I'll replace mine with yours (and retain your authorship).

I didn't realize the "x as y" construct got converted to C/C++ properly, so that definitely makes things easier for me.

froydnj:
> Also, question for @froydnj: Is it really UB? Or only "implementation defined" (which is well defined if you assume two's complement?). I'm surprised something so simple wouldn't warn with Wpedantic if it was UB, but...

It is really UB in C++14 and below.  I think C++17 and above have different language around left shifts that make it defined for the case @staktrace ran into.

C++14 or so draft: "Otherwise, if E1 has a signed type and non-negative value, and E1 × 2^E2 is representable
in the result type, then that is the resulting value; otherwise, the behavior is undefined." (the "result type" here is the type of the left operand, i.e. `int` unless otherwise specified)

C++17 or so draft: "Otherwise, if E1 has a signed type and non-negative value, and E1 × 2^E2 is representable
in the corresponding unsigned type of the result type, then that value, converted to the result type, is the
resulting value; otherwise, the behavior is undefined."  (I guess technically, the conversion from unsigned to signed is implementation-defined here.)

[Latest draft](http://eel.is/c++draft/expr.shift): "The value of E1 << E2 is the unique value congruent to E1×2^E2 modulo 2N, where N is the width of the type of the result." (Note how the implementation-defined conversion behavior has been removed.)

I don't know what the C standard says, but I'd imagine C11 and below follow C++14, and possibly even C2a or whatever.

emilio:
Ugh, that really kinda sucks. It sucks more that there's literally no warning for this in the simple case... But well this is a bit better than what we're doing now, and it fixes @staktrace's case since we compile with C++17 in mozilla-central.

And users of `cbindgen` can also add a cast where appropriate to cause the behavior to be properly defined. Thanks for tweaking the patch @staktrace! This looks ok to me.

PR #558, opened 2020-08-04
Add support for an optional nonnull attribute

nickwilcox:
This allows clang and swiftc users to use `_Nonnull`

Tested with `CC=clang CXX=clang++ cargo test` as well as `cargo test`

### DIFF
--- a/src/bindgen/cdecl.rs
+++ b/src/bindgen/cdecl.rs
@@ -14,7 +14,7 @@ use crate::bindgen::{Config, Language};
 // http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf
 
 enum CDeclarator {
-    Ptr(bool, bool),
+    Ptr { is_const: bool, is_nullable: bool },
     Ref,
     Array(String),
     Func(Vec<(Option<String>, CDecl)>, bool),
@@ -23,7 +23,7 @@ enum CDeclarator {
 impl CDeclarator {
     fn is_ptr(&self) -> bool {
         match self {
-            CDeclarator::Ptr(..) | CDeclarator::Ref | CDeclarator::Func(..) => true,
+            CDeclarator::Ptr { .. } | CDeclarator::Ref | CDeclarator::Func(..) => true,
             _ => false,
         }
     }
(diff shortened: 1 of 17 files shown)


src/bindgen/cdecl.rs
```diff
@@ -14,7 +14,7 @@ use crate::bindgen::{Config, Language};
 // http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf
 
 enum CDeclarator {
-    Ptr(bool),
```
emilio, line 17: Can we name these? One unnamed bool is bad enough, but two is super easy to mess up, so:

```
Ptr { is_const: bool, non_null: bool }
```

Though it seems like it may be better to keep `nullable: bool` rather than `non_null`.

src/bindgen/cdecl.rs
```diff
@@ -178,11 +186,17 @@ impl CDecl {
             let next_is_pointer = iter_rev.peek().map_or(false, |x| x.is_ptr());
 
             match *declarator {
-                CDeclarator::Ptr(ref is_const) => {
-                    if *is_const {
-                        out.write("*const ");
+                CDeclarator::Ptr(ref is_const, ref is_nonnull) => {
```
emilio, line 189: Just keep the code like it was, then do:

```
if is_nonnull {
    if let Some(ref attr) = config.non_null_attribute {
        write!(out, "{} ", attr);
    }
}
```

? I don't see why having the match on both is useful.

src/bindgen/config.rs
```diff
@@ -765,6 +765,8 @@ pub struct Config {
     pub documentation: bool,
     /// How documentation comments should be styled.
     pub documentation_style: DocumentationStyle,
+    /// Optional attribute to apply to pointers that are required to not be null
+    pub non_null_attribute: Option<String>,
```
emilio, line 769: I think rather than it being in the top level section it may be worth introducing a `[ptr]` section of sorts like the `ConstantConfig` and such. It just feels a little out of place... wdyt?

src/bindgen/ir/ty.rs
```diff
@@ -211,6 +211,8 @@ impl ArrayLength {
 pub enum Type {
     ConstPtr(Box<Type>),
     Ptr(Box<Type>),
+    ConstNonNullPtr(Box<Type>),
```
emilio, line 214: Rather than this, can we have the nullability inside `ConstPtr` and `Ptr`, just like in `CDeclarator`? Seems like it'd be less duplicated code over all. Though that being said, isn't a non-nullable pointer basically just a `Ref` / `MutRef`? If so, can we use that instead?

src/bindgen/ir/ty.rs
```diff
@@ -369,12 +371,25 @@ impl Type {
...
             Type::Ptr(..) => true,
+            Type::NonNullPtr(..) => true,
             Type::ConstPtr(..) => true,
+            Type::ConstNonNullPtr(..) => true,
             Type::FuncPtr(..) => true,
             _ => false,
         }
     }
 
+    pub fn make_nullable(&self) -> Option<Self> {
```
emilio, line 382: Ref should be turned into ptr here.

src/bindgen/cdecl.rs
nickwilcox: Done

src/bindgen/cdecl.rs
nickwilcox: Done.

src/bindgen/config.rs
nickwilcox: done

src/bindgen/ir/ty.rs
nickwilcox: I've moved the nullability into `ConstPtr` and `Ptr` and gotten rid of my new variants. I didn't use `Ref`/`MutRef` because they aren't created when this code parses the syntax. I thought they might be legacy - they're only created in one place inside enumeration.rs.

emilio: There's no good reason they can't be created during parsing.

src/bindgen/config.rs
```diff
@@ -692,6 +692,24 @@ impl ParseConfig {
...
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "snake_case")]
+#[serde(deny_unknown_fields)]
+#[serde(default)]
+pub struct PtrConfig {
+    /// Optional attribute to apply to pointers that are required to not be null
+    pub non_null_attribute: Option<String>,
+}
+
+impl Default for PtrConfig {
```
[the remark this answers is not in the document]
emilio: I'll do in a follow-up.

PR #565, opened 2020-08-19
Pascal case primitives

adamski:
This properly addresses our use case where we want to the default of e.g. `MyType______c_char` to be `MyTypeCChar`. The previous PR did not take the `c_char` type or similar into account, only capitalising the first letter for e.g. `MyTypeU32`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ toml = "0.5"
 proc-macro2 = "1"
 quote = "1"
 heck = "0.3"
+convert_case = "0.4.0"
 
 [dependencies.syn]
 version = "1.0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,6 @@ toml = "0.5"
 proc-macro2 = "1"
 quote = "1"
 heck = "0.3"
-convert_case = "0.4.0"
 
 [dependencies.syn]
 version = "1.0.3"
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -299,8 +299,10 @@ pub struct ExportConfig {
     pub item_types: Vec<ItemType>,
     /// Whether renaming overrides or extends prefixing.
     pub renaming_overrides_prefixing: bool,
-    /// Name mangling character. Defaults to `_`
-    pub mangle_separator: Option<String>,
+    /// Remove the underscores used for name mangling
+    pub remove_underscores: bool,
+    /// Whether to capitalise primitives, e.g. if mangle_separator is empty
+    pub pascal_case_primitives: bool,
 }
 
 impl ExportConfig {
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -287,6 +287,8 @@ pub struct ExportConfig {
     pub include: Vec<String>,
     /// A list of items to not include in the generated bindings
     pub exclude: Vec<String>,
+    /// The rename rule to apply to the type names
+    pub rename_types: Option<RenameRule>,
     /// Table of name conversions to apply to item names
     pub rename: HashMap<String, String>,
     /// Table of raw strings to prepend to the body of items.
@@ -301,8 +303,6 @@ pub struct ExportConfig {
     pub renaming_overrides_prefixing: bool,
     /// Remove the underscores used for name mangling
     pub remove_underscores: bool,
-    /// Whether to capitalise primitives, e.g. if mangle_separator is empty
-    pub pascal_case_primitives: bool,
 }
 
 impl ExportConfig {
--- a/src/bindgen/ir/opaque.rs
+++ b/src/bindgen/ir/opaque.rs
@@ -122,7 +122,8 @@ impl Item for OpaqueItem {
         let mangled_path = mangle::mangle_path(
             &self.path,
             generic_values,
-            library.get_config().export.mangle_separator.as_deref(),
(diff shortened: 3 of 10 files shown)


Cargo.toml
```diff
@@ -24,6 +24,7 @@ toml = "0.5"
 proc-macro2 = "1"
 quote = "1"
 heck = "0.3"
+convert_case = "0.4.0"
```
emilio, line 27: We have `heck` already to do case conversions, let's not introduce another dependency just because.

src/bindgen/config.rs
```diff
@@ -248,6 +248,8 @@ pub struct ExportConfig {
     pub renaming_overrides_prefixing: bool,
     /// Name mangling character. Defaults to `_`
     pub mangle_separator: Option<String>,
+    /// Whether to capitalise primitives, e.g. if mangle_separator is empty
+    pub pascal_case_primitives: bool,
```
emilio, line 252: We have another rename rules (see `RenameRule`) that you can use, it seems using that would be better for consistency, rather than special-casing pascal-case here.

src/bindgen/ir/opaque.rs
```diff
@@ -125,6 +125,7 @@ impl Item for OpaqueItem {
             &self.path,
             generic_values,
             library.get_config().export.mangle_separator.as_deref(),
+            library.get_config().export.pascal_case_primitives,
```
emilio, line 128: Maybe pass `config` directly?

emilio (changes requested):

adamski:
So I was thinking about this, and my feeling is that the previous (now merged) PR that allows for different separators, is possibly not that useful (at least I can't think of a use case where someone would want a different character to `_`).

What we ultimately want is to be able to PascalCase generated types/typedefs. So my proposal is to merge the two (remove underscores and PascalCase primitives), to something like `rename_types = "PascalCase"`. Perhaps a `remove_underscores` bool in addition to a `rename_types` rule.

src/bindgen/config.rs
adamski: OK thanks, I was not aware of that, see my comment below.

Cargo.toml
adamski: Got it. Agreed.

emilio:
`remove_underscores` + `rename_types` would be the same as `mangling_separator = ""` and `rename_types`, right? If so, I think there's no need to change the existing option unless you feel really strongly about it.

adamski:
> `remove_underscores` + `rename_types` would be the same as `mangling_separator = ""` and `rename_types`, right?

Yes, although I'm also suggesting they could be merged into one option.

> If so, I think there's no need to change the existing option unless you feel really strongly about it.

OK, I'm just struggling to think of a use case where someone would want a different `mangle_seperator` to `_`...

adamski:
I haven't passed down `config` purely because it complicates the tests, which indicates to me that perhaps it's clearer not to pass the config down... If you prefer to pass `config` down, the tests will need to create an `ExportConfig` instance to pass into the mangler.

src/bindgen/ir/opaque.rs
adamski: See may comment below

adamski:
@emilio please review the updated changes when you have time. Thank you.

src/bindgen/rename.rs
```diff
@@ -21,6 +22,7 @@ impl<'a> IdentifierType<'a> {
             IdentifierType::StructMember => "m",
             IdentifierType::EnumVariant(..) => "",
             IdentifierType::FunctionArg => "a",
+            IdentifierType::Type => "t",
```
emilio, line 25: This should be an empty string.

src/bindgen/mangle.rs
```diff
@@ -3,17 +3,37 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 use crate::bindgen::ir::{Path, Type};
+use crate::bindgen::rename::IdentifierType;
+use crate::bindgen::rename::RenameRule;
+#[cfg(test)]
+use crate::bindgen::rename::RenameRule::PascalCase;
```
emilio, line 9: Maybe move this to the test functions, or just use `RenameRule::PascalCase`?

src/bindgen/mangle.rs
```diff
@@ -143,7 +180,7 @@ fn generics() {
 
     // Foo<f32> => Foo_f32
     assert_eq!(
-        mangle_path(&Path::new("Foo"), &vec![float()], None),
+        mangle_path(&Path::new("Foo"), &vec![float()], false, None::<RenameRule>),
```
emilio, line 183: I don't think you need the turbofish here.

emilio (commented):
This looks generally good, but I have some nits and one error in to_str. I'm still not a fan of removing an existing option for `mangle_separator` somewhat arbitrarily, but I guess it's relatively new so it can be fine.

Can you squash the commits too?

emilio:
Thanks! This looks fine. I have a few nits but I addressed them in a follow-up commit in https://github.com/eqrion/cbindgen/pull/575.

In particular, the rename_types should probably be on its own subsection of the configuration, otherwise being under `[export]` seems like you'd rename all the types in here which is not what you're doing. I also added some docs and some integration tests.

emilio:
Closing because I merged this in #575

PR #568, opened 2020-08-26
Add support for specifying line ending style

tgolsson:
Hello!

This is a PR to allow specification of the desired line ending style in the generated file. The background for this is working in a git repository shared between Windows and Linux users. The windows users have their git `core.autocrlf` flag setup to check out CRLF, submit LF, and Linux users only work with LF. Both the input and output files are tracked in source control to highlight inadvertent changes in the generated API. 

Since cbindgen generates with only '\n' (based on `write!`) their checked-out CRLF header file becomes LF, and they both get a meaningless whitespace-only diff as-well-as a warning from git that they have a LF file in their CRLF repository.

This changeset introduces a new configuration option that gives four options for line endings: LF, CR, CRLF, or Native. The last option will try to follow the configuration on the target_os, which in practice means CRLF on Windows and LF everywhere else. The default value will remain LF, making this opt-in and fully backwards compatible.

### DIFF
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -704,6 +704,51 @@ pub struct PtrConfig {
     pub non_null_attribute: Option<String>,
 }
 
+/// Controls what type of line endings are used in the generated code.
+#[derive(Debug, Clone, Copy, Deserialize)]
+pub enum LineEndingStyle {
+    /// Use Unix-style linefeed characters
+    LF,
+    /// Use Mac-style carrie-return characters
+    CR,
+
+    /// Use Windows-style carriage-return and linefeed characters
+    CRLF,
+
+    /// Use the native mode for the platform
+    Native,
+}
+
+impl Default for LineEndingStyle {
+    fn default() -> Self {
+        LineEndingStyle::LF
+    }
+}
+
+impl LineEndingStyle {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::LF => &"\n",
+            Self::CR => &"\r",
+            Self::CRLF => &"\r\n",
+            Self::Native => {
+                #[cfg(target_os = "windows")]
+                {
+                    Self::CRLF.as_str()
+                }
+                #[cfg(target_os = "macos")]
+                {
+                    Self::CR.as_str()
+                }
+                #[cfg(not(any(target_os = "windows", target_os = "macos")))]
+                {
+                    Self::LF.as_str()
+                }
+            }
+        }
+    }
+}
+
 /// A collection of settings to customize the generated bindings.
 #[derive(Debug, Clone, Deserialize)]
 #[serde(rename_all = "snake_case")]
@@ -745,6 +790,8 @@ pub struct Config {
     pub line_length: usize,
     /// The amount of spaces in a tab
     pub tab_width: usize,
+    /// The type of line endings to generate
+    pub line_endings: Option<LineEndingStyle>,
     /// The language to output bindings for
     pub language: Language,
     /// Include preprocessor defines in C bindings to ensure C++ compatibility
@@ -801,6 +848,7 @@ impl Default for Config {
             braces: Braces::SameLine,
             line_length: 100,
             tab_width: 2,
+            line_endings: None,
             language: Language::Cxx,
             cpp_compat: false,
             style: Style::Type,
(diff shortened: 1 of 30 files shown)


src/bindgen/config.rs
```diff
@@ -48,6 +48,61 @@ impl FromStr for Language {
...
+        LineEndingStyle::LF
+    }
+}
+
+impl LineEndingStyle {
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::LF => &"\n",
+            Self::CR => &"\r",
+            Self::CRLF => &"\r\n",
```
emilio, line 75: no need for the ampersands here.

src/bindgen/config.rs
```diff
@@ -745,6 +800,8 @@ pub struct Config {
     pub line_length: usize,
     /// The amount of spaces in a tab
     pub tab_width: usize,
+    /// The type of line endings to generate
+    pub line_endings: Option<LineEndingStyle>,
```
emilio, line 804: We could probably just make this `LineEndingStyle`, and initialize it with `LineEndingStyle::default()` below, instead of branching every line-break :)

src/bindgen/writer.rs
```diff
@@ -137,7 +138,7 @@ impl<'a, F: Write> SourceWriter<'a, F> {
     }
 
     pub fn new_line(&mut self) {
-        writeln!(self.out).unwrap();
```
emilio, line 140: `self.out.write_all(self.bindings.config.line_endings.as_str().as_bytes())` would be slightly more performant I think.

emilio (commented):
This looks generally sensible. Can we have some tests for this to avoid regressing it? Probably we can't test the `Native` option easily, but others should be pretty easy.

tgolsson:
@emilio Thanks for the feedback, will address your points in a new commit. I added some tests using the test machinery but I'm not sure how it's supposed to work - are there any more steps than adding the input and outputs to tests/rust and tests/expected? It seemed to just clobber and replace my files.

src/bindgen/config.rs
tgolsson: I figured "Option" here implied optional in the toml? Is the toml always overlaid on top of the default values and all keys are optional?

src/bindgen/writer.rs
tgolsson: Good point, yes. I was so annoyed by it not letting me pass the static str reference as a literal in `write!` that I didn't consider dropping down to a straight call on the file.

src/bindgen/config.rs
emilio: Yes, all keys are optional, otherwise your patch wouldn't pass CI :)

emilio:
You're supposed to add tests to `tests/rust`. If you need to change the configuration as in this case you need to add both a rust file and a toml file with the same name. `tests/expected` is generated from `test/rust`. CI does check that running tests doesn't change the output (via the `CBINDGEN_TEST_VERIFY=1` environment variable).

Additionally, you may need `cargo +nightly fmt`, see https://travis-ci.org/github/eqrion/cbindgen/builds/721245490 (sorry, I don't know at which point did the travis status stop showing up in GitHub :/).

tgolsson:
Thanks! So running the tests and verifying that the generated files are correct; I then commit that and that locks them in as good? Cool. For the native; yes, hard to test. We can verify that specifying Native == LF? I assume your CI environment runs !Windows. OTOH; that'll dirty the file for Windows developers.

emilio:
Yeah, right, that'd be unfortunate for contributors that use Windows, so let's not test that :)

tests/rust/linestyle_crlf.toml
```diff
@@ -0,0 +1,2 @@
+language = "c"
```
emilio, line 1: Any reason to force `language = "c"`? (that forces even the cpp to be c)

tgolsson: Thanks; no! Will fix. The language here doesn't even matter; I was just trying to cut down on the amount of generated content. (Since all we care about is the line ending.)

tgolsson:
@emilio CI seemed to pass here - https://travis-ci.org/github/eqrion/cbindgen/pull_requests. Any more comments or good to go?

emilio (approved):
Nope, looks good, thanks!

PR #571, opened 2020-09-06
Fix clippy warnings

Luni-4:
Fix clippy warnings

### DIFF
--- a/src/bindgen/builder.rs
+++ b/src/bindgen/builder.rs
@@ -23,6 +23,7 @@ pub struct Builder {
 }
 
 impl Builder {
+    #[allow(clippy::new_without_default)]
     pub fn new() -> Builder {
         Builder {
             config: Config::default(),
--- a/src/bindgen/declarationtyperesolver.rs
+++ b/src/bindgen/declarationtyperesolver.rs
@@ -6,6 +6,7 @@ use std::collections::HashSet;
 
 use crate::bindgen::ir::Path;
 
+#[derive(Default)]
 pub struct DeclarationTypeResolver {
     structs: HashSet<Path>,
     enums: HashSet<Path>,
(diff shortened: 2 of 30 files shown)


src/bindgen/declarationtyperesolver.rs
```diff
@@ -6,6 +6,7 @@ use std::collections::HashSet;
 
 use crate::bindgen::ir::Path;
 
+#[derive(Default)]
```
emilio, line 9: We should remove `DeclarationTypeResolver::new()` then.

src/bindgen/parser.rs
```diff
@@ -98,8 +96,10 @@ struct Parser<'a> {
 }
 
 impl<'a> Parser<'a> {
-    fn should_parse_dependency(&self, pkg_name: &String) -> bool {
-        if self.parsed_crates.contains(pkg_name) {
+    fn should_parse_dependency(&self, pkg_name: &str) -> bool {
+        let pkg_name = pkg_name.to_string();
```
emilio, line 100: This allocates for no good reason, while before it didn't. I agree that `&String` is generally useless, but this is worse, so let's annotate it with `#[allow(..)]` instead, and maybe change the `Vec::contains` calls in a follow-up to `.iter().any()` or something else that allows us to use a `&str`.

src/bindgen/builder.rs
```diff
@@ -12,7 +12,7 @@ use crate::bindgen::library::Library;
 use crate::bindgen::parser::{self, Parse};
 
 /// A builder for generating a bindings header.
-#[derive(Debug, Clone)]
+#[derive(Debug, Default, Clone)]
```
emilio, line 15: I think adding this is a mistake. `Default` doesn't do the same as ::new(), because of `std_types`. Can we disable that clippy warning altogether?

src/bindgen/ir/item.rs
```diff
@@ -73,7 +73,7 @@ pub enum ItemValue<T: Item> {
     Single(T),
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Default, Clone)]
```
emilio, line 76: Same, let's remove `::new()`, or make `new` just call `Default::default()`?

emilio (changes requested):
Thanks! This looks good with a few minor changes.

emilio (approved):
Thanks!

Luni-4:
Thanks to you for merging my PR! :)

PR #575, opened 2020-09-22
Pascal case primitives

emilio:
This squashes #565 and cleans up a bit on top.


adamski:
Thanks for taking care of this and merging 👍

Issue #149, opened 2018-03-10
Do not export private fields


RazrFalcon:
```rust
#[repr(C)]
pub struct my_struct(SomePrivateType);
```
Will be transformed to:
```c
typedef struct {
    SomePrivateType _0;
} my_struct;
```

But should be (I guess):
```c
typedef struct my_struct my_struct;
```

RReverser:
What if there is also a public field? I guess instead of making struct opaque it's better to transform these private fields to byte array or something... but then, C doesn't have public/private modifiers anyway, so this shouldn't matter or help much.

eqrion:
I don't think we should make the struct opaque in this case, as that'd force you to always use it behind a pointer. Which could be unnecessary.

Exporting it as a byte blob with alignment could be an option, but I'm unsure if that's worth the effort.

PR #579, opened 2020-09-24
feat(ir) Add ability to ignore specific struct fields

Hywan:
With this PR, it is now possible to add `/// cbindgen:ignore` on specific struct fields to ignore them.

Thus:

```rust
pub struct OneFieldIgnored {
    x: i32,
    /// cbindgen:ignore
    y: i64,
    z: i32,
}
```

is bound to:

```c
typedef struct {
    int32_t x;
    int32_t z;
} OneFieldIgnored;
```

It is useful for a project I'm working on, but it can also be helpful for issue like https://github.com/eqrion/cbindgen/issues/149.

Thoughts?

### DIFF
--- a/src/bindgen/ir/structure.rs
+++ b/src/bindgen/ir/structure.rs
@@ -16,7 +16,7 @@ use crate::bindgen::mangle;
 use crate::bindgen::monomorph::Monomorphs;
 use crate::bindgen::rename::{IdentifierType, RenameRule};
 use crate::bindgen::reserved;
-use crate::bindgen::utilities::IterHelpers;
+use crate::bindgen::utilities::{IterHelpers, SynAttributeHelpers};
 use crate::bindgen::writer::{ListType, Source, SourceWriter};
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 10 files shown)


emilio:
How is this useful? `OneFieldIgnored` would have different layout in C than in rust, causing at best just confusion (`z` from C would be `y`), and at worst stack overruns and segfaults.

Hywan:
Imagine you have a type ```typedef struct S S;` that you have to implement in Rust. It is created by a `s_new()` function, and deleted by `s_delete(s *s)` function.

But maybe, in Rust, you need to attach some fields to it, like `struct S { inner: some_crate::S }`. You don't want cbindgen to create a binding for `inner`, whether it is a public or a private field.

I don't have a usecase for multiple fields actually, just a struct with one field.

From the point of view of C, `S` is an opaque type in this case.

emilio:
Ok, that does make sense to me. Then I think this is the wrong thing to do. An `opaque` annotation for which we don't generate any fields or such would be slightly better. Cbindgen already has the concept of opaque types.

emilio:
Though at that point why marking the struct as `#[repr(C)]` to begin with? cbindgen will treat it as opaque if you don't, and that will allow you to get rust layout optimizations and so on in your struct, so it seems even better?

Hywan:
I agree it makes no sense. Sorry!

Hywan:
@emilio 

> An opaque annotation for which we don't generate any fields or such would be slightly better. Cbindgen already has the concept of opaque types.

How do you generate an opaque type with cbindgen?

emilio:
Generally, by making it not `#[repr(C)]`.

petrochenkov:
I wanted to use this for a union like this:
```
union U {
    a: [u64; 2],
    b: u128,
}
```

cbindgen cannot produce anything useful for `u128`, so it's ok to make this field unavailable from the C side while keeping all the other fields available.
I guess I'll have use the "typical cbindgen workaround" instead.
```rust
/// cbindgen:ignore
union U {
    a: [u64; 2],
    b: u128,
}

union U_FFI {
    a: [u64; 2],
}
```
```toml
[export]
include = ["U_FFI"]
[export.rename]
U = "U_FFI"
```

Issue #580, opened 2020-09-28
Add a generated stringification method


staktrace:
I'd like to add the ability for cbindgen to generate a stringifier for enums and structs. It could be controlled by a `derive-ostream` option similar to `derive-eq` and friends. For enums it would generate something like:

```
std::ostream& operator<<(std::ostream& aStream, const MyEnum& aEnum) {
  switch (aEnum) {
    case One:
      aStream << "One";
      break;
    ...
  }
  return aStream;
}
```

and for structs it would do something like this:

```
struct MyStruct {
  int x;
  int y;
  ...
  friend std::ostream& operator<<(std::ostream& aStream, const MyStruct& aStruct) {
    return aStream << "{ x=" << aStruct.x << ", y=" << aStruct.y << " }";
  }
  ...
};
```

Any objections/suggestions? The goal here is to remove the hand-written [stuff](https://searchfox.org/mozilla-central/rev/f27594d62e7f1d57626889255ce6a3071d67209f/gfx/layers/wr/WebRenderLayersLogging.cpp#14-70) that we have in m-c as part of bug the cleanup I'm doing in bug 1666802 and friends.

emilio:
Seems sensible to me. ADTs may be a bit funnier but nothing too hard I think :)

staktrace:
This is fixed by #582 and followups.

PR #582, opened 2020-09-30
Add ostream serializer

staktrace:
Fixes part of #580. Tagged enums are not yet supported. I'll try to get that working but wanted put these patches up for initial feedback.

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -8,8 +8,8 @@ use crate::bindgen::config::{Config, Language};
 use crate::bindgen::declarationtyperesolver::DeclarationTypeResolver;
 use crate::bindgen::dependencies::Dependencies;
 use crate::bindgen::ir::{
-    AnnotationSet, Cfg, ConditionWrite, Documentation, GenericParams, GenericPath, Item,
-    ItemContainer, Path, Repr, ReprStyle, ReprType, Struct, ToCondition, Type,
+    AnnotationSet, AnnotationValue, Cfg, ConditionWrite, Documentation, GenericParams, GenericPath,
+    Item, ItemContainer, Path, Repr, ReprStyle, ReprType, Struct, ToCondition, Type,
 };
 use crate::bindgen::library::Library;
 use crate::bindgen::mangle;
@@ -100,6 +100,7 @@ impl EnumVariant {
         generic_params: GenericParams,
         mod_cfg: Option<&Cfg>,
         self_path: &Path,
+        enum_annotations: &AnnotationSet,
     ) -> Result<Self, String> {
         let discriminant = match variant.discriminant {
             Some((_, ref expr)) => match value_from_expr(expr) {
@@ -142,7 +143,10 @@ impl EnumVariant {
         }
 
         let variant_cfg = Cfg::append(mod_cfg, Cfg::load(&variant.attrs));
-        let annotations = AnnotationSet::load(&variant.attrs)?;
+        let mut annotations = AnnotationSet::load(&variant.attrs)?;
+        if let Some(b) = enum_annotations.bool("derive-ostream") {
+            annotations.add("derive-ostream", AnnotationValue::Bool(b));
+        }
         let body = match variant.fields {
             syn::Fields::Unit => VariantBody::Empty(annotations),
             syn::Fields::Named(ref fields) => {
@@ -333,6 +337,8 @@ impl Enum {
         let mut variants = Vec::new();
         let mut is_tagged = false;
 
+        let annotations = AnnotationSet::load(&item.attrs)?;
+
         for variant in item.variants.iter() {
             let variant = EnumVariant::load(
                 repr.style == ReprStyle::Rust,
@@ -340,12 +346,11 @@ impl Enum {
                 generic_params.clone(),
                 mod_cfg,
                 &path,
+                &annotations,
             )?;
             is_tagged = is_tagged || !variant.body.is_empty();
             variants.push(variant);
         }
-
-        let annotations = AnnotationSet::load(&item.attrs)?;
 
         if let Some(names) = annotations.list("enum-trailing-values") {
             for name in names {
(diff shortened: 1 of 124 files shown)


staktrace:
Added some support for tagged enums, at least the ones that get turned into unions. I haven't tested the other case yet (where it gets turned into structs with a separated tag).

staktrace:
Ok, added a patch for the tagged enums with separated tags. Stuff passes locally, haven't tried it on Firefox code yet.

staktrace:
Needed a small tweak, but now ready for review. @emilio or @eqrion

src/bindgen/ir/structure.rs
```diff
@@ -549,6 +549,40 @@ impl Source for Struct {
...
+                    .function
+                    .rename_args
+                    .apply("stream", IdentifierType::FunctionArg);
+                let instance = config
+                    .function
+                    .rename_args
+                    .apply("instance", IdentifierType::FunctionArg);
+                write!(
+                    out,
+                    "friend std::ostream& operator<<(std::ostream& {}, const {}& {})",
```
emilio, line 569: I assume this doesn't need to be inline because it's in the class definition right?

src/bindgen/ir/enumeration.rs
```diff
@@ -764,6 +776,56 @@ impl Source for Enum {
...
+                write!(out, "switch ({}.tag)", instance);
+                out.open_brace();
+                let vec: Vec<_> = self
+                    .variants
+                    .iter()
+                    .map(|x| {
+                        if let VariantBody::Body { ref name, .. } = x.body {
+                            format!(
+                                "case {}::{}::{}: {} << {}.{}; break;",
+                                self.export_name(),
```
emilio, line 804: Do we need the export_name() here? we don't generally use it in other derived code I think.

tests/expectations/derive_ostream.cpp
```diff
@@ -47,8 +47,62 @@ struct D {
...
+union F {
+  enum class Tag : uint8_t {
+    Foo,
+    Bar,
+    Baz,
+  };
+
+  friend std::ostream& operator<<(std::ostream& stream, const Tag& instance) {
+    switch (instance) {
+      case F::Tag::Foo: stream << "Foo"; break;
```
emilio, line 59: Yeah, it seems a bit weird to use `Tag` in the declaration but use `F::Tag` here. Is this required? If so it'd be good to have a comment.

emilio (commented):
Just some very minor nitpicks, but looks reasonable otherwise.

src/bindgen/ir/structure.rs
staktrace: Yep, I can add a comment about that.

tests/expectations/derive_ostream.cpp
staktrace: Yeah the compiler was complaining about not knowing what `Tag` was if I just used `Tag::Foo` here. I was kinda surprised by it but didn't spend much time trying to figure it out. I can add a comment about it.

src/bindgen/ir/enumeration.rs
staktrace: (Assuming this comment goes with the other comment on the expectations file, where the generated code is `F::Tag::Foo` instead of `Tag::Foo` - in which case, yes, I needed this to satisfy the compiler)

staktrace:
Updated with a couple of comments clarifying the things pointed out in review.

src/bindgen/ir/enumeration.rs
```diff
@@ -142,7 +143,10 @@ impl EnumVariant {
         }
 
         let variant_cfg = Cfg::append(mod_cfg, Cfg::load(&variant.attrs));
-        let annotations = AnnotationSet::load(&variant.attrs)?;
+        let mut annotations = AnnotationSet::load(&variant.attrs)?;
+        if let Some(b) = enum_annotations.bool("derive-ostream") {
+            annotations.add("derive-ostream", AnnotationValue::Bool(b));
```
emilio, line 148: Maybe this should only add if the annotation is not present on the variant?

emilio (approved):
> Yeah the compiler was complaining about not knowing what Tag was if I just used Tag::Foo here. I was kinda surprised by it but didn't spend much time trying to figure it out. I can add a comment about it.

Wat, C++ name resolution rules won't ever stop surprising me... This looks good, I have a minor follow-up that I'd like to run through you, plus my nit below. But we can merge this as-is and iterate on another PR.

src/bindgen/ir/enumeration.rs
staktrace: Just to enumerate all the cases as I thought through this:
- in the general case, if the enum has `derive-ostream=true` and one of the variant bodies has `derive-ostream=false`, that's not going to compile because the serializer function generated for the enum will try to use the serializer function for the variant body, which won't exist. I guess the user could define it somewhere else, and so we could allow it on those grounds. If they're explicitly setting the annotation on the variant body then we can assume they know what they're doing.
- the rest of the cases below refer to the other interesting scenario, where the enum has `derive-ostream=false` and the variant inside has `derive-ostream=true`
-- the untagged enum case, all the variants are going to be of type `VariantBody::Empty`, and we don't even read the derive-ostream annotation from those. So yes, I can avoid adding the annotation in that case (or for all `VariantBody::Empty` cases in general)
-- the tagged enum case with `separate_tag = true`: in this case we can generate the serializer for the struct that represents the variant body, regardless of whether or not the enclosing enum has a serializer. So in this case, yes, I can avoid propagating the annotation down
-- the tagged enum case with `separate_tag = false`: in this case the variant body contains a field that is the tag, so if we are generating a serializer for the variant body, then we need the tag serialized as well, otherwise there will be a compilation failure. But that scenario is actually already broken in the current patch if the enum doesn't explicitly set `derive-ostream=false` but instead just has a default of false. So I guess this scenario is not made any worse. And again, this is a scenario where the user did not set the annotation on the enum but instead on a specific variant body, so maybe we can assume they know what they're doing.

So yes overall your suggestion makes sense. I'll submit another PR with a patch.

staktrace: https://github.com/eqrion/cbindgen/pull/585

PR #585, opened 2020-10-01
Follow-up to #582 to address review comment.

staktrace:
This avoids propagating the derive-ostream annotation from an enum to
the variant body if the variant body already has an explicit annotation
set. This allows for more possibilities ending in compilation failure
but generally only when the user explicitly sets these annotations without
defining necessary serializers themselves. We should be able to assume
the user knows what they're doing if they trigger these particular scenarios.


staktrace:
Oh I only just saw that you made a PR which does a better job of this. Whoops.

PR #587, opened 2020-10-03
Support `sort_by` for constants

petrochenkov:
Also support a global default for `sort_by` affecting both functions and constants.

Other items are sorted according to their dependency graph, as I understand, so this option is not introduced for them.

UPD: The default sorting order for all items is also switched to "None".


petrochenkov:
Ok, replaced `mem::replace` with `mem::take` like clippy on CI wants, hopefully this won't change the minimal supported version.

docs.md
```diff
@@ -885,6 +885,13 @@ allow_static_const = true
 # default: false
 allow_constexpr = false
 
+# This rule specifies if the order of functions will be sorted in some way.
```
emilio, line 888: s/function/constants, right?

src/bindgen/ir/item.rs
```diff
@@ -75,18 +75,20 @@ pub enum ItemValue<T: Item> {
 
 #[derive(Debug, Clone)]
 pub struct ItemMap<T: Item> {
-    data: BTreeMap<Path, ItemValue<T>>,
+    data: IndexMap<Path, ItemValue<T>>,
```
emilio, line 78: Why is `IndexMap` needed? Not necessarily opposed to this, but unless I'm missing something this change is kind of unrelated and should probably be done in a different PR.

emilio:
My other question is... this doesn't deal with constants that depend on other constants right?

So something like:

```rust
pub const B: usize = 4;
pub const A: usize = B +1;
```

Will be mishandled in C++ mode (in C we kinda get by because C macros)...

Looking a bit more into it, this is already a problem I guess. But given that, I think introducing this option may not be great...

I guess given it's opt-in is possibly fine but...

emilio (commented):
Ah, I see... so `IndexMap` is needed to truly preserve input order (it seems this has been a BTreeMap since forever, d'oh).

Ok. So with that out of the way... my only real concern is whether `SortKey::Name` is the right default order for constants, specially given we don't track their interdependencies right (https://github.com/eqrion/cbindgen/issues/592). I think `SortKey::None` may be a more sensible default order given that.

petrochenkov:
>Ah, I see... so IndexMap is needed to truly preserve input order (it seems this has been a BTreeMap since forever, d'oh).

Yes, `IndexMap` preserves the original order (`sort_by = "None"`) while having the same interface as `BTreeMap` which loses the original order and can only maintain `sort_by = "Name"`.

>My other question is... this doesn't deal with constants that depend on other constants right?

This PR actually improves the situation with https://github.com/eqrion/cbindgen/issues/592 because with `sort_by = "None"` you can now fix the dependencies issues by reordering Rust code if necessary.

>my only real concern is whether SortKey::Name is the right default order for constants

I also think `sort_by = "None"` is a better default, perhaps even for functions.
I kept the old defaults just to maximize backward compatibility and minimize the diff.
I'll change it if you think it's the right thing to do.

emilio:
> This PR actually improves the situation with #592 because with `sort_by = "None"` you can now fix the dependencies issues by reordering Rust code if necessary.

Right, that's great.

> > my only real concern is whether SortKey::Name is the right default order for constants
> 
> I also think `sort_by = "None"` is a better default, perhaps even for functions.
> I kept the old defaults just to maximize backward compatibility and minimize the diff.
> I'll change it if you think it's the right thing to do.

Yeah, I think reordering constants in a single source file unexpectedly is wrong. Let's make `sort_by = "None"` the default, I'll make sure to note it in the release notes. Thank you!

petrochenkov:
Updated.

emilio (approved):
LGTM, thanks!

Issue #583, opened 2020-09-30
"Unhandled const definition" on a constant with a type alias (typedef) type


petrochenkov:
```rust
type MyInt = i32;

const MY_CONST: MyInt = 10;
```

`cbindgen` will produce a "Unhandled const definition" error on `MY_CONST` because `MyInt` is not a "`can_handle`" type
https://github.com/eqrion/cbindgen/blob/1fc4cb072422e722f8bcb3af5766f84587a180c6/src/bindgen/ir/constant.rs#L393-L395
https://github.com/eqrion/cbindgen/blob/1fc4cb072422e722f8bcb3af5766f84587a180c6/src/bindgen/ir/constant.rs#L366-L374

The "`can_handle`" restriction looks artificial and unnecessary, if it is commented out, then the correct constant definition is generated.
```
// C++
static const MyInt MY_CONST = 10;

// C
#define MY_CONST 10
```
In both cases it doesn't matter what `MyInt` actually is, cbindgen just can put it into the constant definition textually and it will work in most cases if the right hand side can be converted to C/C++.

emilio:
Yeah, I think we should tweak `can_handle` to handle the kind of expressions we actually support rather than doing it based on the type of the expression...

PR #588, opened 2020-10-03
Support constants with arbitrary types

petrochenkov:
When converting `const` items migrate from whitelist to blacklist approach to the constant's type.
As long as we can convert the right hand side with `Literal::load` we assume that we can convert the left hand side as well, perhaps using user-defined types defined in the crate.

The one exception is `str` type.
Despite being able to convert string literals with `Literal::load`, we cannot convert `str`.
We can't even introduce an opaque type for it because it's unsized and pointers to it are fat.

Closes https://github.com/eqrion/cbindgen/issues/583

### DIFF
--- a/src/bindgen/ir/constant.rs
+++ b/src/bindgen/ir/constant.rs
@@ -363,14 +363,19 @@ pub struct Constant {
     pub associated_to: Option<Path>,
 }
 
-fn can_handle(ty: &Type, expr: &syn::Expr) -> bool {
-    if ty.is_primitive_or_ptr_primitive() {
-        return true;
-    }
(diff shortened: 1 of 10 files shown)


petrochenkov:
Alternatively, we could just reject string literals on the right side.

That would perhaps be even more robust in the face of cases like
```rust
pub type A = &'static str;
pub const C: A = "my string";
```

petrochenkov:
>Alternatively, we could just reject string literals on the right side.

That approach is simpler as well.
It's implemented in https://github.com/eqrion/cbindgen/pull/589, closing this PR in favor of that.

PR #589, opened 2020-10-04
Support constants with arbitrary types (take 2)

petrochenkov:
(Take 1 - https://github.com/eqrion/cbindgen/pull/588.)

When converting `const` items stop checking their types and check only values instead.
As long as we can convert the right hand side with `Literal::load` we assume that we can convert the left hand side as well, perhaps using user-defined types defined in the crate.

The one exception is `str` type.
Despite being able to convert string literals with `Literal::load`, we cannot convert `str`.
We can't even introduce an opaque type for it because it's unsized and pointers to it are fat.
So we just remove the support for string literals from `Literal::load`, processing of `const` items is the only place where it is used anyway.

Closes https://github.com/eqrion/cbindgen/issues/583

### DIFF
--- a/tests/rust/constant_user_defined_type.rs
+++ b/tests/rust/constant_user_defined_type.rs
@@ -1,0 +1,17 @@
+#[repr(C)]
+pub struct S {
+    field: u8,
+}
+
+/// cbindgen:enum-class=false
+#[repr(C)]
+pub enum E {
+    V,
+}
+use E::*;
+
+pub type A = u8;
+
+pub const C1: S = S { field: 0 };
+pub const C2: E = V;
+pub const C3: A = 0;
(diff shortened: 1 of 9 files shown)


tests/rust/constant_user_defined_type.rs
```diff
@@ -0,0 +1,17 @@
+#[repr(C)]
+pub struct S {
+    field: u8,
+}
+
+/// cbindgen:enum-class=false
```
emilio, line 6: In the future we should probably also allow paths to be constants to support regular enum classes, but this is ok for now.

emilio (approved):
Heh, interestingly this causes Firefox to stop building because it uncovers a case of https://github.com/eqrion/cbindgen/pull/587#issuecomment-703466832 :-)

But this is clearly an improvement.

The `Str` special-case is there since d38ffa5f, but we don't generate anything for it so removing is fine.

PR #590, opened 2020-10-04
Support generation of Cython bindings

petrochenkov:
[Cython](https://cython.org) [[wiki](https://en.wikipedia.org/wiki/Cython)] is a language extension for Python that allows writing C-based Python packages in a more humane way (that includes easier reuse of third-party C libraries).

Cython "headers" (`.pxd` files) are basically C headers re-syntaxed in a more Python-like way.
So cbindgen seems to be a good place to implement generation of such bindings from Rust code.
As you can see from the patch, it mostly requires changes to the cbindgen "backend" that generates the resulting text.
(The patch is better read with whitespace differences ignored.)


emilio (commented):
This looks generally reasonable. As long as it is well tested and such I think there would be no particular issue with this.

petrochenkov:
Great! I'll work on the testing infra then.

petrochenkov:
Status update:
- The testing infra is ready
- 100% tests pass the "Rust -> Cython" step
- 43% tests pass the "Cython -> C" step

petrochenkov:
Status update:
- The whole test suite passes.

Remaining items:
- [ ] Figure out why `cython` is mad when its input file extension is `.pxd`.
- [ ] Re-enable doc comments in the generated files.
- [ ] Add some docs.
- [ ] Cleanup code.

petrochenkov:
@emilio 
This should be ready now.

The original PR description still apply:
>Cython "headers" (`.pxd` files) are basically C headers re-syntaxed in a more Python-like way.
>As you can see from the patch, it mostly requires changes to the cbindgen "backend" that generates the resulting text.

The only additional "general idea" of the implementation is that C bindings (the actual implementation) and Cython bindings (its "bridge" to Python) for a library are supposed to be generated together and then built into a single Python package.
So, they should ideally be producible from a single cbindgen config.

As a result we do not error on things like `non_null` or `no_return` inexpressible in Cython, but rather omit them from generated code.
Things like `#include ...` are also omitted, not due to producing Cython compilation errors (they are comments in Python syntax), but rather to avoid noise.

petrochenkov:
This PR is based on top of https://github.com/eqrion/cbindgen/pull/612 and https://github.com/eqrion/cbindgen/pull/613 which also touch some code not directly related to Cython.

petrochenkov:
This is going to conflict with the bitfield PR pretty significantly, I'll rebase.

emilio (commented):
So this looks pretty sensible over-all. I need to check all the changes a bit more closely, but I wanted to point out that when I run `cargo test` with this patch applied, I get a few diffs in the `linestyle_cr*` tests, see https://crisal.io/tmp/cbindgen-cython-diff.txt (github choked at uploading this file lol). It's newline changes, actually adding `crlf` to the `crlf` tests...

petrochenkov:
>when I run cargo test with this patch applied, I get a few diffs in the linestyle_cr* tests ... It's newline changes, actually adding crlf to the crlf tests...

Hmm, it doesn't happen on CI or on my machine.
Perhaps some configuration issue with your git setup?
In theory the configuration in https://github.com/eqrion/cbindgen/blob/master/.gitattributes should override global options, but maybe it doesn't always happen.

emilio:
> > when I run cargo test with this patch applied, I get a few diffs in the linestyle_cr* tests ... It's newline changes, actually adding crlf to the crlf tests...
> 
> Hmm, it doesn't happen on CI or on my machine.
> Perhaps some configuration issue with your git setup?
> In theory the configuration in https://github.com/eqrion/cbindgen/blob/master/.gitattributes should override global options, but maybe it doesn't always happen.

Right, I wonder why it only happens with these tests rather than with the existing ones...

> Or perhaps the repo needs to be "re-initialized" in some sense for `.gitattributes` to start having effect.
> UPD: Couldn't reproduce this on a fresh repo checkout locally.

I could repro this on a fresh checkout, and made sure that there's no global config related to line feeds on my machine... :/

Will try to repro on a vm or something to try to figure out, this is quite weird.

src/bindgen/ir/constant.rs
```diff
@@ -553,17 +561,22 @@ impl Constant {
 
         if (config.constant.allow_static_const || allow_constexpr)
             && config.language == Language::Cxx
+            || config.language == Language::Cython
```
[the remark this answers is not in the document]
petrochenkov: Done.

petrochenkov:
ping @emilio

emilio:
Sorry for the lag, last week was pretty crazy :/

Ok, I checked the tests, and the issue was that `git am` (or GitHub's patch export) was somehow messing up with the newlines. So if I apply this PR like this:

```
$ curl -L https://github.com/eqrion/cbindgen/pull/590.patch | git am -
```

Somehow the crlf test doesn't end up with the right cr characters. But if I directly check out your branch then everything looks fine here too. I'll try to finish the review today, thanks for the patience.

src/bindgen/writer.rs
```diff
@@ -153,28 +153,42 @@ impl<'a, F: Write> SourceWriter<'a, F> {
...
-        self.new_line();
-        if semicolon {
-            self.write("};");
-        } else {
-            self.write("}");
+        match self.bindings.config.language {
+            Language::Cxx | Language::C => {
+                self.pop_tab();
+                self.new_line();
+                if semicolon {
```
emilio, line 183: This could be simplified moving the pop_tab to above the match and such. But followup is fine.

tests/rust/mod_attr.cython.toml
```diff
@@ -0,0 +1,9 @@
+header = """
+DEF FOO = 0
+DEF BAR = 0
```
emilio, line 3: Is there a chance we wouldn't need these, by maybe doing something smart like:

```
#ifdef __cython__
DEF FOO = 0
DEF BAR = 0
#endif 
```

I believe that should make both cython and C/C++ happy with the same toml file, right?

tests/rust/prefix.cython.toml
```diff
@@ -0,0 +1,5 @@
+[export]
+prefix = "PREFIX_"
+exclude = [
+  "Y", # right hand side crashes Cython compiler
```
emilio, line 4: Ok, this one seems harder to workaround :)

tests/rust/manuallydrop.cython.toml
```diff
@@ -0,0 +1,4 @@
+[export]
+exclude = [
```
emilio, line 2: Hmm, yeah, this one seems also a bit unfortunate... I wonder if there's something we could do to make these unnecessary.

emilio (commented):
So the code changes look good.

It's a bit of a bummer to have to have different configuration files for cython.

How crazy does it seem to run the C preprocessor through the cbindgen output to remove all the ifdefs, and pass `-DCBINDGEN_CYTHON` so that we could stash cython-specific stuff there?

Seems like that should avoid most of the duplicate config files which are needed because they contain something like `#ifdef __cplusplus /* do something smart */ #endif`.

Some of the others look a bit harder but...

emilio (approved):
Let's get this in and follow-up on separate PRs.

petrochenkov:
Great!
I'll look at ["Remove most of the cython-specific config files by running the preprocessor."](https://github.com/eqrion/cbindgen/pull/618) tomorrow.

I submitted one more Cython-related PR which I had stashed for some time - https://github.com/eqrion/cbindgen/pull/619.

PR #595, opened 2020-10-10
Add missing function `with_style` to the Builder

yurymik:
It's impossible to select desired style when `cbindgen` is configured from the build script. This PR adds the missing function.

I happen to notice that none of the tests even try to create a `Builder` object, so I wasn't sure what style of tests is preferable here. I've created a trivial one just to hit newly added lines, please let me know if it has to be changed to some other format.

### DIFF
--- a/src/bindgen/builder.rs
+++ b/src/bindgen/builder.rs
@@ -6,7 +6,7 @@ use std::path;
 
 use crate::bindgen::bindings::Bindings;
 use crate::bindgen::cargo::Cargo;
-use crate::bindgen::config::{Braces, Config, Language};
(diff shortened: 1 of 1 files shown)


yurymik:
Clippy fails with errors in 

```
src/bindgen/cdecl.rs:28:9
src/bindgen/ir/ty.rs:170:9
src/bindgen/ir/ty.rs:380:41
src/bindgen/parser.rs:457:77
```

and none of these files was modified in this PR.

emilio (approved):
This looks fine, thanks!

(You can technically do `with_config(Config { style, ...Default::default() })` or something but...)

Yeah, it seems rust stable got an update and the new clippy is complaining more...

PR #600, opened 2020-10-13
CI: Add a minimum Rust version check?

petrochenkov:
There's no clippy check forcing new features now (fbdad17dc630bd2e0cf7975c879a6738e3949074), so the minimum supported Rust version is currently enforced by dependencies (`remove_dir_all-0.5.3`) and should be relatively stable.
Should it be checked on CI to avoid bumping the version requirements accidentally?


emilio (approved):
Yes, agreed, this is great, thank you!

PR #602, opened 2020-10-17
test suite: Use suffixes instead of directories for test variants

petrochenkov:
Implement the suggestion from https://github.com/eqrion/cbindgen/pull/598#issuecomment-710907817.


tests/tests.rs
```diff
@@ -144,32 +144,27 @@ fn run_compile_test(
...
-        Language::Cxx => "cpp",
-        Language::C => {
-            if cpp_compat {
-                "compat.c"
-            } else {
-                "c"
-            }
-        }
+    let compat_ext = match (language, cpp_compat) {
+        (Language::C, false) => ".nocompat",
```
emilio, line 149: I think this would be slightly nicer keeping `.compat.` as the special case rather than the other way around.

emilio (commented):
This looks fine to me, I agree this makes stuff more obvious.

I'm happy to merge this with the `[WIP]` removed and `.compat` being the special extension rather than the other way around.

petrochenkov:
Updated.

emilio (approved):
Sweet, thank you!

Issue #485, opened 2020-03-08
References in enum fields not supported?


drlagos:
The docs say references are supported and translated to pointer types, but it does not say under which conditions that is the case. I am trying to expose the following tagged union:

```rust
#[repr(C)]
pub enum BoolOp<'a> {
    Lit {value: bool},
    And {left: &'a BoolOp<'a>, right: &'a BoolOp<'a>},
}
```

This generates nothing in the exported header, not even an opaque type. The type only shows up in function parameters and return types. 

I'd rather not replace the references with pointers and ruin the ergonomics on the Rust side of things. What should I do?

drlagos:
I just checked and it works if I give the references the static lifetime. That's not great from a memory management POV...

emilio:
https://github.com/eqrion/cbindgen/blob/5a4d74b9118928946dd10a49c4891a2610c55437/src/bindgen/parser.rs#L897

That check has been there since forever, probably a matter of removing it and adding a test nowadays.

petrochenkov:
Fixed in https://github.com/eqrion/cbindgen/pull/604.

PR #604, opened 2020-10-20
Remove artificial restriction on lifetime parameters on enums

petrochenkov:
Fixes https://github.com/eqrion/cbindgen/issues/485


emilio (approved):
Thanks!

Issue #239, opened 2018-10-31
Is it possible to map usize to size_t and isize to ptrdiff_t?


lu-zero:
I'm not sure if there is a mean to annotate to do this mapping.

RReverser:
Why do you want this? Changing usize to size_t would be incorrect and might break on certain platforms.

lu-zero:
Which platforms exactly?

IIRC `uintptr` is an optional type and could be missing or being different from the register size (e.g. x32 and mips hybrids)

RReverser:
You might want to check answers to same question previously asked in https://github.com/eqrion/cbindgen/issues/198. Rust's `usize` is explicitly defined as pointer-wide unsigned integer, which maps only to `uintptr_t`, but C and C++ define `size_t` differently which means they can mismatch in sizes.

lu-zero:
The way we use usize to hold `len()` and `size_of()` values, makes it a little more confusing. 

I'm not even sure rust itself supports architectures with `sizeof(size_t) != sizeof(uintptr_t)` and `sizeof(ptrdiff_t) != sizeof(intptr_t)`.

Given that intptr_t is optional and makes C people complain because looks weird to them, I'd rather map it to size_t as long the two are compatible, thus why I'd rather have a way to annotate the type so it maps to what the user expects (size_t for array lengths and ptrdiff_t for strides).

The defaults can stay as they are :)

RReverser:
> I'm not even sure rust itself supports architectures with sizeof(size_t) != sizeof(uintptr_t) and sizeof(ptrdiff_t) != sizeof(intptr_t).

I'm pretty sure it does, it just doesn't have equivalent of `size_t` and instead uses pointer-sized integers for everything, as you noticed. However, when translation to C/C++, it's important to follow their standards, not Rust ones, to be actually safely compatible across all possible platforms where Rust might be used via FFI bindings.

> Given that intptr_t is optional and makes C people complain because looks weird to them

That sounds like a strange personal preference, not something to base choice of type representation on, especially because

> as long the two are compatible

is not actually guaranteed at all and causes user to rely on UB, which can go really wrong.

lu-zero:
My problem is getting the users what they expect, I have no problems in adding a build/configure time checks to make sure the incompatible platforms are managed and then use the right type and then annotate it so it is expressed as they want.

lu-zero:
[...]

PR #606, opened 2020-10-21
Add an option for converting usize/isize into size_t/ptrdiff_t

petrochenkov:
This PR adds a new global option `usize_is_size_t` that mirrors `size_t_is_usize` in bindgen and allows to map `usize`/`isize` into the more traditional pair `size_t`/`ptrdiff_t` instead of `uintptr_t`/`intptr_t` into which they are converted currently.

(I'd personally want to see this as the default, similarly to https://github.com/rust-lang/rust-bindgen/pull/1902, but I didn't change any defaults in this PR.)

cc https://github.com/eqrion/cbindgen/issues/239

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -691,6 +691,10 @@ rename_args = "PascalCase"
 # default: "None"
 sort_by = "Name"
 
+# If this option is true `usize` and `isize` will be converted into `size_t` and `ptrdiff_t`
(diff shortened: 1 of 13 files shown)


emilio (approved):
Sorry, I had missed the mail for this one, looks good to me.

kekeimiku:
why `ptrdiff_t` instead of `ssize_t`?

PR #607, opened 2020-10-22
Allow controlling the Cargo profile used for macro expansion

### LINKED ISSUE #603: Allow specifying a profile for `cargo expand`
When `parse.expand` is enabled, cbindgen invokes Cargo to expand macros in the specified crates. If the dependencies of the crates in question have already been built, Cargo is smart enough to not build them from scratch. However, if the dependencies were only built in release mode, Cargo won't see them as valid build artifacts, and *will* rebuild everything (using the debug profile) just to do the expansion. If the `cargo` invocation in `cargo_expand.rs` allowed passing `--release`, this extra build could be avoided.

jrose-signal:
If there's already a release build, it's better for cbindgen to reuse the build artifacts from that to expand macros rather than starting from scratch with a debug build. Controlled with `--profile (debug|release)` as well as `parse.expand.profile` in cbindgen.toml, though hardcoding a profile in a config file seems unlikely.

Fixes #603. Happy to adjust per feedback.

### DIFF
--- a/src/bindgen/cargo/cargo_expand.rs
+++ b/src/bindgen/cargo/cargo_expand.rs
@@ -24,6 +24,14 @@ pub enum Error {
     Compile(String),
 }
 
+/// Which Cargo profile (group) to use when expanding macros.
+pub enum Profile {
+    /// Do not pass `--release` when expanding macros
+    Debug,
+    /// Pass `--release` when expanding macros
+    Release,
+}
+
 impl From<io::Error> for Error {
     fn from(err: io::Error) -> Self {
         Error::Io(err)
@@ -65,6 +73,7 @@ pub fn expand(
     expand_all_features: bool,
     expand_default_features: bool,
     expand_features: &Option<Vec<String>>,
+    profile: Profile,
 ) -> Result<String, Error> {
     let cargo = env::var("CARGO").unwrap_or_else(|_| String::from("cargo"));
     let mut cmd = Command::new(cargo);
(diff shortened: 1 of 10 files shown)


src/bindgen/cargo/cargo_expand.rs
```diff
@@ -24,6 +24,14 @@ pub enum Error {
     Compile(String),
 }
 
+/// Which Cargo profile (group) to use when expanding macros.
+pub enum Profile {
```
emilio, line 28: Any reason to have this enum and the config.rs enum duplicated?

emilio (commented):
This looks generally good. I have to review the tests more in detail but meanwhile just one minor question.

src/bindgen/cargo/cargo_expand.rs
jrose-signal: I wasn't really sure how to do the dependencies. If the canonical version is in `cargo_expand`, then `cargo_expand` is exporting public types now; if the canonical version is in `config`, then `cargo_expand` needs to import `config`. Preference?

emilio:
Thanks, I merged it in https://github.com/eqrion/cbindgen/commit/0ba241498e4ceccf052ffa91883d53b7cdfb3a85 with a small follow-up in https://github.com/eqrion/cbindgen/commit/ef05231b5c81d5613762cb41b92e1bb2e987ab01 to remove the duplicate enum.

Thanks!

Issue #609, opened 2020-10-25
Support a cbindgen annotation for generating bitfield sizes in C code (`: SIZE`)


petrochenkov:
as described in https://github.com/eqrion/cbindgen/issues/608

```rust
struct S {
    #[cfg(not(feature = "cbindgen"))]
    foo_and_bar: u64,

    #[cfg(feature = "cbindgen")]
    /// cbindgen:bitfield="8"
    foo: u64,
    #[cfg(feature = "cbindgen")]
    /// cbindgen:bitfield="56"
    bar: u64,
}
```
will generate
```c
struct S {
    uint64_t foo: 8;
    uint64_t bar: 56;
};
```

In general, the string passed to the annotation will be copypasted into the generated field code:
```rust
/// cbindgen:bitfield="BITFIELD_STRING"
name: Type,
```
=>
```rust
Type name: BITFIELD_STRING;
```

PR #611, opened 2020-10-26
Support an annotation for generating bitfield sizes in C code (`: SIZE`)

petrochenkov:
Closes https://github.com/eqrion/cbindgen/issues/609

The first commit is a relatively large refactoring that turns `(String, Type, Documentation)` into a proper structure `Field`.


src/bindgen/ir/field.rs
```diff
@@ -0,0 +1,47 @@
...
+            })
+        } else {
+            None
+        })
+    }
+}
+
+impl Source for Field {
+    fn write<F: Write>(&self, config: &Config, out: &mut SourceWriter<F>) {
+        self.documentation.write(config, out);
```
emilio, line 47: You need to handle `config.documentation` here don't you? At least you removed similar checks in a couple places.

tests/rust/bitfield/cbindgen.toml
```diff
@@ -0,0 +1,3 @@
+[parse.expand]
```
emilio, line 1: Instead of making this a nightly-only test via parse.expand you could instead use the `defines` section with something like `"feature = cbindgen" = "CBINDGEN"` and `#define CBINDGEN` in the `header` to effectively ifdef stuff out, but your call.

src/bindgen/ir/field.rs
emilio: Ah, nevermind, it's handled in `Documentation::write`.

emilio (approved):
Looks good, just one nit, but if you feel strongly about using `expand` then that's fine for me.

tests/rust/bitfield/cbindgen.toml
petrochenkov: I think I'll keep the existing version, translation of `cfg`s into `ifdef`s doesn't currently work on fields (https://github.com/eqrion/cbindgen/pull/316), and the version with expansion is closer to reality since you don't really want those `ifdef`s in the generated code in this case.

emilio (approved):

emilio:
(There was a minor rebase conflict so I fixed it manually and pushed. Thank you!)

PR #614, opened 2020-10-30
Support wider range of expressions in enum discriminants

petrochenkov:
```rust
enum E {
    V = EXPR
}
```
now supports the same set of expressions as
```
const C: Type = EXPR;
```
with paths and binary operators probably being the most important case in practice
```
enum E {
    V = CONST,
    U = CONST + 1,
}
```


emilio (approved):
Sweet, thanks!

PR #618, opened 2020-11-17
Remove most of the cython-specific config files by running the preprocessor.

emilio:
I _think_ I'm a bit happier with this than the alternative of having to
maintain two config files.

The main changes:

 * We run the preprocessor on the cbindgen output. This is useful so
   that we can use ifdefs for cython in header/trailers.

 * If there's both `<file>.toml` and `<file>.cython.toml`, we generate a
   config file concatenating them. This is useful when cython just wants
   to avoid generating a file due to crashes or what not.


emilio:
@petrochenkov wdyt?

petrochenkov:
I'll look tomorrow.

I've been using the next trick to avoid multiple configs (for https://github.com/eqrion/cbindgen/pull/619 in particular):
```rust
after_includes = "
#ifdef __FILE__
  something c/c++
#else
  something cython
#endif
"
```
(`__FILE__` must always be defined for both C and C++), but it requires "something c/c++" to start with `#` (e.g. includes) so it doesn't break Cython.
Perhaps this can be expanded to cover `<file>.cython.toml` cases as well without running preprocessor.

petrochenkov:
Some alternative methods of solving this can be found in https://github.com/eqrion/cbindgen/pull/620.
Perhaps we can land it partially and just keep the few remaining Cython-specific configs.

I'm not too happy about running preprocessor on generated Cython code because it doesn't match workflow used in practice, and could hide things like C-style comments or something, that would fail otherwise.

emilio:
Superseded by #620.

PR #619, opened 2020-11-17
cython: Add options for specifying header and imports

petrochenkov:
C header `.h` and Cython `.pxd` are generated as a pair, so it's important to be able to generate them from a single config file.
Two new options in this PR allow to do it without hacks.
One option allows to "link" the `.pxd` to `.h`, and another one allows to add some imports.

(I don't plan to add any more Cython-specific options in the near future, these two are the ones with highest return-on-investment.)

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -979,6 +979,19 @@ features = ["cbindgen"]
 # `&mut T` and `NonNull<T>` all require a valid pointer value. 
 non_null_attribute = "_Nonnull"
 
+# Options specific to Cython bindings.
+
+[cython]
+
+# Header specified in the top level `cdef extern from header:` declaration.
+#
+# default: *
+header = '"my_header.h"'
+
+# `from module cimport name1, name2` declarations added in the same place
+# where you'd get includes in C.
+[cython.cimports]
+module = ["name1", "name2"]
 ```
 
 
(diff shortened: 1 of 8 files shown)


docs.md
```diff
@@ -979,6 +979,19 @@ features = ["cbindgen"]
...
+
+# Header specified in the top level `cdef extern from header:` declaration.
+#
+# default: *
+header = '"my_header.h"'
+
+# `from module cimport name1, name2` declarations added in the same place
+# where you'd get includes in C.
+[cython.cimports]
+module = ["name1", "name2"]
```
emilio, line 994: This looks fine though I think this could pretty much use `after_includes`.

emilio (approved):

docs.md
petrochenkov: If imports are added to `after_includes`, then you'd either have two separate configs for C and Cython (which is not desirable), or use
```
#if 0
from module cimport name1, name2
#endif
```
emitted into both `.h` and `.pxd` (which is not pretty).

emilio: Fair enough.

petrochenkov:
@emilio 
Apologies, ouput from this feature is non-reproducible due to `HashMap`, this caused the CI failure on master.
I fixed it in https://github.com/eqrion/cbindgen/pull/622.

(Some other options using `HashMap`s can potentially cause similar issues too.)

PR #620, opened 2020-11-18
tests: Remove Cython-specific configs

petrochenkov:
This is an alternative to https://github.com/eqrion/cbindgen/pull/618 that doesn't require running preprocessor or concatenating configs. Cython-specific configs are removed using a combination of several changes.

- Removing Cython-specific code from C/C++ by putting it into `#if 0`. (Anything starting with `#` is a comment in Cython, and doesn't need to be removed.)
    ```
    #if 0

    Cython-specific

    #endif
    ```
- Removing C/C++-specific code from Cython by putting it into multi-line comments (Python docstrings)... and removing the docstring quotes from C/C++, not very pretty.
    ```
    #if 0
    ''' '
    #endif

    C/C++-specific

    #if 0
    ' '''
    #endif
    ```
- Tweaking tests in insignificant ways (changing indents, removing insignificant comments).
- Tweaking tests in significant ways (simplifying constants that crash Cython compiler, only 2 tests).
- Some actual improvements to cbindgen logic allowing it to not generate invalid code.


emilio (approved):
I hate it / love it :-)

I think this is fine. A bit unfortunate to have to comment out some tests... Do you know if these bugs are filed upstream?

petrochenkov:
>Do you know if these bugs are filed upstream?

I'll try to find out, and report if they are not yet reported.

emilio:
Thanks!

petrochenkov:
Reported in https://github.com/cython/cython/issues/3918.

PR #622, opened 2020-11-25
cython: Fix a non-reproducible test

petrochenkov:
Fixes the CI failure on master - https://github.com/eqrion/cbindgen/runs/1454777165.

Also add comments for https://github.com/eqrion/cbindgen/pull/620#discussion_r526255315

UPD: Looks like other `HashMap`s in config are used for search rather than for output, so they don't introduce this kind of randomness.


emilio (approved):
Thanks!

PR #623, opened 2020-11-25
ci: Check `cargo test` with minimim supported Rust version

petrochenkov:


petrochenkov:
https://github.com/eqrion/cbindgen/pull/607 broke `cargo test` on the minimum supported version:
```
error: environment variable `CARGO_BIN_EXE_cbindgen` not defined
Error:   --> tests/profile.rs:43:25
   |
43 |     let cbindgen_path = env!("CARGO_BIN_EXE_cbindgen");
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```
cc @jrose-signal

jrose-signal:
Oops. [That's a *Rust* minimum requirement, not a dependency one.](https://blog.rust-lang.org/2020/04/23/Rust-1.43.0.html#new-cargo-environment-variable-for-tests). Should I conditionalize out this test for older Rusts? (Or use a more clunky way of finding the cbindgen executable.)

petrochenkov:
The regular test suite retrieves the path to cbindgen executable the old way, using `OUT_DIR`.
Perhaps `tests\profile.rs` could do the same thing for now?

jrose-signal:
I'll put up a PR later today.

jrose-signal:
Actually, wait, this test is going to fail regardless because it's about expansion, which only works with nightly builds. How does that work with the other expansion tests?

petrochenkov:
`cargo +1.40.0 test nonexistent-test --verbose` here doesn't run any tests, only builds in test mode.

(But when I discovered this, I did run tests, with `RUSTC_BOOTSTRAP=1` :sweat_smile:)

jrose-signal:
Ahh, of course. I didn't think through what that was going to do.

emilio:
Can you rebase this on top of master to confirm it's green? Looks good otherwise.

petrochenkov:
Rebased, CI is green.

emilio (approved):

PR #624, opened 2020-11-25
tests: Provide a fallback for CARGO_BIN_EXE_ for earlier Rusts

jrose-signal:
Unblocks #623.

### DIFF
--- a/tests/profile.rs
+++ b/tests/profile.rs
@@ -40,7 +40,20 @@ fn build_using_bin(extra_args: &[&str]) -> tempfile::TempDir {
         .tempdir()
         .expect("Creating tmp dir failed");
 
-    let cbindgen_path = env!("CARGO_BIN_EXE_cbindgen");
+    let cbindgen_path = match option_env!("CARGO_BIN_EXE_cbindgen") {
+        Some(path) => path.into(),
+        None => {
+            // We must be on an older version of Rust.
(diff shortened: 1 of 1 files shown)


jrose-signal:
Failure is unrelated to my change.

petrochenkov:
The CI failure should be fixed in https://github.com/eqrion/cbindgen/pull/622.

tests/profile.rs
```diff
@@ -40,7 +40,20 @@ fn build_using_bin(extra_args: &[&str]) -> tempfile::TempDir {
...
+        Some(path) => path.into(),
+        None => {
+            // We must be on an older version of Rust.
+            // Guess where cbindgen would be relative to OUT_DIR.
+            let mut path = PathBuf::from(env!("OUT_DIR"));
+            path.pop();
+            path.pop();
+            path.pop();
+            path.push("cbindgen");
+            path.into_os_string()
```
emilio, line 53: This is not correct with cross-compilation or what not, I think, but as a best-effort thing should do.

emilio (approved):

emilio (approved):

tests/profile.rs
jrose-signal: It's at least no worse than what the other tests are doing.

PR #625, opened 2020-11-26
cython: Leave out values of constants and enumerators

petrochenkov:
As advised in https://github.com/cython/cython/issues/3918.
They are ignored in `extern` blocks anyway (they are taken from the corresponding C headers in the resulting package).

I still kept them as comments because they are useful to humans as documentation.

This helps to re-enable some test cases removed in https://github.com/eqrion/cbindgen/pull/620.


emilio (approved):
Quite surprising that cython does nothing even with enumeration values, but sure :)

Thanks for filing it and sending the fix!

petrochenkov:
>Quite surprising that cython does nothing even with enumeration values, but sure :)

It does nothing only in `extern` declarations, which we are generating with cbindgen, and for which their implementation is taken from C.
Enumeration values in Cython implementation code are very much used.

Issue #353, opened 2019-06-09
Add a feature to simplify enums used as tagged unions


porky11:
When enum variants contain data, they will be converted to a tagged union in C.
The variants will be converted to new types with complicated long names.

In cases, where the enum variants only contain a single field, it seems preferable not to define new types for the fields of the untagged union in the C struct, representing the tagged union, but just use the single field types.

An example, where this causes unnecessary structs, is [wgpu](https://github.com/gfx-rs/wgpu).
I just described in [this issue](https://github.com/gfx-rs/wgpu/issues/219), how I'd prefer the C code to look like.

Since I don't see a reason to have single value wrapper structs in a C binding, I'd prefer it to just use the existing types by default in this case.
But I can also think of configuration options for enums.
For consistency reasons it may be more reasonable to always create a new type, except the specific option is set.

emilio:
ABI of structs may be different, so not 100% sure this is fixable. Though I guess these are never passed by value. Not sure if the presence of a struct in C can change the ABI in any architecture.

C has anonymous structs which we could use too, I think.

porky11:
Anonymous structs seem like a good idea. They could also be used in other cases, where there are multiple fields.
In order to avoid naming conflicts for anonymous structs, you would just prefix the anonymous struct field names with enum field names. But the special case, where an enum variant only has a single unnamed field, the `_0` suffix will be dropped.

Some example:

```rust
enum Test {
    Variant1(i32),
    Variant2(i32, i32),
    Variant3 { x: i32 },
    Variant4 { x: i32, y: i32 },
}
```

This will turn into following C types:

```C
typedef enum {
    Variant1,
    Variant2,
    Variant3,
    Variant4,
} Test_Tag;

typedef struct {
    Test_Tag tag;
    union {
        struct {
            int variant1;
        };
        struct {
            int variant2_0;
            int variant2_1;
        };
        struct {
            int variant3_x;
        };
        struct {
            int variant4_x;
            int variant4_y;
        };
    };
} Test;
```

Anonymous structs seem like a good idea. They could also be used in other cases, where there are multiple fields.
In order to avoid naming conflicts for anonymous structs, you would just prefix the anonymous struct field names with enum field names. But the special case, where an enum variant only has a single unnamed field, the `_0` suffix will be dropped.

Some example:

```rust
enum Test {
    Variant1(i32),
    Variant2(i32, i32),
    Variant3 { x: i32 },
    Variant4 { x: i32, y: i32 },
}
[...]

PR #627, opened 2020-11-29
Refactor `Enum::write`

petrochenkov:
Break up enormous `Enum::write` into several smaller functions, add comments, try to avoid some naming ambiguities.

(Also `must_use` annotations for enums are now added in C mode and not only in C++ code, because they both use common `fn open_struct_or_union` logic now.)

(It's better to review this with whitespace changes ignored.)

This is a preparation for fixing https://github.com/eqrion/cbindgen/issues/353.

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -77,7 +77,7 @@ pub struct EnumVariant {
 
 impl EnumVariant {
     fn load(
-        is_tagged: bool,
+        inline_tag_field: bool,
         variant: &syn::Variant,
         generic_params: GenericParams,
         mod_cfg: Option<&Cfg>,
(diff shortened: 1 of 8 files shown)


src/bindgen/ir/enumeration.rs
```diff
@@ -675,26 +738,31 @@ impl Source for Enum {
...
         if config.language == Language::C && size.is_none() && config.style.generate_typedef() {
             out.close_brace(false);
-            write!(out, " {};", enum_name);
+            write!(out, " {};", tag_name);
         } else {
             out.close_brace(true);
         }
 
+        // Emit typedef specifying the tag enum's size if necessary.
+        // In C++ enums can "inherit" from numeric types (`enum E: uint8_t { ... }`),
```
emilio, line 764: I think "inherit" here is a bit weird, I'd just say that they allow specifying the underlying type. But no big deal either way.

src/bindgen/ir/enumeration.rs
```diff
@@ -586,51 +596,104 @@ impl Item for Enum {
...
+        // If the enum has data, we need to emit a struct or union for the data
+        // and enum for the tag. C++ supports nested type definitions, so we open
+        // the struct or union here and define the tag enum inside it (*).
+        if has_data && config.language == Language::Cxx {
+            self.open_struct_or_union(config, out, inline_tag_field);
+        }
 
-        let is_tagged = self.tag.is_some();
-        let separate_tag = self.repr.style == ReprStyle::C;
+        // Enum the tag enum and everything related to it.
```
emilio, line 616: I think you probably wanted to say `Emit the tag enum [...]` here?

emilio (approved):
This looks great, thank you! I think there's a likely typo and will fix that before merging.

PR #631, opened 2020-12-05
enum: Support inlined definitions for tuple variants with a single field

petrochenkov:
Define nameless structs for enum variants with data inline, without creating separate named `Foo_Body` structs, like described in https://github.com/eqrion/cbindgen/issues/353.
Such inlining is only done for tuple variants with a single field, in other cases it's not an improvement.

#### Unresolved questions:

- The inlining is currently done automatically and cannot be disabled. Do we need control over it? At which granularity exactly? Global config? Annotations for specific variants?

#### Caveats:

- `cbindgen` doesn't currently detect field name conflicts when adding a field named `tag` to structs or unions generated for enums with data. This PR continues that tradition and doesn't detect conflicts when adding fields named `variant_tag` for inlined tags.
- The inlining is not currently done when `derive_tagged_enum_destructor = true` or `derive_tagged_enum_copy_constructor = true` for the given enum. It should be possible to derive destructors and copy constructors when variant definitions are inlined, but it's annoying, e.g. array fields require special cases. I'll probably try to address this limitation next week.

Closes https://github.com/eqrion/cbindgen/issues/353

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -630,7 +630,7 @@ impl Source for Enum {
         }
 
         // Emit the tag enum and everything related to it.
-        self.write_tag_enum(config, out, size, has_data, inline_tag_field, tag_name);
+        self.write_tag_enum(config, out, size, has_data, tag_name);
 
         // If the enum has data, we need to emit structs for the variants and gather them together.
         if has_data {
@@ -669,7 +669,7 @@ impl Source for Enum {
             }
 
             // Emit convenience methods for the struct or enum for the data.
-            self.write_derived_functions_data(config, out, inline_tag_field, tag_name);
+            self.write_derived_functions_data(config, out, tag_name);
 
             // Emit the post_body section, if relevant.
             if let Some(body) = config.export.post_body(&self.path) {
@@ -700,7 +700,6 @@ impl Enum {
         out: &mut SourceWriter<F>,
         size: Option<&str>,
         has_data: bool,
-        inline_tag_field: bool,
         tag_name: &str,
     ) {
         // Open the tag enum.
@@ -797,7 +796,7 @@ impl Enum {
         }
 
         // Emit convenience methods for the tag enum.
-        self.write_derived_functions_enum(config, out, has_data, inline_tag_field, tag_name);
+        self.write_derived_functions_enum(config, out, has_data, tag_name);
     }
 
     /// The code here mirrors the beginning of `Struct::write` and `Union::write`.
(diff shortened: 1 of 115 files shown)


emilio:
So... it seems GCC doesn't allow to put stuff with non-trivial constructors / destructors in these unnamed structs (even when they're defined in an union, which is really odd).

I tried to implement copy-ctors and destructors in 3cd145e65aaba070edd498c8505f40b202329dd5. It seems clang works just fine, but GCC is choking because of that restriction...

I'd love to make this the default as it is IMO a clearly-better default, but that restriction might be annoying.

petrochenkov:
>So... it seems GCC doesn't allow to put stuff with non-trivial constructors / destructors in these unnamed structs (even when they're defined in an union, which is really odd).

Yeah, I've seen that restriction firing on the `destructor_and_copy_ctor` test.
This is a good motivation for a per-variant opt-out at least (in `destructor_and_copy_ctor` the destructor is added manually through config, so cbindgen wouldn't be able to detect it even if it tried).

emilio:
Yeah, a potential alternative is not doing this for C++ mode, since in that case you can generate the casts anyhow, which is nicer?

emilio:
I thought we could work around it by using an anonymous union wrapping the field, but it seems that restriction also applies to unions for some reason? :(

emilio:
(34fd6ee is the above btw)

emilio:
I think I'd prefer to not do this for C++ mode at all, which would allow to always do it on C.

This is because with C++ you can derive the casts anyhow, and that both already does the right thing for single-field variants, and has the advantage that you can add assertions on them.

petrochenkov:
@emilio 
Do you mean entirely removing support for this from C++ mode (i.e. from derived methods), or just disabling it by default and making possible to enable it with an option?

emilio:
I meant keeping C++ codegen as-is (with the inlined casts), and making C use anonymous aggregates.

petrochenkov:
Updated.

src/bindgen/ir/enumeration.rs
```diff
@@ -27,6 +27,14 @@ pub enum VariantBody {
         name: String,
         /// The struct with all the items.
         body: Struct,
+        /// A separate named struct is not created for this variant,
+        /// an unnamed struct is inlined at the point of use instead.
+        /// This is a reasonable thing to do only for tuple variants with a single field.
+        inline: bool,
+        /// Generated cast methods return the variant's only field instead of the variant itself.
+        /// For backward compatibility casts are inlined in a slightly
```
emilio, line 35: This comment might want to be tweaked.

src/bindgen/ir/enumeration.rs
```diff
@@ -27,6 +27,14 @@ pub enum VariantBody {
         name: String,
         /// The struct with all the items.
         body: Struct,
+        /// A separate named struct is not created for this variant,
```
emilio, line 30: Line wrapping is a bit odd here, but fine :)

emilio (approved):
Looks great, thank you!

PR #633, opened 2020-12-06
Partially support `#[cfg]`s on fields

petrochenkov:
Rebase of https://github.com/eqrion/cbindgen/pull/316.

#### Caveats:
- The ugly (but apparently harmless) stray semicolon
    ```c++
    #if !defined(FOO)
      uint8_t field
    #endif
      ;
    ```
- Deriving C++ methods on structs with conditional fields is not supported.

Both issues should be resolved by supporting configuring individual list elements in `write_vertical_source_list`, but that's a separate task.

### DIFF
--- a/src/bindgen/ir/field.rs
+++ b/src/bindgen/ir/field.rs
@@ -2,13 +2,15 @@ use std::io::Write;
 
 use crate::bindgen::cdecl;
 use crate::bindgen::config::{Config, Language};
-use crate::bindgen::ir::{AnnotationSet, Documentation, Path, Type};
+use crate::bindgen::ir::{AnnotationSet, Cfg, ConditionWrite};
+use crate::bindgen::ir::{Documentation, Path, ToCondition, Type};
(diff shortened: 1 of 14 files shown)


src/bindgen/ir/field.rs
```diff
@@ -44,6 +48,12 @@ impl Field {
 
 impl Source for Field {
     fn write<F: Write>(&self, config: &Config, out: &mut SourceWriter<F>) {
+        // Cython doesn't support conditional enum variants.
```
emilio, line 51: conditional fields, right?

tests/rust/cfg.rs
```diff
@@ -40,6 +40,15 @@ struct BarHandle {
     y: f32,
 }
 
+// FIXME: Support deriving methods for structs with conditional fields.
```
emilio, line 43: Could you make sure there's an issue on file to track this?

emilio (approved):
Looks good, just a comment typo :)

PR #635, opened 2020-12-08
Add support for const generics in structs

vmx:
It's now possible to have a const generic in a struct definition.

Example:

    pub struct StructWithConstGeneric<const N: u8>;

It only works for a very limited case. When the struct is used, you
need to supply a defined constant, you cannot use a value directly.

This works:

    const SOME_NUMBER: u8 = 123;
    #[no_mangle]
    pub unsafe extern "C" fn hello_generic(_hello_generic: *const HelloConstGeneric<SOME_NUMBER>) {}

This one doesn't work:

    #[no_mangle]
    pub unsafe extern "C" fn hello_generic(_hello_generic: *const HelloConstGeneric<123>) {}

Even with this limitation it isn't fully working yet. The C++ code doesn't compile. Can someone please tell me what the expected output for C++ should look like?

### DIFF
--- a/src/bindgen/ir/generic_path.rs
+++ b/src/bindgen/ir/generic_path.rs
@@ -17,7 +17,8 @@ impl GenericParams {
                 .params
                 .iter()
                 .filter_map(|x| match *x {
-                    syn::GenericParam::Type(syn::TypeParam { ref ident, .. }) => {
+                    syn::GenericParam::Type(syn::TypeParam { ref ident, .. })
+                    | syn::GenericParam::Const(syn::ConstParam { ref ident, .. }) => {
                         Some(Path::new(ident.to_string()))
                     }
                     _ => None,
--- a/tests/expectations/const_generics.both.c
+++ b/tests/expectations/const_generics.both.c
@@ -1,0 +1,10 @@
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#define SOME_NUMBER 20
+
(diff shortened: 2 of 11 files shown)


emilio:
The C++ should look something like:

```
static const uint8_t SOME_NUMBER = 20;                                             
                                                                                   
template<uint8_t>                                                                
struct StructWithConstGeneric;                                                                                                                                                                                                                
                                                                                                                                                                                                                                              
extern "C" {                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                              
void root(const StructWithConstGeneric<SOME_NUMBER> *a);

}
```

So we need to keep the type distinction in the type argument.

gjtorikian:
I was wondering, if that brief change was made would this be considered mergeable @emilio ? I can cherry pick the commits out and shepherd it in a new PR if this one has stalled.

emilio:
This indeed seems stalled, let's close it for now. A new PR would be welcome, for sure. Thanks!

PiDelport:
I created this issue to track it, in case anyone else wants to pick this up again: #687

vhakulinen:
I had a stab at this, and based on my limited experience with cpp it looks like that this is a lot bigger change than just taking a note of the parameter type in the match statement that this PR changes. The current `GenericParams` looks to me made only work with a single token (e.g. a const variable name) and not with a type (which you need to map to c/cpp types and have different output compared to "just" a variable).

Or then I'm missing something after my 30mins with the code. Am I missing something?

PR #639, opened 2020-12-29
Format release notes in a better way

Luni-4:
You can look how notes are going to be rendered [here](https://github.com/Luni-4/cbindgen/releases/tag/v0.0.1)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -29,16 +29,19 @@ jobs:
       run: |
         strip target/release/cbindgen
 
-    - name: Get cbindgen's version
+    - name: Handle release data and files
       id: tagName
       run: |
         VERSION=$(cargo pkgid | cut -d# -f2 | cut -d: -f2)
         echo "::set-output name=version::$VERSION"
+        tail -n +3 CHANGES | sed '/^##/q' |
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -29,16 +29,19 @@ jobs:
       run: |
         strip target/release/cbindgen
 
-    - name: Get cbindgen's version
+    - name: Handle release data and files
       id: tagName
       run: |
         VERSION=$(cargo pkgid | cut -d# -f2 | cut -d: -f2)
         echo "::set-output name=version::$VERSION"
+        tail -n +3 CHANGES | sed '/^##/q' |
```
emilio, line 37: So this is removing the last header, then stopping at the next, and `sed '$ d'` removes the last line IIUC. Why is `awk` needed, just dedenting? Would be nice to have a comment here.

emilio (approved):
The end result looks great. Mind submitting a follow-up with a tiny comment or confirming that I got that right? Thanks.

.github/workflows/deploy.yml
Luni-4: Yep, correct. I'm going to add a comment in another PR

PR #640, opened 2021-01-04
should_parse_dependency &str arg

arnaudgolfouse:
This applies the 'TODO' at line 99 in `src/bindgen/parser.rs`

The `Parser::should_parse_dependency` method should now be able to take a `&str` argument.
This function is only called in one place (line 165 in `src/bindgen/parser.rs`), so it should not introduce an unexpected issue (and all tests pass anyway :slightly_smiling_face:)


emilio (approved):
Thanks!

Issue #644, opened 2021-01-08
Add parse extra bindings to builder


Kaiser1989:
**Parse -> extra_bindings** is available under _cbindgen.toml_, but missing in the _Builder_.

I added a pull request:

```rust
#[allow(unused)]
pub fn with_parse_extra_bindings<S: AsRef<str>>(mut self, extra_bindings: &[S]) -> Builder {
    self.config.parse.extra_bindings = extra_bindings
        .iter()
        .map(|x| String::from(x.as_ref()))
        .collect();
    self
}
```

emilio:
Fixed by https://github.com/eqrion/cbindgen/pull/645. Thanks!

PR #645, opened 2021-01-08
[#644] Add parse extra bindings to builder

Kaiser1989:
Added missing function to builder:

see issue: Add parse extra bindings to builder #644


emilio (approved):
Looks good, thanks!

PR #652, opened 2021-01-29
feat: Add the `none` documentation style

Hywan:
This new documentation style skips the generation of the documentation.

I'm not sure how to test it. Any clue?

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -484,6 +484,7 @@ tab_width = 3
 # How the generated documentation should be commented.
 #
 # possible values:
+# * "none": documentation is not generated
 # * "c": /* like this */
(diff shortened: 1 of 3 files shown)


emilio:
So we have an existing (and apparently undocumented, lol) `documentation = <bool>` field. This was introduced in 8793c227.

Would that cover your use case? If not, why not? This should be easily testable. Add a test file with docs to `tests/rust/`, and an accompanying toml file specifying your configuration option.

Hywan:
It does cover my needs. Sorry I missed that feature. Thanks!

emilio:
Great! We should make sure to document this.

PR #669, opened 2021-03-19
Fix clippy::inconsistent_struct_constructor

elmarco:
A new clippy warning.


emilio:
Sweet, thanks! And sorry for the lag here.

PR #671, opened 2021-03-22
use --profile=check for macro expansion

### LINKED ISSUE #457: --pretty=expanded fails after calling cbindgen in cdylib crate
# Preface

This one is a bit of a doozy and I can't tell if it's a rustc issue or a cbindgen issue or both, so I'm going to cross-report them so both projects can be aware.

I don't have a minimal reproduction, but it is easily reproducible in my repo.

Repo: https://github.com/BVE-Reborn/bve-reborn
Working Commit: [`52b3ec6`](https://github.com/BVE-Reborn/bve-reborn/commit/52b3ec6c232198e7585909f20404b9ee5323bdab)
Failing Commit: [`7799c04`](https://github.com/BVE-Reborn/bve-reborn/commit/7799c04340edefcd52c51023e1271c0bb67d5ad1)

The only difference is the addition of the crate type of `cdylib`. (`rlib` is there, but it reproduces on my next commit which removed it)

# System Info

`Xubuntu 18.04`

Rust and Cargo locked to `nightly-2020-01-16`
rustc: `rustc 1.42.0-nightly (3291ae339 2020-01-15)`
cargo: `cargo 1.42.0-nightly (ad3dbe10e 2020-01-13)`

CBindgen latest, compiled on the same nightly toolchain:
`cbindgen 0.12.2`

# Repro

The rustc command is the exact command cbindgen runs to get macro expanded data.

This all takes place in the root of my repo.

```
cargo build
cargo rustc --lib --manifest-path bve-native/Cargo.toml  -p "bve-native:0.0.0-Placeholder" "--verbose" -- -Z unstable-options --pretty=expanded 
```

Both succeed as expected. However if you call cbindgen it fails, and the rustc no longer outputs anything at all.

```
# Succeeds
cargo build

# Fails citing an issue with compiling
# Occasionally must be run twice
cbindgen bve-native/ -c bve-native/cbindgen.toml -v

# Returns zero, but outputs nothing.
cargo rustc --lib --manifest-path bve-native/Cargo.toml  -p "bve-native:0.0.0-Placeholder" "--verbose" -- -Z unstable-options --pretty=expanded 
# You can call ^ as many times as you want, it will always fail
```

The problem goes away when you `cargo clean` and reappears by running cbindgen again.

# Comments

I have no idea what or who is causing this issue.

I feel like even if cbindgen is ultimately the "cause" of the bug, the expansion shouldn't be able to fail and not output anything without erroring (and returning `!= 0`) in some way.

# Notes:

Possibly relevant cbindgen code: https://github.com/eqrion/cbindgen/blob/master/src/bindgen/cargo/cargo_expand.rs#L129
rustc issue: https://github.com/rust-lang/rust/issues/68333

Thanks!

cormacrelf:
fixes #457

### DIFF
--- a/src/bindgen/cargo/cargo_expand.rs
+++ b/src/bindgen/cargo/cargo_expand.rs
@@ -92,6 +92,7 @@ pub fn expand(
 
     cmd.arg("rustc");
     cmd.arg("--lib");
+    cmd.arg("--profile=check");
     cmd.arg("--manifest-path");
     cmd.arg(manifest_path);
     if let Some(features) = expand_features {


emilio (approved):
Looks sensible to me, thanks!

PR #672, opened 2021-03-23
enum: take renaming into account for variant prefix

elmarco:


emilio (approved):
Thanks, this looks sensible!

PR #676, opened 2021-04-06
Only fetch dependencies for current platform by default

### LINKED ISSUE #675: Only use metadata for current target
Unlike most other cargo tools, `cargo metadata` counter-intuitively defaults to downloading all dependencies for _all_ targets unless the argument `--filter-platform` is passed. This means that with the current logic, `cbindgen` ends up downloading potentially a very large number of dependencies whenever this is invoked:
https://github.com/eqrion/cbindgen/blob/93eed3aaf3a6ed94ae5bdf59e5df4c508fe05a5b/src/bindgen/cargo/cargo_metadata.rs#L189-L194

Since `cbindgen` doesn't actually need the dependencies themselves (I don't think), I believe `cbindgen` should probably _always_ specify `--filter-platform`, and just specify the host platform. Cargo finds the host platform by running `rustc -Vv` and extracing the `host:` line, which `cbindgen` could probably also do pretty easily:
https://github.com/rust-lang/cargo/blob/fde8ee3ba0cba4eef75f197d42586ef4664f057b/src/cargo/util/rustc.rs#L48-L68

This is also what `rust-analyzer` recently switch to doing: https://github.com/rust-analyzer/rust-analyzer/pull/6912

There's also precedent for relying on `rustc` to determine these kinds of things, such as using it to find the sysroot:
https://github.com/rust-analyzer/rust-analyzer/blob/48802e54d1669c68fd27b007533dd6aa5221d089/crates/project_model/src/sysroot.rs#L126-L127

One way to make this nicer would be if `cargo metadata` supported a "special" value for `--filter-platform` called `host`. I'll look at submitting that as a PR upstream, but the more immediate fix (and also one that would continue to work for older compiler releases) would be to just use `rustc -Vv`.

jonhoo:
cbindgen invokes `cargo metadata` to find the sources of dependencies so
that it can generate bindings for types that originate in those
dependencies. However, `cargo metadata` defaults to fetching
dependencies for _all_ platforms unless it is specifically invoked with
the `--filter-platforms` argument.

This PR makes cbindgen pass that argument to `cargo metadata` using the
host platform, which will significantly reduce the time of the first
call in cases where the dependency tree includes many target-specific
dependencies.

Fixes #675.

### DIFF
--- a/src/bindgen/cargo/cargo_metadata.rs
+++ b/src/bindgen/cargo/cargo_metadata.rs
@@ -195,7 +195,8 @@ pub fn metadata(
                 // If `rustc` fails to run, we just fall back to not passing --filter-platforms.
                 //
                 // NOTE: We set the current directory in case of rustup shenanigans.
-                let rustc = Command::new("rustc")
+                let rustc = env::var("RUSTC").unwrap_or_else(|_| String::from("rustc"));
+                let rustc = Command::new(rustc)
                     .current_dir(manifest_path.parent().unwrap())
                     .arg("-vV")
                     .output();
--- a/src/bindgen/cargo/cargo_metadata.rs
+++ b/src/bindgen/cargo/cargo_metadata.rs
@@ -204,7 +204,14 @@ pub fn metadata(
                 if let Ok(out) = rustc {
                     if let Ok(stdout) = std::str::from_utf8(&out.stdout) {
                         let field = "host: ";
-                        let value = stdout.lines().find_map(|l| l.strip_prefix(field));
+                        let value = stdout.lines().find_map(|l| {
+                            // XXX l.strip_prefix(field) re-implemented to preserve MSRV
+                            if l.starts_with(field) {
+                                Some(&l[field.len()..])
+                            } else {
+                                None
+                            }
+                        });
                         if let Some(value) = value {
                             target = Some(value.to_string());
                         }
--- a/src/bindgen/cargo/cargo_metadata.rs
+++ b/src/bindgen/cargo/cargo_metadata.rs
@@ -180,7 +180,7 @@ impl Eq for Dependency {}
 pub fn metadata(
     manifest_path: &Path,
     existing_metadata_file: Option<&Path>,
-    only_host: bool,
+    only_target: bool,
 ) -> Result<Metadata, Error> {
     let output;
     let metadata = match existing_metadata_file {
@@ -188,8 +188,13 @@ pub fn metadata(
         None => {
             let mut target = None;
 
-            if only_host {
-                // cargo metadata defaults to giving information for _all_ targets.
+            if only_target {
+                target = std::env::var_os("TARGET");
+            }
+
+            if only_target && target.is_none() {
+                // We must be running as a standalone script, not under cargo.
+                // Let's use the host platform instead.
                 // We figure out the host platform through rustc and use that.
                 // We unfortunatelly cannot go through cargo, since cargo rustc _also_ builds.
                 // If `rustc` fails to run, we just fall back to not passing --filter-platforms.
@@ -213,7 +218,7 @@ pub fn metadata(
                             }
                         });
                         if let Some(value) = value {
-                            target = Some(value.to_string());
+                            target = Some(value.into());
                         }
                     }
                 }
--- a/src/bindgen/cargo/cargo_metadata.rs
+++ b/src/bindgen/cargo/cargo_metadata.rs
@@ -180,37 +180,41 @@ impl Eq for Dependency {}
 pub fn metadata(
     manifest_path: &Path,
     existing_metadata_file: Option<&Path>,
+    all_platforms: bool,
 ) -> Result<Metadata, Error> {
     let output;
     let metadata = match existing_metadata_file {
         Some(path) => Cow::Owned(std::fs::read_to_string(path)?),
         None => {
-            // cargo metadata defaults to giving information for _all_ targets.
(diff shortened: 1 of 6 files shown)


jonhoo:
The CI failure is due to a new clippy lint (`clippy::wrong-self-convention`) — PR incoming for that.

jonhoo:
Fix in https://github.com/eqrion/cbindgen/pull/677.

src/bindgen/cargo/cargo_metadata.rs
```diff
@@ -180,16 +180,51 @@ impl Eq for Dependency {}
...
+            let mut target = None;
+
+            if !all_platforms {
+                // cargo metadata defaults to giving information for _all_ targets.
+                // We figure out the host platform through rustc and use that.
+                // We unfortunatelly cannot go through cargo, since cargo rustc _also_ builds.
+                // If `rustc` fails to run, we just fall back to not passing --filter-platforms.
+                //
+                // NOTE: We set the current directory in case of rustup shenanigans.
+                let rustc = Command::new("rustc")
```
emilio, line 198: This should probably use `env::var("RUSTC").unwrap_or_else...` like `CARGO` below.

src/bindgen/cargo/cargo_metadata.rs
```diff
@@ -180,16 +180,51 @@ impl Eq for Dependency {}
...
+                    .current_dir(manifest_path.parent().unwrap())
+                    .arg("-vV")
+                    .output();
+                log::debug!("Discovering host platform by {:?}", rustc);
+                if let Ok(out) = rustc {
+                    if let Ok(stdout) = std::str::from_utf8(&out.stdout) {
+                        let field = "host: ";
```
emilio, line 205: Hmm, this also seems pretty odd of a default... Thinking a bit more about this, shouldn't this default to fetch the _target_ dependencies?

emilio (commented):
So I'm unsure about which one should be the default tbh. It seems reasonable to add cfg conversions and them just work... Plus I'm not in love that this breaks for pretty much all cross-compilation use cases by default...k

That plus the fact that _not_ fetching all dependencies is technically a breaking change makes me err more on the side of inverting the flag.

Actually, does cargo read `CARGOFLAGS` internally? If so, this is the same as `CARGOFLAGS="--filter-platform <target>" cbindgen, isn't it? If it isn't, I think a more generic solution is just to allow passing flags to cargo, one way or another... wdyt?

jonhoo:
Yeah, both options aren't great here. If the default is to not fetch all dependencies, then cross-compilation breaks (if types from target-specific dependencies are used). If the default is to fetch all dependencies, then builds that do not use target-specific dependencies are slowed down due to the additional downloads and scanning. I think I agree with you that it's better to not break by default than to be slow by default though, so swapped the default in https://github.com/eqrion/cbindgen/pull/676/commits/02d8bfd82c1fe79d4f0deb61fcb104160dca935b.

No, cargo [doesn't read `CARGOFLAGS`](https://github.com/rust-lang/cargo/search?q=CARGOFLAGS&type=issues). I don't think that's what we'd want anyway, since this is specifically _just_ a flag to `cargo metadata` — all other cargo commands default to only operating on the current platform. So, for example, if you call `cargo rustc`, it _won't_ fetch dependencies for other targets. It's really just `cargo metadata` that has an inverted default compared to the other tools. So for example when we call `cargo rustc` here, that is _only_ building for the current target platform:
https://github.com/eqrion/cbindgen/blob/54dfcf89e419afbf60f123088462ef91edd85916/src/bindgen/cargo/cargo_expand.rs#L93

src/bindgen/cargo/cargo_metadata.rs
jonhoo: I'm not sure I follow? From what I can tell, cbindgen is never passed a target platform, so the only reasonable thing to use here is the host platform (which is what this code does). Or rather, it uses the `host` field of `rustc -vV` for whatever `rustc` cbindgen is told to use.

emilio: Right, my point is that cbindgen doesn't have target information, but using the host also isn't the right thing if you run cbindgen as part of the build.

jonhoo: Ah, you're imagining that someone runs, say, `cargo build --target xyz`, and then `build.rs` invokes `cbindgen`, and it ends up using the host target even though `xyz` was explicitly passed in?

It looks like we can use some of the environment variables [cargo sets for build scripts](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts) here. Specifically, the `CARGO_CFG_` variables. I'll give that a stab!

jonhoo: Ah, or better yet, it also just sets `TARGET`.

jonhoo: Take a look at https://github.com/eqrion/cbindgen/pull/676/commits/b5abdf6fe048116cf66f260e2c8627dfd0ef8b42 — I think that does what you want!

jonhoo:
Gentle nudge on this @emilio — I think it now works the way you wanted it to?

emilio (approved):
Yes, this looks sensible, thanks!

jonhoo:
Happy to help! For what it's worth, I _do_ think flipping the default on the next breaking change might be worthwhile — it's probably better to not automatically fetch more dependencies than usual by default, and then have crates that _need_ cross-compilation opt into it?

PR #677, opened 2021-04-06
Fix new clippy lint clippy::wrong-self-convention

jonhoo:
This should be backwards-compatible in practice since `IdentifierType`
is `Copy`. In theory it's not though, as someone _might_ be passing that
function by name to something that requires that exact signature.


emilio (approved):
Thanks!

PR #688, opened 2021-06-07
add `Builder::with_using_namespaces`

dicej:


emilio (approved):
Sorry, had missed this one, looks good!

emilio:
There are some CI failures but they're unrelated to your patch (clippy version update looks like). So will merge as-is and address the clippy things separately.

PR #695, opened 2021-06-19
Ignore PhantomPinned

kkoning:
cbindgen already ignores `PhantomData`.  It should ignore `PhantomPinned` for the same reasons.


emilio:
Thanks! rustfmt was complaining so I merged your change manually with the formatting fix.

PR #697, opened 2021-06-19
Simplify Pin<T> to T

kkoning:
cbindgen already simplifies `MaybeUninit<T>` and `ManuallyDrop<T>` to `T`. This adds `Pin<T>` as well.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -161,6 +161,8 @@ cbindgen contains the following hardcoded mappings (again completely ignoring na
 * PhantomData => *evaporates*, can only appear as the field of a type
 * PhantomPinned => *evaporates*, can only appear as the field of a type  
 * () => *evaporates*, can only appear as the field of a type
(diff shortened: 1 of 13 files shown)


kkoning:
This is another one-liner, but this time treating `Pin<T>` the same way as `MaybeUninit<T>` and `ManuallyDrop<T>`, and with a new test.

Given this source:

```rust
#[repr(C)]
struct PinTest {
    pinned_box: Pin<Box<i32>>,
    pinned_ref: Pin<&mut i32>
}

#[no_mangle]
pub extern "C" fn root(s: Pin<&mut i32>, p: PinTest) {}
```

the binding created is:

```c
typedef struct {
  int32_t *pinned_box;
  int32_t *pinned_ref;
} PinTest;

void root(int32_t *s, PinTest p);
```

emilio (approved):
Lgtm, thanks! The clippy lint failure is not your fault and should be fixed already in trunk.

Issue #569, opened 2020-09-01
Crashes when built using -Cpanic=abort


h33p:
proc-macro2 crashes during parsing when building with `-Cpanic=abort` flag. `cbindgen-0.14.4/src/bindgen/parser.rs:229` leads all the way to `proc-macro2-1.0.19/src/detection.rs:59` where it panics and aborts. This is a proc-macro2 issue in question: https://github.com/alexcrichton/proc-macro2/issues/218

Solution would be to incorporate `proc_macro2::fallback::force()` somewhere in the code.

jschwe:
Can this issue be closed? The upstream issues was closed end of 2021.

PR #698, opened 2021-06-20
parser: force proc macro fallback

elbe0046:
Incorporate https://github.com/alexcrichton/proc-macro2/pull/220 to fix the case where a binary which is built with `panic = "abort"` aborts when calling into cbindgen functionality.

Addresses https://github.com/eqrion/cbindgen/issues/569.

While there is a test included that covers both additions I haven't made any changes to documentation.  If there is desire to have these changes land in some form I'd just be looking for guidance on what doc changes would be appropriate, if any, in addition to any feedback on the overall direction of this change.  Thanks!

### DIFF
--- a/src/bindgen/parser.rs
+++ b/src/bindgen/parser.rs
@@ -206,6 +206,7 @@ impl<'a> Parser<'a> {
                         self.config.parse.expand.profile,
                     )
                     .map_err(|x| Error::CargoExpand(pkg.name.clone(), x))?;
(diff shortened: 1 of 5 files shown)


emilio:
Does this have any performance implications? Can the `panic=abort` binary call this instead?

elbe0046:
> Can the panic=abort binary call this instead?

Of course, good thinking - thanks for the quick turnaround too!

PR #699, opened 2021-06-21
Follow multiple transparent links in Option layout

### LINKED ISSUE #326: Using an `Option<>` with a function type alias generates incorrect bindings
With the following code:

```rust
use libc::int32_t;
type DoFn = unsafe extern "C" fn (x: int32_t, y: int32_t) -> int32_t;

#[repr(C)]
#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
struct Holder {
    func: Option<DoFn>,
}
```

bindgen will generate the following incorrect header:

```c
#include <stdarg.h>
// ..
struct Option_DoFn;

struct Holder {
    Option_DoFn func;
};
```

instead of the expected

```c
#include <stdarg.h>
// ..
typedef int32_t(*DoFn)(int32_t x, int32_t y);

struct Holder {
    DoFn func;
};
```

### LINKED ISSUE #690: #[repr(transparent)] struct with a NonZero field
As of #646 `Option<NonZeroU32>` gets translated to `uint32_t` correctly.  If you have a `#[repr(transparent)]` struct containing a `NonZeroU32` field, that gets translated to `uint32_t` as well.  But if you have an `Option` of that struct, the transparency doesn't compose:

``` rust
use std::num::NonZeroU32;

#[no_mangle]
pub extern "C" fn unwrapped(value: NonZeroU32, opt: Option<NonZeroU32>) {}

#[repr(transparent)]
pub struct Wrapped {
    value: NonZeroU32,
}

#[no_mangle]
pub extern "C" fn wrapped(value: Wrapped, opt: Option<Wrapped>) {}
```

results in:

``` c
#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>

struct Option_Wrapped;

typedef uint32_t Wrapped;

void unwrapped(uint32_t value, uint32_t opt);

void wrapped(Wrapped value, struct Option_Wrapped opt);
```

Ideally the `opt` parameter of `wrapped` could have `Wrapped` as its type.  Would that be possible?

(This is with cbindgen 0.19.0)

dcreager:
Option can take advantage of layout optimizations when its inner type is nullable or zeroable.  Before, we were only looking at the Option's direct generic parameter when deciding whether we could use this optimization.  But the layout optimization "propagates through" typedefs and transparent structs.  We now repeatedly follow these type aliases until we reach a non-aliased type, and only then check whether the type is nullable/zeroable.

Fixes #690
Fixes #326


tests/rust/nested_nonzero.rs
```diff
@@ -0,0 +1,16 @@
...
+    value: std::num::NonZeroU32,
+    _phantom: std::marker::PhantomData<T>,
+}
+
+struct File;
+
+#[repr(C)]
+struct Node {
+    file: Handle<File>,
+    maybe_file: Option<Handle<File>>,
```
dcreager, line 12: One thing I don't like about this patch is that you end up with different types for these two fields.

tests/expectations/nested_nonzero.tag.c
```diff
@@ -0,0 +1,13 @@
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+typedef uint32_t Handle_File;
+
+struct Node {
+  Handle_File file;
+  uint32_t maybe_file;
```
dcreager, line 10: like here.  Ideally both fields would be `Handle_File`.

src/bindgen/transparent_types.rs
```diff
@@ -0,0 +1,49 @@
...
+
+use std::collections::HashMap;
+
+use crate::bindgen::ir::{GenericParams, ItemMap, Path, Struct, Type, Typedef};
+
+/// Keeps track of named types that have the same underlying representation as some other type.
+/// This happens via `#[repr(transparent)]` structs and via typedefs.
+#[derive(Default)]
+pub struct TransparentTypes {
+    transparent: HashMap<Path, (Type, GenericParams)>,
```
emilio, line 13: So it is quite a bit unfortunate to have to go through all items again creating this data structure when we already have a `<Path, Struct>` etc hashmap in `ItemMap`, it just feels like a lot of upfront work for something that could be much more efficient.

tests/expectations/nested_nonzero.tag.c
emilio: Yeah, agreed... In practice it shouldn't be a huge deal for C. For C++ it can be a bit more annoying (I can imagine someone defining their own `Handle<>` for example.

src/bindgen/transparent_types.rs
dcreager: Yeah, I definitely agree.  At first I tried passing `&ItemMap<Struct>` and `&ItemMap<Typedef>` references in to the `simplify_standard_types` methods, but that runs afoul of the borrow checker, since we're calling them from within a `for_all_items_mut` loop that is updating the contents of those `ItemMap`s.  Do you have any suggestions for how to break that double-borrow?

tests/expectations/nested_nonzero.tag.c
dcreager: The latest patch fixes this, at the expense of the `infinite_recursion_typedef_monomorph` test going into an infinite loop.  I'd really love to be able to use the `Monomorphs` data here, but that's not generated until later in the process, and it looks like it depends on all of the types already having been simplified.

jonnius:
Just in case you are looking for a complex example. I tried to run a build from this PR on [libsignal-client](https://github.com/signalapp/libsignal-client/tree/master/rust/bridge/ffi) and ran into a stackoverflow (with 16 GB of RAM). I guess this might be result of (endless) recursion.

PR #701, opened 2021-07-06
Add NS_ENUM support, for Swift

cormacrelf:
See [NS_ENUM docs][appledocs]. See this excerpt from docs.md for motivation:

[appledocs]: https://developer.apple.com/documentation/swift/objective-c_and_c_code_customization/grouping_related_objective-c_constants

> Normally, the ergonomics of C enums in Swift is very poor; you can't use the variants directly and pass them to function accepting the typedef name, because Swift sees the variants each as just raw integers and the typedef name as a newtype. So you essentially have to redefine the enum in Swift, get its `.rawValue` and convert it to the C enum type, and convert back in the other direction.
>
> With `NS_ENUM`/`CF_ENUM`, you can import e.g. `#[repr(i32)] pub enum A { Abc,
Def }` in Swift and use it as `A.abc` as you would with a Swift enum. 

Config is like so:

```toml
sys_includes = ["CoreFoundation/CoreFoundation.h"]
[enum]
swift_enum_macro = "CF_ENUM"
```

Output is like so:

```c
typedef CF_ENUM(uint64_t, A) {
  a1 = 0,
  a2 = 2,
  a3,
  a4 = 5,
};
```

It clobbers `cpp_compat`'s effect on enums, which is normally to add a C++11 fixed type in an ifdef (`enum Name: uint32_t { ... }`). This is because CF_ENUM/NS_ENUM will already do that; the difference is it works in Objective-C as well, and maybe they do some other annotation magic to get Swift to recognise enums.

### DIFF
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -573,6 +573,14 @@ pub struct EnumConfig {
     /// Whether to generate empty, private default-constructors for tagged
     /// enums.
     pub private_default_tagged_enum_constructor: bool,
+    /// An optional macro to use when generating enum typedefs, which must take two arguments
+    /// along the lines of `NS_ENUM` (`NS_ENUM(NSInteger, MyEnumTypedefName)`).
+    /// See the [Apple documentation][appledoc] for details.
+    ///
+    /// Only relevant when targeting C; also disables the `cpp_compat` effect on enum definitions.
+    ///
+    /// [appledoc]: https://developer.apple.com/documentation/swift/objective-c_and_c_code_customization/grouping_related_objective-c_constants
+    pub swift_enum_macro: Option<String>,
 }
 
 impl Default for EnumConfig {
@@ -592,6 +600,7 @@ impl Default for EnumConfig {
             derive_ostream: false,
             enum_class: true,
             private_default_tagged_enum_constructor: false,
+            swift_enum_macro: None,
         }
     }
 }
@@ -659,6 +668,12 @@ impl EnumConfig {
             return x;
         }
         self.private_default_tagged_enum_constructor
+    }
+    pub(crate) fn swift_enum_macro(&self, annotations: &AnnotationSet) -> Option<String> {
+        if let Some(x) = annotations.atom("swift-enum-macro") {
+            return x;
+        }
+        self.swift_enum_macro.clone()
     }
 }
 
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -301,6 +301,7 @@ impl Source for EnumVariant {
 pub struct Enum {
     pub path: Path,
     pub export_name: String,
+    pub swift_enum_macro: Option<String>,
     pub generic_params: GenericParams,
     pub repr: Repr,
     pub variants: Vec<EnumVariant>,
@@ -319,6 +320,22 @@ impl Enum {
     /// Enum with data turns into a union of structs with each struct having its own tag field.
     fn inline_tag_field(repr: &Repr) -> bool {
         repr.style != ReprStyle::C
+    }
+
+    fn typedef_inline_enum_macro(&self, config: &Config) -> Option<&str> {
+        let is_c = config.language == Language::C;
+        let is_fixed_type = self
+            .repr
+            .ty
+            .map(|ty| ty.to_primitive())
+            // technically you're not going to get #[repr(double)] because that's not valid rust,
+            // but let's check anyway
+            .map_or(false, |prim| prim.can_be_enum_fixed_type());
+        if is_c && is_fixed_type {
+            self.swift_enum_macro.as_ref().map(|x| x.as_str())
+        } else {
+            None
+        }
     }
 
     pub fn add_monomorphs(&self, library: &Library, out: &mut Monomorphs) {
@@ -412,6 +429,8 @@ impl Enum {
             None
         };
 
+        let swift_enum_macro = config.enumeration.swift_enum_macro(&annotations);
+
         Ok(Enum::new(
             path,
             generic_params,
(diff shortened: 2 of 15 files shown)


src/bindgen/ir/ty.rs
```diff
@@ -306,6 +306,24 @@ impl PrimitiveType {
     fn can_cmp_eq(&self) -> bool {
         true
     }
+
+    pub(crate) fn can_be_enum_fixed_type(&self) -> bool {
```
emilio, line 310: I think I'd remove this. We don't check it for other cases, and as you said it's invalid rust so there's little point in checking it.

src/bindgen/ir/enumeration.rs
```diff
@@ -446,6 +467,7 @@ impl Enum {
             cfg,
             annotations,
             documentation,
+            swift_enum_macro,
```
emilio, line 470: Given this can be computed with config+annotations, what's the point of having it in the enum struct? I might be missing something, just skimming over from the phone though

src/bindgen/ir/enumeration.rs
```diff
@@ -321,6 +322,22 @@ impl Enum {
         repr.style != ReprStyle::C
     }
 
+    fn typedef_inline_enum_macro(&self, config: &Config) -> Option<&str> {
+        let is_c = config.language == Language::C;
```
emilio, line 326: Why don't we want to generate this for C++?

tests/expectations/swift_enum.both.c
```diff
@@ -0,0 +1,184 @@
+#import <CoreFoundation/CoreFoundation.h>
```
emilio, line 1: These headers are not available in CI so you need to use a local macro for the test to compile or what not. Let me know if you need help with that.

src/bindgen/ir/enumeration.rs
cormacrelf: Since this is primarily targeted at ObjC and Swift and people using the (Core)Foundation headers I figure people aren't using C++ output for that. I could be wrong, but I don't see it being useful in that case. All that's at stake is making people remove this from the configuration when switching language, I guess.

I added that line originally to avoid code path in Python mode when writing the "ctypedef" line.

tests/expectations/swift_enum.both.c
cormacrelf: Roughly something like this should do it I think. The macro is called after the typedef keyword.

```
#define CF_ENUM(_type, _name) _type _name; enum _name
```

src/bindgen/ir/enumeration.rs
cormacrelf: The only point is to allow customising it with comment annotations. But I suppose it's not like people have other macros lined up ready to use instead of CF_ENUM. It's a pretty bad mount point for the syntax, such that you can't really do anything useful other than add compiler specific annotations. You can't use the macro to add a name after the enum variants, nor to add a typedef after the whole enum declaration and semicolon because enums can't be forward-declared (unlike structs/unions). As you can tell by the pretty bad macro in my other comment.

So yeah, I didn't put much thought into that, I was just copying the other Swift-related config. Will remove.

src/bindgen/ir/enumeration.rs
cormacrelf: The unmentioned context is that

1. technically yes, NS_ENUM will work fine in Apple toolchains in C++ mode, I believe it will produce unscoped enums with fixed types just like cpp_compat currently does; but
2. targeting C++ presently gives you enum classes, which are better in literally every way than NS_ENUM.

Basically we can probably afford to be smarter than our users here.

Final thought is along similar lines: The cpp_compat option clearly considers that the C output is meant to be used with `[enum] prefix_with_name = true`, as is customary for C. Hence the use of unscoped fixed type enums (`enum Name : type` instead of scoped `enum class Name : type`) in the ifdef. Should that option, which currently defaults to false to match C++ scoped enums, default to true instead for C?

cormacrelf: Ah, one more thing, Swift cannot import C++ and all C++ interop currently has to be done via writing or generating C bindings for the C++ code. There is a [document ('manifesto') with some design considerations][manifesto] for a potential Swift/C++ interop, but if Swift ever gets to import C++ directly, it will obviously address fixed-type enums and enum classes. So NS_ENUM for C++ output will never be necessary.

[manifesto]: https://github.com/apple/swift/blob/main/docs/CppInteroperabilityManifesto.md

cormacrelf:
@emilio I think all the issues are resolved, thanks for the reviews, please have another look when you have a chance :)

src/bindgen/ir/enumeration.rs
```diff
@@ -328,9 +328,7 @@ impl Enum {
             .repr
             .ty
             .map(|ty| ty.to_primitive())
-            // technically you're not going to get #[repr(double)] because that's not valid rust,
-            // but let's check anyway
-            .map_or(false, |prim| prim.can_be_enum_fixed_type());
+            .is_some();
```
emilio: I believe this is not what you intended. This is effectively `self.repr.ty.is_some()` so it should be simplified to that probably.

src/bindgen/ir/enumeration.rs
emilio: Ok, sounds good, thanks.

src/bindgen/config.rs
```diff
@@ -669,12 +669,6 @@ impl EnumConfig {
         }
         self.private_default_tagged_enum_constructor
     }
-    pub(crate) fn swift_enum_macro(&self, annotations: &AnnotationSet) -> Option<String> {
```
emilio: To be clear, I wasn't particularly against the annotation (though it probably isn't terribly useful and we can add it back if ever needed). I just thought that storing the "resolved" value shouldn't be necessary, since we store the annotations already.

qmathe:
Hi, 

I'd really interested to have this PR merged. What remains to be done? 

It seems to me, there is a single comment that was not addressed, but that's all: https://github.com/mozilla/cbindgen/pull/701#discussion_r667082507

PR #703, opened 2021-07-08
Rename items with a rule

cormacrelf:
Everything else can have a rename rule. Why not items too! 

Example below. Nice and neat, no `#![allow(non_camel_case_types)]` required.

```rust
#[repr(c)]
struct StructDef {
    x: i32,
    y: f32,
}
type TypeAlias = StructDef;
// doesn't apply to functions, obviously
// also this PR fixes misleading example in docs for [export.rename] which purported to apply to a function name
#[no_mangle]
extern "C" fn capi_write(obj: *mut StructDef) {}
```

```toml
[export]
prefix = "capi_"
rename_items = "snake_case"
```

```c
typedef struct {
    int32_t x;
    float y;
} capi_struct_def;
typedef capi_struct_def capi_type_alias;
void capi_write(capi_struct_def *def);

```

### DIFF
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -368,7 +368,7 @@ impl ExportConfig {
             }
         }
         if let Some(rename_rule) = self.rename_items.not_none() {
-            let cow = rename_rule.apply(&item_name, ident_type);
+            let cow = rename_rule.apply(item_name, ident_type);
             *item_name = cow.into_owned();
         }
         if let Some(ref prefix) = self.prefix {


cormacrelf:
I realised this is a dupe of #419. This one has tests though.

tests/expectations/rename_items.c
```diff
@@ -0,0 +1,39 @@
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#define capi_constant_abc 10
```
cormacrelf, line 6: This is basically the use case for allowing one rule per item type as well as a default global one. I don't often see lower_snake_case #defines in C, even in a largely lower_snake_case API. This PR already has the infrastructure for that, passing `IdentifierType` in to `ExportConfig::rename`, but hasn't implemented it (yet).

emilio (commented):
So this looks probably ok, but wouldn't it be more consistent to just add a `ConstantConfig::rename` field of sorts? I think the point was that some constants might want to be `static` and can't be renamed, so that might be a bit trickier.

orium:
Hi.  Any change of getting this merged/release soon?

pixsperdavid:
I need this feature. Is this PR likely to ever be merged, or would I be better off submitting a fresh PR for the same issue?

cormacrelf:
Hi David, I added you as a collaborator, so feel free to rebase this branch onto master, and push straight back to it.

pixsperdavid:
@cormacrelf Many thanks! Will do so.

PR #706, opened 2021-07-28
Update --pretty=expanded to -Zunpretty=expanded

badboy:
`--pretty` is gone as of https://github.com/rust-lang/rust/pull/83491

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -399,7 +399,7 @@ trailer = "/* Text to put at the end of the generated file */"
 # default: doesn't emit an include guard
 include_guard = "mozilla_wr_bindings_h"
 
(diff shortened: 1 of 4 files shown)


badboy:
> why does this commit remove `-Zunstable-options` ? `-Zunpretty=expanded` is unstable so now you can't use this option in cbindgen on the stable compiler.

This was already true before though. The `-Z` option is only available on nightly. Previously `-Zunstable-options` was required to get `--pretty=expanded`. That one was removed and `-Zunpretty=expanded` doesn't need `unstable-options`.

iitalics:
you're right, my bad

PR #716, opened 2021-09-16
Add documentation_style to with short and full options

seanmonstar:
When the "short" option is used, only the first line of the docs is output for each item.

The motivation for this change is to encourage writing longer docs with examples, while still keeping the header file more concise. The header file top can include a link to the full published docs instead.

### DIFF
--- a/src/bindgen/ir/documentation.rs
+++ b/src/bindgen/ir/documentation.rs
@@ -4,7 +4,7 @@
 
 use std::io::Write;
 
-use crate::bindgen::config::{Config, DocumentationStyle, Language};
+use crate::bindgen::config::{Config, DocumentationLength, DocumentationStyle, Language};
 use crate::bindgen::utilities::SynAttributeHelpers;
 use crate::bindgen::writer::{Source, SourceWriter};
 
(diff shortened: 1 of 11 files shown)


seanmonstar:
@emilio sorry to spring a PR without an issue first... what do you think of the feature?

emilio:
Sorry, it's been pretty weird times for me lately so haven't been able to do much non-work stuff, looking now

src/bindgen/ir/documentation.rs
```diff
@@ -76,7 +76,12 @@ impl Source for Documentation {
             _ => (),
         }
 
-        for line in &self.doc_comment {
+        let end = match config.documentation_length {
```
emilio, line 79: Can we handle it also in the cython case just for completeness? Should be trivial moving this a few lines above but if you don't have the time I'm happy to do it.

emilio (approved):
Looks good to me, thanks.

emilio:
I can fix cython for you before publishing, thanks, and sorry again for the massive lag.

emilio:
a1694cd57c604280efbe7a3ec0c3cb226680ba91 fixes cython

PR #720, opened 2021-10-31
Add `type_replacements` option to control output of name mangling

YuhanLiin:
This PR introduces the `mangle.type_replacements` option, allowing users to replace type parameter names with their own strings when name mangling. This enables users to customize and control the output of the name mangling process.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -609,6 +609,14 @@ rename_types = "PascalCase"
 # Whether the underscores from the mangled name should be omitted.
 remove_underscores = false
 
+# Table of replacement strings for substituting type parameters when name mangling.
+# The keys are processed type names, and the values are the replacement strings.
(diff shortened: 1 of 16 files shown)


emilio (commented):
Can you elaborate on the use cases for this that can't be done via just inserting a typedef or such in the prefix? In general I'm not thrilled to add another layer to mangling, this doesn't feel quite like mangling IMO.

YuhanLiin:
It's mainly useful for generic enums, since it allows users to control how enum variant names are mangled. So something like `enum Foo<Bar> { Variant }` doesn't generate a name like `FooVariant_Bar`. In fact you can even disable the mangling altogether by replacing `Bar` with nothing, so you just get `FooVariant` instead. I think this is the only feature that can't be replicated with typedefs, so we could add it as a standalone feature instead.

PR #721, opened 2021-11-02
RFC: Map RawFd to Int

elmarco:
On all Unix systems, RawFd is defined as c_int.


emilio (approved):
Makes sense to me, thanks.

elmarco:
thanks for the quick reply!

PR #727, opened 2022-01-07
Fixes2 - fix array types for C

mjhatamy:
Now Array types are passed properly.

Example: 
```Rust
pub extern "C" fn key_to_hex(raw_hex: &mut [c_char; GXT_KEY_LEN_HEX], raw_key: &[u8; GXT_KEY_LEN])
```

```C
void key_to_hex(char rawHex[GXT_KEY_LEN_HEX], uint8_t rawKey[GXT_KEY_LEN]);
```

### DIFF
--- a/src/bindgen/ir/function.rs
+++ b/src/bindgen/ir/function.rs
@@ -385,9 +385,9 @@ impl SynFnArgHelpers for syn::FnArg {
                     Some(x) => x,
                     None => return Ok(None),
                 };
-                if let Type::Array(..) = ty {
-                    return Err("Array as function arguments are not supported".to_owned());
-                }
+                // if let Type::Array(..) = ty {
+                //     return Err("Array as function arguments are not supported".to_owned());
+                // }
                 Ok(Some(FunctionArgument {
                     name,
                     ty,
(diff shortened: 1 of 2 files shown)


emilio (changes requested):
This is not right, unless I'm missing something.

In C, the array decays to a pointer (the size doesn't even matter, actually). In Rust a slice reference is effectively a struct {ptr, size}. So this will crash I'm pretty sure.

emilio:
Ah, i might be wrong actually. I was thinking of slices. Does rust just pass a pointer to the first element for arrays?

mjhatamy:
@emilio , We could pass arguments to a function as a pointer or reference or every declare it as an array.

For example:
 1- void test1(char rawHex[5]);
 2- void test1(char *rawHex);
3 - void test1(char(* rawHex)[5]);

All are valid, but here we want to generate a C binding for a rust library. It is much easier and better to have a more meaningful function and arguments details and safer boundaries.
This way, in a large project, team members, will be able to write codes faster and better, and safer.

emilio:
I agree, my point is: Does rust actually use the same ABI for `arg: &[T]` than C uses for `*const T`? I'd need to check but if you know the answer that'd be great.

emilio:
I think the patch might be incorrect as per the above, but happy to be wrong. Closing as for the lack of response and tests failing.

PR #740, opened 2022-03-20
Update clap to 3.1

messense:


messense:
Unfortunately clap 3 requires MSRV 1.54.0.

emilio (approved):
Looks good, thank you. A bit unfortunate having to raise the MSRV for this but I'm doing to do a release ASAP so we may as well. Thanks!

PR #742, opened 2022-03-23
Add support for generic struct parametrized by unit type

### LINKED ISSUE #659: Generic types parametrized by the unit type fail to be generated
A minimal example:
```rust
#[repr(C)]
pub struct MyStruct<T> {
    int_field: u32,
    generic_field: T,
}

#[no_mangle]
pub extern "C" fn my_test() -> MyStruct<()> {
    MyStruct {
        int_field: 0,
        generic_field: ()
    }
}
```

This gives `WARN: Can't find MyStruct. This usually means that this type was incompatible or not found.` and doesn't generate the MyStruct type.

I'd imagine intended output would look something like this?

```
typedef struct MyStruct_unit {
  uint32_t int_field;
}
```

In my particular case, this would help with returning a result type that has no payload data from an exported function. Thanks!

vmx:
It is no possible to paraemtrize a generic struct by the unit type.
The resulting struct will have the suffix `_c_void`.

This Rust code:

```
pub struct MyStruct<T> {
    int_field: u32,
    generic_field: T,
}

pub extern "C" fn my_test() -> MyStruct<()> {
    MyStruct {
        int_field: 0,
        generic_field: ()
    }
}
```

Will result in this struct in C:

```
typedef struct {
  uint32_t int_field;
} MyStruct_c_void;
```

Fixes #659.

### DIFF
--- a/src/bindgen/ir/structure.rs
+++ b/src/bindgen/ir/structure.rs
@@ -11,7 +11,7 @@ use crate::bindgen::declarationtyperesolver::DeclarationTypeResolver;
 use crate::bindgen::dependencies::Dependencies;
 use crate::bindgen::ir::{
     AnnotationSet, Cfg, ConditionWrite, Constant, Documentation, Field, GenericParams, Item,
-    ItemContainer, Path, Repr, ReprAlign, ReprStyle, ToCondition, Type, Typedef,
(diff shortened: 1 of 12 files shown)


vmx:
As it breaks other tests, I'm trying another approach, stay tuned.

vmx:
I sadly couldn't find a good way quickly. I don't know if I'll find the time coming back to it, hence closing the PR for now.

PR #748, opened 2022-03-31
Cython: Omit per-variant tags in unions generated for Rust enums

petrochenkov:
The common `tag` should already be enough, all other tags are identical to it.
In Cython case they only create noise because the declarations only introduce names and do not determine layouts.

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -710,7 +710,7 @@ impl Source for Enum {
             }
 
             // Emit fields for all variants with data.
-            self.write_variant_fields(config, out);
+            self.write_variant_fields(config, out, inline_tag_field);
 
             // Close union of all variants with data, only in the non-inline tag scenario.
             // See the comment about Cython on `open_brace`.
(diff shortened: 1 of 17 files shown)


src/bindgen/ir/enumeration.rs
```diff
@@ -962,11 +967,15 @@ impl Enum {
...
                     // support unnamed structs.
+                    // For the same reason with Cython we can omit per-variant tags (the first
+                    // field) to avoid extra noise, the main `tag` is enough in this case.
                     if config.language != Language::Cython {
                         out.write("struct");
                         out.open_brace();
                     }
-                    out.write_vertical_source_list(&body.fields, ListType::Cap(";"));
+                    let start_field =
+                        usize::from(inline_tag_field && config.language == Language::Cython);
```
emilio, line 977: A bit subtle, might be a bit clearer with:

```rust
let fields_to_write = if inline_tag_field && config.language == Language::Cython {
    &body.fields[1..]
} else {
    &body.fields
};
```

But sure, fine.

emilio (approved):

PR #751, opened 2022-04-13
Support rename rule for union body members

kavoor:
I am trying to use C++ macros to generate accessors for the fields in a rust enum. Since they are converted to SnakeCase by default, its impossible to write a macro that does this using only the original enum variant name. 

Can we add an option to support different rename rules for the union body members as well?

### DIFF
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -559,6 +559,10 @@ impl StructConfig {
 pub struct EnumConfig {
     /// The rename rule to apply to the name of enum variants
     pub rename_variants: RenameRule,
+    /// The rename rule to apply to the names of the union fields in C/C++
+    /// generated from the Rust enum. Applied before rename_variants
+    /// rename rule. Defaults to SnakeCase.
+    pub rename_variant_name_fields: RenameRule,
     /// Whether to add a `Sentinel` value at the end of every enum
     /// This is useful in Gecko for IPC serialization
     pub add_sentinel: bool,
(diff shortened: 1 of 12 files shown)


kavoor:
@emilio What's the time table for getting something like this into a release?

src/bindgen/config.rs
```diff
@@ -559,6 +559,9 @@ impl StructConfig {
 pub struct EnumConfig {
     /// The rename rule to apply to the name of enum variants
     pub rename_variants: RenameRule,
+    /// The rename rule to apply to the names of union body members.
+    /// Applied before rename_variants rename rule. Defaults to SnakeCase.
+    pub rename_union_body_members: RenameRule,
```
emilio, line 564: So this is not for Rust unions (as in `union Foo {}` but C/C++ unions). I think this should just be `rename_fields` for consistency with `config.{struct,union}.rename_fields`, wdyt?

emilio (changes requested):
This needs a test, and I think the option naming could be more consistent? But let me know if you feel strongly, I guess just `rename_fields` could be seen as renaming the `foo` in `Bar { foo: i32, bar: i32 }`, even though it wouldn't.

In terms of getting it to a release, once it's merged I'm happy to publish it right away.

src/bindgen/config.rs
emilio: Maybe `rename_variant_name_field` would be more explicit? I dunno.

kavoor:
Whenever I run 'cargo test' I get errors like " panicked at 'build should succeed: CargoExpand("expand-dep", Compile("" which seem unrelated to my changes. I looked at contributing.md and only saw something about installing cython, which I did.

emilio:
On which OS? Does running `rustup override add nightly` help?

kavoor:
Thanks, that ended up being the issue! I think this should work now. Let me know if I need other tests.

emilio (approved):
Thanks, looks good!

emilio:
Ah, rustfmt is complaining about the trailing whitespace, but I can address manually, assuming all other tests pass.

PR #759, opened 2022-05-07
Const generics

jorendorff:
A rough implementation of const generics.

Well, this codebase is a delight to work with. I thought of this as a tricky feature, but it was mostly obvious what to do.

It helped that there was already a place where values could appear in types: array lengths. This reuses `ArrayLength`. (I can rename that to `ConstExpr`, if you like.)

The main iffy thing with this PR is that it monomorphizes `ArrayString<20>` and `ArrayString<N>` to two different C types, `ArrayString_20` and `ArrayString_N`, even if `N` is 20. To fix that, we'd have to fully evaluate constant expressions, which requires implementing Rust name lookups. I might need a little help if you want to go that route...

### DIFF
--- a/src/bindgen/ir/generic_path.rs
+++ b/src/bindgen/ir/generic_path.rs
@@ -5,7 +5,7 @@ use syn::ext::IdentExt;
 
 use crate::bindgen::config::{Config, Language};
 use crate::bindgen::declarationtyperesolver::{DeclarationType, DeclarationTypeResolver};
-use crate::bindgen::ir::{Path, Type};
+use crate::bindgen::ir::{ArrayLength, Path, Type};
 use crate::bindgen::utilities::IterHelpers;
 use crate::bindgen::writer::{Source, SourceWriter};
 
@@ -19,7 +19,8 @@ impl GenericParams {
                 .params
                 .iter()
                 .filter_map(|x| match *x {
-                    syn::GenericParam::Type(syn::TypeParam { ref ident, .. }) => {
+                    syn::GenericParam::Type(syn::TypeParam { ref ident, .. })
+                    | syn::GenericParam::Const(syn::ConstParam { ref ident, .. }) => {
                         Some(Path::new(ident.unraw().to_string()))
                     }
                     _ => None,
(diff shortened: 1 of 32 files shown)


src/bindgen/ir/generic_path.rs
```diff
@@ -123,11 +235,15 @@ impl GenericPath {
...
+    pub fn is_single_identifier(&self) -> bool {
+        self.generics.is_empty()
+    }
+
     pub fn rename_for_config(&mut self, config: &Config, generic_params: &GenericParams) {
         for generic in &mut self.generics {
             generic.rename_for_config(config, generic_params);
         }
-        if !generic_params.contains(&self.path) {
+        if !generic_params.iter().any(|param| param.name == self.path) {
```
emilio, line 246: Might be worth implementing a proper `contains` in `GenericParams` perhaps, if this check needs to be done in many places... For now it's just two so it's probably fine.

emilio (approved):
Whoa, this is great, thank you so much! I don't really have many concerns after having taken a look, it ended up being much simpler than I expected when I read the PR title ;)

The quirk you point out in the PR description is somewhat annoying to address very generally, because cbindgen doesn't understand much of the Rust module type system.

It's an issue for typedefs as well, see:

```rust
#[repr(i32)]
pub enum Foo { A, B }

#[repr(C)]
pub struct Generic<T> {
    t: T,
}

pub type Bar = Foo;

#[no_mangle]
pub extern "C" fn root(a: &Generic<Foo>, b: &Generic<Bar>) {}
```

Which generates:

```c
#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>

enum Foo {
  A,
  B,
};
typedef int32_t Foo;

typedef struct Generic_Foo {
  Foo t;
} Generic_Foo;

typedef Foo Bar;

typedef struct Generic_Bar {
  Bar t;
} Generic_Bar;

void root(const struct Generic_Foo *a, const struct Generic_Bar *b);
```

So I think it's something that we can punt on for now, in the sense that we're not generating unsound code or anything...

emilio:
(Ah, I forgot that CI doesn't run by default for new contributors, so I pushed the button to approve it, and will merge it once it's green. Feel free to ping if I forget. Thanks again!)

jorendorff:
Friendly ping! This is green now.

PR #760, opened 2022-05-11
Const generics: handle bool and char literals

jorendorff:
This renames `ArrayLength` to `ConstExpr` and slightly expands the range of supported arguments to const generics.

Previously only identifiers and int literals were allowed. With this PR, `true` and `false` work:

```
#[repr(C)]
pub struct HashTable<K, V, const IS_MAP: bool> {
    ...
}

type MySet = HashTable<Str, u8, false>;  // <--- passing `false` here works now
type MyMap = HashTable<Str, Str, true>;
```

In C, these are mangled as `true` and `false`. The PR also adds char and byte literals, for good measure. They mangle as the integer value of the literal.

### DIFF
--- a/src/bindgen/cdecl.rs
+++ b/src/bindgen/cdecl.rs
@@ -5,7 +5,7 @@
 use std::io::Write;
 
 use crate::bindgen::declarationtyperesolver::DeclarationType;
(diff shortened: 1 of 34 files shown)


emilio (approved):
LGTM!

Issue #761, opened 2022-05-11
Bug nesting const generics 


jorendorff:
The test case below generates broken C code. 

```rust
// Propagating const arguments through generics that use generics.

#[repr(C)]
struct Inner<N: usize> {
    bytes: [u8; N],
}

#[repr(C)]
struct Outer<N: usize> {
    inner: Inner<N>, // don't declare two different structs named `Inner_N`
}

#[no_mangle]
pub extern "C" fn one() -> Outer<1> {
    Outer { inner: Inner { bytes: [0] } }
}

#[no_mangle]
pub extern "C" fn two() -> Outer<2> {
    Outer { inner: Inner { bytes: [0, 0] } }
}
```

`cargo run -- --lang c --style tag $FILE` produces this:

```c
#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>

struct Inner_N {
  uint8_t bytes[N];
};

struct Outer_1 {
  struct Inner_N inner;
};

struct Outer_2 {
  struct Inner_N inner;
};

struct Outer_1 one(void);

struct Outer_2 two(void);
```

The corresponding thing works fine with type parameters, so there must be something I failed to hook up in #759, or maybe we aren't modeling the value namespace at all. Filing the bug because I don't have time to look right now. I should have time in the next few days, though.

PR #762, opened 2022-05-14
Fix specialization of `SomeType<N>` when `N` is a const parameter.

jorendorff:
See #761. The test case that triggers the bug is like this:

```rust
#[repr(C)]
pub struct Inner<const N: usize> {
    pub bytes: [u8; N],
}

#[repr(C)]
pub struct Outer<const N: usize> {
    pub inner: Inner<N>, // BUG: declares two different structs named `Inner_N`
}

...code that uses Outer<1> and Outer<2>...
```

When monomorphizing `Outer<1>`, specializing the type `Inner<N>` should result in `Inner<1>`, but it produced `Inner<N>` instead. This happens because `N` was initially parsed as a type, not a constant, so we were thoughtlessly calling `Type::specialize` on it. This can't possibly work; the desired answer `1` is not a type.

The fix is to explicitly handle this special case where the identifier was initially parsed as a type but it's a const.

### DIFF
--- a/src/bindgen/ir/generic_path.rs
+++ b/src/bindgen/ir/generic_path.rs
@@ -159,7 +159,20 @@ pub enum GenericArgument {
 impl GenericArgument {
     pub fn specialize(&self, mappings: &[(&Path, &GenericArgument)]) -> GenericArgument {
         match *self {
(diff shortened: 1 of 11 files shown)


jorendorff:
Friendly review ping. :)

emilio (approved):

PR #778, opened 2022-08-16
Use mutable reference in Bindings::write

### LINKED ISSUE #776: `Bindings::write<F: Write>` consumes argument
I'm trying to use `cbindgen` to generate a bunch of c++ structures for use in JSON deserialization via [nlohmann/json](https://github.com/nlohmann/json).  I don't need (or really want) full binary interop as I'm not calling the rust library directly and all info is passed through strings.

I'd like therefore to manipulate the generated header file some before it's written.  Mostly I'm just replacing some opaque types with `std::string` for the initial parsing.

It seems like I could use a `std::io::BuffWriter` object with `Bindings::write`, then further manipulate the results before writing the header file to disk but `Bindings::write` consumes it's argument:

```rust
let buffer = std::io::BufWriter::new(Vec::new());
cbindgen::Builder::new()
    .generate().unwrap()
    .write(buffer);

// Can't then get the result here:
let bytes = buffer.into_inner();
```

I'd be happy to work on a PR to enable this unless there's a better way of going about things.

berkowski:
fixes #776

Allows further manipulation of generated code after calling `Bindings::write`

### DIFF
--- a/src/bindgen/bindings.rs
+++ b/src/bindgen/bindings.rs
@@ -138,7 +138,8 @@ impl Bindings {
             if let Some(parent) = path::Path::new(path.as_ref()).parent() {
                 fs::create_dir_all(parent).unwrap();
             }
-            self.write(File::create(path).unwrap());
+            let mut out = File::create(path).unwrap();
+            self.write(&mut out);
             return true;
         }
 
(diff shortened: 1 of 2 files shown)


berkowski:
Turns out this isn't necessary.  See comments at end of linked issue.

PR #785, opened 2022-09-14
Merge generic enum tags

### LINKED ISSUE #784: Merge generic tag enums
At the moment, when a generic tagged enum is used, a tag enum is generated for every specialization of the generic. However, the possible tags and their values do not change based on the generic parameters, so these tag enums are identical.

There should be an option to merge these enums into one that ignores the generics, perhaps `enum.merge_generic_tag`.

This would shorten types and improve readability for common use cases such as result types, where the tag is often checked manually after a function call, especially in cases where the generic types are long.

# Example

Rust code:

```rust
#[repr(C)]
pub enum COption<T> {
    Some(T),
    None,
}

#[no_mangle]
pub extern "C" fn example(_: COption<u8>, _: COption<u32>) {}
```

Current output:

```c
typedef enum COption_u8_Tag {
  Some_u8,
  None_u8,
} COption_u8_Tag;

typedef struct COption_u8 {
  COption_u8_Tag tag;
  union {
    struct {
      uint8_t some;
    };
  };
} COption_u8;

typedef enum COption_u32_Tag {
  Some_u32,
  None_u32,
} COption_u32_Tag;

typedef struct COption_u32 {
  COption_u32_Tag tag;
  union {
    struct {
      uint32_t some;
    };
  };
} COption_u32;
```

Ideal output:

```c
typedef enum COption_Tag {
  Some,
  None,
} COption_Tag;

typedef struct COption_u8 {
  COption_Tag tag;
  union {
    struct {
      uint8_t some;
    };
  };
} COption_u8;

typedef struct COption_u32 {
  COption_Tag tag;
  union {
    struct {
      uint32_t some;
    };
  };
} COption_u32;
```

Benjamin-Davies:
See attached issue.

Resolves #784

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -657,7 +657,7 @@ impl Item for Enum {
 
         let monomorph = Enum::new(
             mangled_path,
-            self.path.clone(),
+            self.unmangled_path.clone(),
             GenericParams::default(),
             self.repr,
             self.variants
@@ -775,10 +775,7 @@ impl Enum {
         tag_name: &str,
     ) {
         // Only emit the tag enum once if merge_generic_tags is set.
-        if config.language != Language::Cxx
-            && config.enumeration.merge_generic_tags
-            && self.tag_written.get()
-        {
+        if config.enumeration.merge_generic_tags && self.tag_written.get() {
             return;
         }
         self.tag_written.set(true);
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -270,9 +270,13 @@ impl EnumVariant {
         mappings: &[(&Path, &GenericArgument)],
         config: &Config,
     ) -> Self {
+        let name = if config.enumeration.merge_generic_tags {
+            self.name.clone()
+        } else {
+            mangle::mangle_name(&self.name, generic_values, &config.export.mangle)
+        };
         Self::new(
-            // mangle::mangle_name(&self.name, generic_values, &config.export.mangle),
-            self.name.clone(),
+            name,
             self.discriminant.clone(),
             self.body.specialize(generic_values, mappings, config),
             self.cfg.clone(),
@@ -321,7 +325,9 @@ impl Source for EnumVariant {
 #[derive(Debug, Clone)]
 pub struct Enum {
     pub path: Path,
+    pub unmangled_path: Path,
     pub export_name: String,
+    pub unmangled_export_name: String,
     pub generic_params: GenericParams,
     pub repr: Repr,
     pub variants: Vec<EnumVariant>,
@@ -434,6 +440,7 @@ impl Enum {
         };
 
         Ok(Enum::new(
+            path.clone(),
             path,
             generic_params,
             repr,
@@ -448,6 +455,7 @@ impl Enum {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
         path: Path,
+        unmangled_path: Path,
         generic_params: GenericParams,
         repr: Repr,
         variants: Vec<EnumVariant>,
@@ -457,9 +465,12 @@ impl Enum {
         documentation: Documentation,
     ) -> Self {
         let export_name = path.name().to_owned();
+        let unmangled_export_name = unmangled_path.name().to_owned();
         Self {
             path,
+            unmangled_path,
             export_name,
+            unmangled_export_name,
             generic_params,
             repr,
             variants,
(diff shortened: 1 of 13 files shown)


morrisonlevi:
Looks good to me as a total outsider. @emilio, are you the right person to ping these days? You have the most commits last year.

emilio (commented):
Sorry for the lag here, I haven't been able to take a look lately at cbindgen stuff.

In general the feature makes sense, but instead of the `Rc<Cell<>>` shenanigans and having to keep the unmangled paths around, it seems it could be implemented more cleanly and efficiently in `instantiate_monomorph`. At that point, you know you have a generic, and you're going to monomorphize it. It should be feasible to synthesize a tag enum there, and just carry an `is_monomorphic: bool` or so to avoid generating the tag if needed.

src/bindgen/ir/enumeration.rs
```diff
@@ -438,6 +439,7 @@ impl Enum {
             repr,
             variants,
             tag,
+            false,
```
emilio, line 442: `/* external_tag = */ false,` please :)

tests/tests.rs
```diff
@@ -78,6 +78,11 @@ fn run_cbindgen(
         output,
         str::from_utf8(&cbindgen_output.stderr).unwrap_or_default()
     );
+    if (language, style, cpp_compat) == (Language::C, Some(Style::Type), true) {
+        std::io::stderr()
+            .write_all(&cbindgen_output.stderr)
+            .unwrap();
+    }
```
emilio, line 85: This looks somewhat drive-by? Is this intentionally left here?

src/bindgen/ir/enumeration.rs
```diff
@@ -648,6 +681,18 @@ impl Item for Enum {
     }
 
     fn add_dependencies(&self, library: &Library, out: &mut Dependencies) {
+        if let Some(tag) = self.tag.clone() {
+            let path = Path::new(tag);
+
+            // If there is an external tag enum, then add it as a dependency.
+            if let Some(items) = library.get_items(&path) {
```
emilio, line 688: This looks a lot like what `Type::add_dependencies_ignoring_generics` does for a `Type::path`, can we factor this out into a common method in `Dependencies`? Something like `Dependencies::add_path(&mut self, &Library, &Path)` or so?

tests/tests.rs
Benjamin-Davies: Oops, sorry

mxyns:
hi all, i am interested in getting this PR merged as it would greatly help me in my project :slightly_smiling_face: Is there something I can do to make this happen? 
@emilio are you still the person to ping or should is there someone else you can refer us to?

mxyns:
hi @emilio 
Do you think this could be merged soon? 
It looks like @Benjamin-Davies has addressed your last comment about tidying up commits.

morrisonlevi:
I emailed Emilio and Ben and Ben gave approval for squash-merging! So we just need to get conflicts resolved, and we should good to go.

morrisonlevi:
Example rust code:

```rs
#[repr(C)]
pub enum Result<T, E> {
    Ok(T),
    Err(E),
}

#[no_mangle]
pub extern "C" fn fn1() -> Result<u32, u64> {
    Result::Ok(0)
}

#[no_mangle]
pub extern "C" fn fn2() -> Result<u64, u32> {
    Result::Ok(0)
}
```

With this `cbindgen.toml`:

```toml
language = "C"
include_guard = "DDOG_CBINDGEN_TEST_H"
no_includes = true
sys_includes = ["stdbool.h", "stddef.h", "stdint.h"]

[export]
prefix = "ddog_cbindgen_"

[enum]
merge_generic_tags = true
prefix_with_name = true
```

Results in this output:

```c
#ifndef DDOG_CBINDGEN_TEST_H
#define DDOG_CBINDGEN_TEST_H

#include <stdbool.h>
#include <stddef.h>
#include <stdint.h>

typedef enum ddog_cbindgen_Result_Tag {
  DDOG_CBINDGEN_RESULT_OK,
  DDOG_CBINDGEN_RESULT_ERR,
} ddog_cbindgen_Result_Tag;

typedef enum Result_Tag {
  DDOG_CBINDGEN_RESULT_U32_U64_OK_U32_U64,
  DDOG_CBINDGEN_RESULT_U32_U64_ERR_U32_U64,
} Result_Tag;

typedef struct ddog_cbindgen_Result_u32__u64 {
  Result_Tag tag;
  union {
    struct {
      uint32_t ok;
    };
    struct {
      uint64_t err;
    };
  };
} ddog_cbindgen_Result_u32__u64;

typedef enum Result_Tag {
  DDOG_CBINDGEN_RESULT_U64_U32_OK_U64_U32,
  DDOG_CBINDGEN_RESULT_U64_U32_ERR_U64_U32,
} Result_Tag;

typedef struct ddog_cbindgen_Result_u64__u32 {
  Result_Tag tag;
  union {
    struct {
      uint64_t ok;
    };
    struct {
      uint32_t err;
    };
  };
} ddog_cbindgen_Result_u64__u32;

struct ddog_cbindgen_Result_u32__u64 fn1(void);

struct ddog_cbindgen_Result_u64__u32 fn2(void);

#endif  /* DDOG_CBINDGEN_TEST_H */
```

Notes:
 1. ` enum ddog_cbindgen_Result_Tag` and `Result_Tag` are both present. The latter shouldn't be there.
 2. `Result_Tag` has cases named like `DDOG_CBINDGEN_RESULT_U64_U32_OK_U64_U32` instead of using `DDOG_CBINDGEN_RESULT_OK`.

PR #792, opened 2022-10-17
fix/docs_typo

novafacing:
Tiny fix, the documented example has the `documentation_style` key twice instead of
`documentation_length`.


emilio (approved):
Thanks!

PR #793, opened 2022-10-21
Better line breaking in function pointer types

fredszaq:
The current implementation never goes to a new line when writing a function pointer type, this can lead to long, difficult to read lines.

The goal of this change is to make that a bit more sensible.

Here is a rust code example

```
   1   │ pub type MyCallback = Option<unsafe extern "C" fn(a: usize, b: usize)>;
   2   │
   3   │ pub type MyOtherCallback =
   4   │     Option<unsafe extern "C" fn(a: usize, lot: usize, of: usize, args: usize)>;
   5   │
   6   │ #[no_mangle]
   7   │ pub extern "C" fn my_function(a: MyCallback, b: MyOtherCallback) {}
```

right now when generating the corresponding C header we get

```
   1   │ #include <stdarg.h>
   2   │ #include <stdbool.h>
   3   │ #include <stdint.h>
   4   │ #include <stdlib.h>
   5   │
   6   │ typedef void (*MyCallback)(uintptr_t a, uintptr_t b);
   7   │
   8   │ typedef void (*MyOtherCallback)(uintptr_t a, uintptr_t lot, uintptr_t of, uintptr_t args);
   9   │
  10   │ void my_function(MyCallback a, MyOtherCallback b);
```

line 8 here is already quite long and will be even longer if we add new args to `MyOtherCallback`

With the changes in this commit, we now get

```
   1   │ #include <stdarg.h>
   2   │ #include <stdbool.h>
   3   │ #include <stdint.h>
   4   │ #include <stdlib.h>
   5   │
   6   │ typedef void (*MyCallback)(uintptr_t a, uintptr_t b);
   7   │
   8   │ typedef void (*MyOtherCallback)(uintptr_t a,
   9   │                                 uintptr_t lot,
  10   │                                 uintptr_t of,
  11   │                                 uintptr_t args);
  12   │
  13   │ void my_function(MyCallback a, MyOtherCallback b);
```

which is way better and more scalable if new args are added to `MyOtherCallback`

The behavior is configurable using the already existing `fn.args` configuration parameter. In this case setting it to `Horizontal` gives back the same .h as previously and setting it to `Vertical` makes the generator go to a new line even for the shorter `MyCallback` declaration:

```
   1   │ #include <stdarg.h>
   2   │ #include <stdbool.h>
   3   │ #include <stdint.h>
   4   │ #include <stdlib.h>
   5   │
   6   │ typedef void (*MyCallback)(uintptr_t a,
   7   │                            uintptr_t b);
   8   │
   9   │ typedef void (*MyOtherCallback)(uintptr_t a,
  10   │                                 uintptr_t lot,
  11   │                                 uintptr_t of,
  12   │                                 uintptr_t args);
  13   │
  14   │ void my_function(MyCallback a,
  15   │                  MyOtherCallback b);
```

### DIFF
--- a/src/bindgen/cdecl.rs
+++ b/src/bindgen/cdecl.rs
@@ -4,11 +4,11 @@
 
 use std::io::Write;
 
+use crate::bindgen::config::Layout;
 use crate::bindgen::declarationtyperesolver::DeclarationType;
 use crate::bindgen::ir::{ConstExpr, Function, GenericArgument, Type};
 use crate::bindgen::writer::{ListType, SourceWriter};
 use crate::bindgen::{Config, Language};
-use crate::bindgen::config::Layout;
 
 // This code is for translating Rust types into C declarations.
 // See Section 6.7, Declarations, in the C standard for background.
@@ -289,7 +289,11 @@ impl CDecl {
                         out.write("void");
                     }
 
-                    fn write_vertical<F:Write>(out: &mut SourceWriter<F>, config: &Config, args: &Vec<(Option<String>, CDecl)>) {
+                    fn write_vertical<F: Write>(
+                        out: &mut SourceWriter<F>,
+                        config: &Config,
+                        args: &Vec<(Option<String>, CDecl)>,
+                    ) {
                         let align_length = out.line_length_for_align();
                         out.push_set_spaces(align_length);
                         for (i, &(ref arg_ident, ref arg_ty)) in args.iter().enumerate() {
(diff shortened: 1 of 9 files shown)


fredszaq:
Hey @emilio! Did you have time to take a look at this? I can do any change that you would deem necessary. Thanks a lot!

src/bindgen/cdecl.rs
```diff
@@ -314,6 +318,19 @@ impl CDecl {
                             arg_ty.write(out, arg_ident, config);
                         }
                     }
+
+
+                    match layout {
+                        Layout::Vertical => write_vertical(out, config, args),
+                        Layout::Horizontal => write_horizontal(out, config, args),
+                        Layout::Auto => {
+                           if out.line_length_for_align() + out.measure(|out| write_horizontal(out, config,args)) > config.line_length {
```
emilio, line 327: I'm not a fan of having to do the work twice in the common case, but we should be able to improve it. Looks good, thanks!

emilio (approved):

emilio:
Would be great to get a test for this, and it seems it needs formatting tho. Can you get to it? Otherwise I can try to do it when I find some time this week.

fredszaq:
added the test and formatted the code

src/bindgen/cdecl.rs
fredszaq: what do you mean" do the work twice" ?

fredszaq:
great ! can we expect a version with it ?

PR #796, opened 2022-11-10
Add with_cpp_compat to the builder

irh:
I noticed recently that the `cpp_compat` flag isn't available when using the builder, so I figured I'd open a PR.


emilio (approved):
Thanks!

PR #799, opened 2022-11-13
Move expand infinite recursion fix

### LINKED ISSUE #758: Having problems expanding macros with build.rs
I encountered a problem with the expansion of macros using build.rs. Here is a minimal example to point out the problems I'm having. I have the following `lib.rs`.  

```rust 
// lib.rs file

macro_rules! number {
    ($name:ident, $number:expr) => {
        #[no_mangle]
        pub extern "C" fn $name () -> u64 {
            $number
        }
    }
}

number!(two, 2);
```

with the following `cbindgen.toml`

```rust
language = "C"

[parse.expand]
crates = ["testing"]
```

If I run `rustup run nightly cbindgen ./ --config cbindgen.toml --crate testing --output bindings.h`, the macros are expanded, and I get the expected file `bindings.h`. 

However, if I try to expand macros and build the binding using a `build.rs` file, the compilation takes a lot and returns an error at the end of it. The build files I've tried are the following: 

```rust 
// With the builder struct

extern crate cbindgen;
use std::env;

fn main() {
    cbindgen::Builder::new()
        .with_crate(env::var("CARGO_MANIFEST_DIR").unwrap())
        .with_parse_expand(&["testing"])
        .generate()
        .expect("Unable to generate bindings")
        .write_to_file("bindings.h");
}
```

and 

```rust 
// With the generate function using the same cbindgen.toml as above
extern crate cbindgen;

fn main() {
    cbindgen::generate(std::env::var("CARGO_MANIFEST_DIR").unwrap())
        .expect("Unable to generate bindings")
        .write_to_file("bindings.h");
}
```

When running `RUST_BACKTRACE=1 cargo build` I get a ton of lines like this (not copying all because would be huge): 

```shell
\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\n               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/ops/function.rs:227:5\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\
```

and the final lines of the error are 
```shell
  stack backtrace:
     0: rust_begin_unwind
               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/std/src/panicking.rs:498:5
     1: core::panicking::panic_fmt
               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/panicking.rs:116:14
     2: core::result::unwrap_failed
               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/result.rs:1690:5
     3: core::result::Result<T,E>::expect
               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/result.rs:975:23
     4: build_script_build::main
               at ./build.rs:4:5
     5: core::ops::function::FnOnce::call_once
               at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/ops/function.rs:227:5
  note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

```

Is this a known issue? Or am I maybe doing something wrong? 

Thank you!

rmalmain:
The original fix was supposed to avoid infinitely expanding the macros when cbindgen was generated in `build.rs` with the `parse.expand` option.

However, it was not completely fixing the issue: a call to `parse_lib` which triggers the infinite recursion occurs before the old check was actually done (in the parent `generate` method).

Here, the check is simply done earlier to completely avoid the infinite loop.

I believe this fix would resolve #758.

### DIFF
--- a/src/bindgen/builder.rs
+++ b/src/bindgen/builder.rs
@@ -345,6 +345,24 @@ impl Builder {
     }
 
     pub fn generate(self) -> Result<Bindings, Error> {
+        // If macro expansion is enabled, then cbindgen will attempt to build the crate
+        // and will run its build script which may run cbindgen again. That second run may start
+        // infinite recursion, or overwrite previously written files with bindings.
+        // So if we are called recursively, we are skipping the whole generation
+        // and produce "noop" bindings that won't be able to overwrite anything.
+        if std::env::var("_CBINDGEN_IS_RUNNING").is_ok() {
+            return Ok(Bindings::new(
+                self.config,
+                Default::default(),
+                Default::default(),
+                Default::default(),
+                Default::default(),
+                Default::default(),
+                Default::default(),
+                true,
+            ));
+        }
+
         let mut result = Parse::new();
 
         if self.std_types {
--- a/src/bindgen/library.rs
+++ b/src/bindgen/library.rs
@@ -54,24 +54,6 @@ impl Library {
     }
 
     pub fn generate(mut self) -> Result<Bindings, Error> {
-        // If macro expansion is enabled, then cbindgen will attempt to build the crate
-        // and will run its build script which may run cbindgen again. That second run may start
-        // infinite recursion, or overwrite previously written files with bindings.
-        // So if we are called recursively, we are skipping the whole generation
-        // and produce "noop" bindings that won't be able to overwrite anything.
-        if std::env::var("_CBINDGEN_IS_RUNNING").is_ok() {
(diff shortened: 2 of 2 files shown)


emilio (approved):

kczimm:
Is there any consideration for having this PR initiate a new `0.24.4` release?

PR #805, opened 2023-01-03
Parse `...` as a `VaList`

### LINKED ISSUE #376: Support variadic arguments (`args: ...` syntax)
There are [many cases in Relibc](https://gitlab.redox-os.org/search?utf8=%E2%9C%93&search=%2F%2F+TODO%3A+Can+be+implemented+in+rust+when+cbindgen+supports+%22...%22+syntax&group_id=&project_id=64&search_code=true&repository_ref=master) where Rust functions are wrapped in C functions only to make them use `...` syntax.

Related Rust RFC: https://github.com/rust-lang/rfcs/blob/master/text/2137-variadic.md

sholtrop:
Closes #376

### DIFF
--- a/src/bindgen/ir/ty.rs
+++ b/src/bindgen/ir/ty.rs
@@ -538,6 +538,9 @@ impl Type {
                 }
                 return Err("Tuples are not supported types.".to_owned());
             }
+            syn::Type::Verbatim(ref tokens) if tokens.to_string() == "..." => {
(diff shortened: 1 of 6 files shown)


emilio (approved):

lygstate:
This doesn't looks right,
according to https://github.com/rust-lang/rfcs/blob/master/text/2137-variadic.md
It's should be 
```
int32_t va_list_test2(...);
```

Issue #812, opened 2023-02-19
Usage of parse.expand fails when using stable Rust Toolchain


doxblek:
The `parse.expand` feature leverages the `-Zunpretty=expanded` flag of the Rust Compiler to expand macros and then evaluating the generated code.

This compiler flag can only be used in a nightly setup. While Rust allows to override the toolchain (e.g. `cargo +nightly …`), this is not possible when using the prebuilt cbindgen CLI tool (e.g. `cbindgen --config cbindgen.toml --output myheader.h`). Calling cbindgen in such way results in a cryptic error message.

Instead, a workaround is required: `rustup default nightly && cbindgen … && rustup default stable`.

This workaround should not be required, since cbindgen could leverage the Rust functionality to override the used toolchain for the current execution. Rustup even provides a flag (`--install`) to install the nightly toolchain if it is not yet installed.

emilio:
This is expected, also `CARGO=$(rustup +nightly which cargo)` should do it.

We don't really want to use nightly without any opt-in from the user, IMHO.

jcaden:
I've submitted a PR that I think solves the issue not forcing the user to have nightly installed https://github.com/eqrion/cbindgen/pull/818

doxblek:
I would like either approach

PR #813, opened 2023-02-19
Enforce usage of nightly toolchain for macro expansion

doxblek:
resolves https://github.com/eqrion/cbindgen/issues/812

See the attached issue

### DIFF
--- a/src/bindgen/cargo/cargo_expand.rs
+++ b/src/bindgen/cargo/cargo_expand.rs
@@ -69,8 +69,15 @@ pub fn expand(
     expand_features: &Option<Vec<String>>,
     profile: Profile,
 ) -> Result<String, Error> {
-    let cargo = env::var("CARGO").unwrap_or_else(|_| String::from("cargo"));
-    let mut cmd = Command::new(cargo);
+    let rustup = String::from("rustup");
+    let mut cmd = Command::new(rustup);
+    cmd.arg("run");
+
+    // Ensure nightly toolchain is installed and used for macro expansion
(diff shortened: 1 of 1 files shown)


jschwe:
I honestly don't think tools like `cbindgen` should go ahead and install / update toolchains. I think providing a nice error message if the nightly toolchain is not used and/or not installed would be preferable. (But I'm not a maintainer here, just someone passing by )

doxblek:
> I honestly don't think tools like `cbindgen` should go ahead and install / update toolchains. I think providing a nice error message if the nightly toolchain is not used and/or not installed would be preferable. (But I'm not a maintainer here, just someone passing by )

I think you are right. Probably it is better to switch to the nightly toolchain but not installing it automatically if it is missing. Along with a descriptive error message in case the toolchain is not installed to guide the user to install it on their own, this might be a good solution.

Could any maintainer give me feedback on that? Then I would update my PR accordingly

emilio:
Yeah, also assuming `rustup` exists is ok for dev machines, but not more generally. I think using nightly needs to be an explicit action on the user, so I think the code as-is is probably fine.

You can trivially use:

```
CARGO=$(rustup +nightly which cargo) cbindgen
```

jcaden:
I've submitted an alternative to this PR #818

emilio:
Yeah, I don't think we want either. There are various ways of making `-Zpretty=expanded` but we don't want to make people accidentally rely on them. This can be done on the caller quite easily if they want.

PR #818, opened 2023-03-08
Enable experimental features when running `-Zunpretty=expanded`

jcaden:


jcaden:
Thanks for your response. 

Could this be added to the documentation. I didn't find it and I struggled a little bit until I managed to make it run with stable version.

PR #819, opened 2023-03-08
CI: Fix warnings in Github actions

jschwe:
Fixes the following warning:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions-rs/cargo@v1, actions-rs/toolchain@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

You can see the warnings at the bottom of example this job: https://github.com/eqrion/cbindgen/actions/runs/4366109547

Steps:
- upgrade checkout to v3
- Replace the actions-rs/cargo action, by simply calling cargo from the commandline. I don't see why one would want to involve a node.js action there in the first place.
- Replace the actions-rs/toolchain with dtolnays version, which uses the shell instead of node.js

### DIFF
--- a/.github/workflows/cbindgen.yml
+++ b/.github/workflows/cbindgen.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     - name: Install stable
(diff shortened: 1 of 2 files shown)


emilio (approved):
Thanks!

PR #820, opened 2023-03-09
Add --depfile option

### LINKED ISSUE #192: Add support for outputting dependencies
It would be nice if cbindgen could provide the full list of input files it used to generate proper dependencies for build systems. @emilio [is adding cbindgen to the Firefox build](https://phabricator.services.mozilla.com/D2447) and his glue code is manually providing dependencies which makes me worry that that list could be incorrect, which would result in broken builds.

### LINKED ISSUE #815: Add optional depfile output
Hi,

I'm working on integrating cbindgen into a foreign build system (CMake), and it would be great if `cbindgen` could output a depfile listing the input source files, so my build system knows when to rerun cbindgen for a given crate.
I would also be willing to work on a PR for this feature, if the chances are good for this feature request to be accepted.

jschwe:
Add an option to output a depfile for outside build-systems to learn the file dependencies of the bindings.

cbindgen already keeps track of all the source-files it visits, the information just had to be collected and passed up the callchain.

The output depfile is the common Makefile `.d` file, which cargo also already produes. example:
```
/path/to/generated_header.h: /path/to/src1 /path/to/src2
```

Whitespace in filepaths is escaped to support filepaths containing whitespace. This requires assuming an encoding, so we assume the filepath is utf-8 compliant (and otherwise lossy convert to utf-8).

Since the left hand side requires a file, it only makes sense to generate a depfile if `--out` is also specified, so the current implementation just ignores `--depfile` if `--out` is missing.

I looked at the sorrounding code, and cbindgen is generally very liberal with `unwrap`, so I also didn't add any error handling for this PR. If the filepaths contain non-unicode characters, I chose to use the `lossy` variant for conversion,

I tested this with CMake and Ninja on Linux, and the integration seems to be working as expected (also with paths containing spaces).
I'm not sure yet how to best add tests to cbindgen itself though - the tests seem to be generated from the build-script and I'm not quite sure how to add my one for this feature. Any pointers regarding would be appreciated.

Closes #815 
Closes #192


src/bindgen/bindings.rs
```diff
@@ -128,6 +131,53 @@ impl Bindings {
...
+            // Add line-continue and line-break and then indent with 4 spaces.
+            // This makes the output more human-readable.
+            depfile.write_all(b" \\\n    ").unwrap();
+            let escaped_path = source_file.to_string_lossy().replace(' ', "\\ ");
+            depfile.write_all(escaped_path.as_bytes()).unwrap();
+        });
+
+        writeln!(&mut depfile).unwrap();
+
+        depfile.flush().unwrap();
```
emilio, line 178: Doesn't closing the file flush as well? I doubt this is needed right?

src/bindgen/parser.rs
```diff
@@ -181,6 +181,8 @@ impl<'a> Parser<'a> {
             }
         }
 
+        self.out.source_files = self.cache_src.keys().map(|k| k.to_owned()).collect();
```
emilio, line 184: So this order is basically random, I'd rather have a well defined order by pushing the path in the `Entry::Vacant` case in `parse_mod`. Otherwise let's have some defined order in the depfile generation?

emilio (changes requested):
Thanks, looks reasonable, but I think we should have a defined order to avoid the ordering depending on cache keys / hashmap iteration order.

It'd also be great to have a test for this.

src/bindgen/bindings.rs
jschwe: When the file is dropped, errors when closing are ignored. Manually flushing and unwrapping results in a panic for errors. I don't have a strong opinion here, since errors when closing the file should be extremely rare. I can remove this line if you prefer.

src/bindgen/parser.rs
jschwe: I pushed a commit which sorts the output before writing. This way all source files are sorted, and the config file (if any) is the last entry. This should make writing tests much easier, since it allows us to just compare an output file with an expected file.

jschwe:
> It'd also be great to have a test for this.

Do you have any recommendations for a test? Should I create a new integration test in tests, or integrate this into the `tests` test?

jschwe:
Update: This PR currently doesn't support expanding (since we use `rustc` to expand the crate, and we don't know the source files ourselves). Probably this can be solved by letting rustc itself emit a `.d` file during the expansion and parsing that too.

Maybe it would be better to put that into a seperate PR though, to keep the diff of this one smaller?

jschwe:
I added some tests which use CMake to verify that the generated depfile works as expected (i.e. the bindings aren't regenerated on a second build).
I previously also added `--depfile` to the tests in `tests` in commit [test --depfile](https://github.com/eqrion/cbindgen/pull/820/commits/18351e0f71e1cf5d9df0528bd5a7b1452e7e5e8b), but I'm leaning towards dropping that commit and only keeping the new CMake based test for the --depfile option. What do you think?

I also discovered that implicit configs (`cbindgen.toml` in the root dir) where previously not put into the depfile, so I added a commit which addresses that. Since the `Config` struct and methods are public, I added new internal methods to avoid any semver breaking changes.

jschwe:
friendly ping

jschwe:
@emilio Could I get another review round? Changes since your last review are summarized in this comment: https://github.com/eqrion/cbindgen/pull/820#issuecomment-1465158996

jschwe:
@emilio Could I get another review round? Changes since your last review are summarized in this comment: https://github.com/mozilla/cbindgen/pull/820#issuecomment-1465158996

src/bindgen/parser.rs
```diff
@@ -79,6 +80,7 @@ pub(crate) fn parse_lib(lib: Cargo, config: &Config) -> ParseResult {
 
     let binding_crate = context.lib.as_ref().unwrap().binding_crate_ref();
     context.parse_crate(&binding_crate)?;
+    context.out.source_files = context.cache_src.keys().map(|k| k.to_owned()).collect();
```
emilio, line 83: This order is not deterministic right? We should sort these or something.

src/bindgen/builder.rs
```diff
@@ -293,6 +296,20 @@ impl Builder {
         self
     }
 
+    /// Track the source (file) of the config to support the `--depfile` option
+    #[allow(unused)]
+    pub(crate) fn with_config_and_src(
+        mut self,
+        config: Config,
+        config_src: Option<path::PathBuf>,
```
emilio, line 304: Any reason we could not just store the `src_path: Option<PathBuf>` in the `Config`? That seems like it'd be cleaner?

emilio (changes requested):
Sorry for the lag in re-reviewing this, personal life got in the way.

I think there are only a couple relatively minor issues.

I think making the path a (crate-private) field in the `Config` struct would simplify a bunch of the core failures.

Thanks for all the testing effort, much appreciated.

src/bindgen/builder.rs
jschwe: The `config_src` should contain the path to the config file (if it exists). 
The `Config` itself is parsed from the config file. It would be strange if the user could set a path in the config file, which should point to the config file itself.

jschwe: Finished the change, this is indeed much cleaner.

jschwe:
Probably https://github.com/mozilla/cbindgen/issues/836 should be fixed first before merging, so that the CI can run again.

src/bindgen/parser.rs
emilio: Ah, great, sorry I missed that!

src/main.rs
```diff
@@ -84,7 +84,18 @@ fn load_bindings(input: &Path, matches: &ArgMatches) -> Result<Bindings, Error>
         // Load any config specified or search in the input directory
         let mut config = match matches.value_of("config") {
             Some(c) => Config::from_file(c).unwrap(),
-            None => Config::from_root_or_default(input),
+            None => {
+                // In contrast to `Config::from_root_or_default` we look for
+                // cbindgen.toml next to the current _file_. There is always
+                // a parent (at the very least the root), so unwrapping is fine.
```
emilio, line 90: This change looks unrelated now. Can we undo this? Changing the default behavior of config files is a bit subtle, and there's no test for this change.

tests/depfile.rs
```diff
@@ -0,0 +1,109 @@
+use std::fs::read_to_string;
+use std::path::PathBuf;
+use std::process::Command;
+
+static CBINDGEN_PATH: &str = env!("CARGO_BIN_EXE_cbindgen");
+
+fn test_project(name: &str, project_path: &str) {
```
emilio, line 7: `name` seems unused.

tests/tests.rs
```diff
@@ -2,10 +2,15 @@ extern crate cbindgen;
...
 use cbindgen::*;
 use std::collections::HashSet;
+use std::fs::File;
+use std::io::Read;
 use std::path::Path;
 use std::process::Command;
 use std::{env, fs, str};
 
+// Set automatically by cargo for integration tests
+static CBINDGEN_PATH: &str = env!("CARGO_BIN_EXE_cbindgen");
```
emilio, line 12: Oh, sweet!

src/main.rs
emilio: (That said I think it's probably an uncontroversial improvement, I just rather let it land with a separate PR)

emilio (changes requested):
Just one minor tweak. Let me look at fixing the actions. Looks good with those, thanks!

tests/depfile.rs
jschwe: Removed name

src/main.rs
jschwe: I reverted it. I can open a separate MR afterwards.

emilio:
Can you rebase? We should have GH actions now.

jschwe:
Rebased and squashed into two commits.

emilio (approved):
Awesome, ty!

wongchaiyatham-max:
> Add an option to output a depfile for outside build-systems to learn the file dependencies of the bindings.
> 
> cbindgen already keeps track of all the source-files it visits, the information just had to be collected and passed up the callchain.
> 
> The output depfile is the common Makefile `.d` file, which cargo also already produes. example:
> ```
> /path/to/generated_header.h: /path/to/src1 /path/to/src2
> ```
> 
> Whitespace in filepaths is escaped to support filepaths containing whitespace. This requires assuming an encoding, so we assume the filepath is utf-8 compliant (and otherwise lossy convert to utf-8).
> 
> Since the left hand side requires a file, it only makes sense to generate a depfile if `--out` is also specified, so the current implementation just ignores `--depfile` if `--out` is missing.
> 
> I looked at the sorrounding code, and cbindgen is generally very liberal with `unwrap`, so I also didn't add any error handling for this PR. If the filepaths contain non-unicode characters, I chose to use the `lossy` variant for conversion,
> 
> 
> I tested this with CMake and Ninja on Linux, and the integration seems to be working as expected (also with paths containing spaces).
> I'm not sure yet how to best add tests to cbindgen itself though - the tests seem to be generated from the build-script and I'm not quite sure how to add my one for this feature. Any pointers regarding would be appreciated.
> Invalids"roots-only™"6f2ddc538acb19180f4c8e96cff94ccf27822e8b
> Closes #815 
> Closes #192
>

PR #825, opened 2023-04-13
Avoid 'duplicate package' warnings when depending on cbindgen via git

irh:
When specifying `cbindgen` as a dependency via git, several 'skipping duplicate package' warnings pop up regarding some of the test crates.

The warning seems to be spurious given that the test packages aren't needed when depending on `cbindgen` (see https://github.com/rust-lang/cargo/issues/10752), but while a fix is being considered for Cargo, this PR disambiguates the duplicated package names by referring to their relative paths.


emilio:
Tests seem to fail still, but other than that it looks fine

emilio (approved):

Issue #821, opened 2023-03-15
Release Request for v0.24.4


ophilli:
Hey @emilio, can you please cut a new release on crates.io?

I'd like to have a release that contains https://github.com/eqrion/cbindgen/commit/b6e73017e679caf3b01217e62642d0722a464887

ophilli:
Hey @emilio & @eqrion, could one of y'all please help cut a new release?

ophilli:
Hey @emilio just following up on this again. I'd really love to have a release that contains https://github.com/mozilla/cbindgen/commit/b6e73017e679caf3b01217e62642d0722a464887.

Thanks!

emilio:
Yep, sorry, will try to get to it this today.

ophilli:
Yay, I’m very happy to hear that you’ll try to publish a new release soon! Thanks for taking a look at this @emilio. I know it’s a lot of work to maintain open source and I really appreciate your contributions to the rust community. 😊

ophilli:
Thanks for cutting these releases today Emilio!

PR #834, opened 2023-05-24
Bump version to v0.24.4

ophilli:
Updated the version in Cargo.toml & the changelist, and then ran `cargo update`.

This is in support of #821.


emilio:
Thanks! <3

PR #840, opened 2023-05-30
Prepare new release

jschwe:
Update the Changelog, update the documented MSRV and bump the version in preparation for a new release.


emilio (approved):
Thanks!

PR #843, opened 2023-06-05
 CI: Restore release action and add semver checks to CI deploy job

jschwe:
This MR restores the release action, which creates a github release upon tag.
We now use the github-cli to release the release instead of the third party action, which is not allowed by mozilla.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,7 +14,7 @@ jobs:
       # someone with permissions to create a tag.
       contents: write
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


jschwe:
For this to work, the following actions would need to be allowed by mozilla:

- obi1kenobi/cargo-semver-checks-action (could be worked around by manually installing and using cargo-semver-checks)
- softprops/action-gh-release (could also be manually done, but that would probably be more work)

jschwe:
@emilio This PR is not tested, since I obviously don't have permissions to create a tag. So currently the restored release action only creates a "draft" release, so that we could verify first if everything actually worked as expected.

Could you release v0.25 with this? I'd like to get a release with the depfile changes, and while we are at it we could verify if the release action works as expected.

jschwe:
@emilio Would you have time to review this and/or release v0.25.0?

jschwe:
@emilio Would you have time to review this and/or release v0.25.0?

emilio (approved):

PR #847, opened 2023-06-19
Upgrade clap 3 to clap 4

### LINKED ISSUE #828: Upgrade to clap 4 (getting rid of `atty` as a transitive dependency)
`cbindgen` is still depending on `clap 3` which in turn depends no `atty`. An unmaintained crate with a [known unsoundness issue](https://rustsec.org/advisories/RUSTSEC-2021-0145.html). We are really trying to get rid of `atty` from our dependency tree, but we also need `cbindgen`. So this issue is about considering an upgrade to `clap 4`.

boxdot:
This also removes the dependency on the unmaintained crate atty.

Closes #828

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,14 +3,52 @@
 version = 3
 
 [[package]]
-name = "atty"
-version = "0.2.14"
+name = "anstream"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
+checksum = "0ca84f3628370c59db74ee214b3263d58f9aadd9b4fe7e711fd87dc452b7f163"
 dependencies = [
- "hermit-abi",
- "libc",
- "winapi",
+ "anstyle",
+ "anstyle-parse",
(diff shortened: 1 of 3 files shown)


jschwe:
What is the current minimum supported Rust version for clap 4? Currently the msrv for cbindgen is 1.57, so that would presumably need to be bumped too.

boxdot:
@jschwe Indeed, thank you. I update the version in Cargo.toml. Is it ok to increase it, or does cbindgen has a hard MSRV requirement?

jschwe:
As far as I know (but I'm not a maintainer) there currently is no official MSRV policy. I don't know if cbindgen is used in Firefox, but I could imagine the [firefox msrv policy](https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html) applies also to cbindgen:

> We will update the Minimum Supported Rust Version (MSRV) when required.

romanofski:
Hope I'm not overstepping but pinging @emilio if a merge and release is possible. Sorry otherwise.

emilio (approved):

jschwe:
The PR breaks the "Test package" test in CI (which I can reproduce locally), therefore I propose reverting it for now in #868.

PR #850, opened 2023-06-22
Fix swapping of `>>=` and `<<=` in constants

jschwe:
`ShlEq` is shift left equals, i.e. `<<=`, so it was previously mapped the wrong way.

### DIFF
--- a/src/bindgen/ir/constant.rs
+++ b/src/bindgen/ir/constant.rs
@@ -321,8 +321,8 @@ impl Literal {
                     syn::BinOp::BitXorEq(..) => "^=",
                     syn::BinOp::BitAndEq(..) => "&=",
                     syn::BinOp::BitOrEq(..) => "|=",
-                    syn::BinOp::ShlEq(..) => ">>=",
-                    syn::BinOp::ShrEq(..) => "<<=",
+                    syn::BinOp::ShlEq(..) => "<<=",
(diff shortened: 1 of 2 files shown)


mversic (approved):

src/bindgen/ir/constant.rs
```diff
@@ -321,8 +321,8 @@ impl Literal {
                     syn::BinOp::BitXorEq(..) => "^=",
                     syn::BinOp::BitAndEq(..) => "&=",
                     syn::BinOp::BitOrEq(..) => "|=",
-                    syn::BinOp::ShlEq(..) => ">>=",
-                    syn::BinOp::ShrEq(..) => "<<=",
+                    syn::BinOp::ShlEq(..) => "<<=",
+                    syn::BinOp::ShrEq(..) => ">>=",
```
emilio, line 325: Looks good but would there be any chance to add a test for this?

emilio (approved):

src/bindgen/ir/constant.rs
jschwe: Upon looking at this again, I don't think it is possible to use `<<=` in constants at all, since the left hand side must be a variable. So probably these cases could also be an `unreachable!()` instead.

emilio: Let's merge this since it's trivial, it's not a big deal.

Issue #846, opened 2023-06-13
Support C# and/or expose intermediate representation?


yanchith:
We have a project, where we generate both C and C# bindings for a Rust library. The API is very C-like, and we take care not to use any fancy features in it.

C# bindings are currently generated for us by csbindgen, which is not on par with cbindgen feature-wise (and its authors don't seem like they want it to be). We thus have to employ a lot of hacks to make it work.

cbindgen, on the other hand already supports most of what we need. Is either adding a new C# backend to cbindgen or exposing the intermediate representation so that people can write their own backends possible and desirable?

We might have to do this anyway, but we'll gladly contribute back.

fredszaq:
Hi there ! I have a project where I'm exposing a rust lib to java/kotlin using JNA, and I'm toying with the idea of adding support inside cbindgen as well. 

From what I see in the current code, the swich between the various supported languages is done via `if`s in the implementation of `Write` for the various types in the `ir` module. This approach worked well for C and C++ but is already showing its limits with Cython, as there are quite a few cases where the `Write` impls are filled with ` if config.language == Language::Cython {`

Adding a proper backend abstraction seems the way to go especially if we add C# and Java as supported language, maybe a trait with default impls called at relevant places in the `Write` impls and implemented for each language ?

What do you think @emilio, is even adding support for new languages something you would consider ? I'm happy to do this refactoring and adding java support if you're OK with it

emilio:
Yeah, I'd say a general back-end abstraction (probably an initial refactor would be to add `CLike` and `Cython` backends) would be a nice path forward.

fredszaq:
@yanchith @emilio  I just opened #853 that contains a refactoring adding language backend abstraction ! Would be glad to have you thoughts on this, this should be at a point where adding new languages should be doable

fredszaq:
And here is https://github.com/mozilla/cbindgen/pull/857 that contains the java backend, can be some inspiration to add the C# one

Tacodiva:
@fredszaq @yanchith Is somebody else working on the C# backend or are you looking for contributers? I'd be happy to give it a crack if you liked. I'm a bit of a Rust novice, but knew C# and it's interop stuff very well. I think I'd be able to do it but don't want to jump the gun if somebody else has already made progress on it or the language backend is too unstable for now.

fredszaq:
[...]

PR #853, opened 2023-06-30
Introduce language backend

fredszaq:
As discussed in #846 this introduces the notion of language backend in order to facilitate the addition of new supported languages to cbindgen

Notable changes: 
 - new `LanguageBackend` trait containing functions to write headers and footers in a source file 
 - 2 `LanguageBackend` impls : `CLikeLanguageBackend` and `CythonLanguageBackend`
 - the `Source` trait is now generic on a LanguageBackend and gets the language backend instead of a config as an argument ( both backend impls, end up wrapping a config)
 - The implementations of `Source` for the various structs of the `ir` have been moved to the corresponding backend module as there is now one implementation of `Source` per `LanguageBackend` for each relevant `ir` struct. 
 - The implementations of `Source` for both `LanguageBackend`s come from the existing impls and have been cleaned to contain only the parts relevant to the language of the Backend for which the impl is for

All tests continue to pass without any modification \o/ (those were quite useful during the refacto)


yanchith:
I am not familiar with the codebase, so I can't really review the patch itself. However, looking at `clike.rs` and `cython.rs`, this would give me a solid form to fill out to add the C# backend I need. Looking forward 👍

fredszaq:
@emilio Did you have time to take a look ? I'm planning to work on the java-jna backend this week, would be nice to have a confirmation we want to go in this direction ;)

src/bindgen/ir/generic_path.rs
```diff
@@ -94,12 +95,15 @@ impl GenericParams {
...
-    fn write_internal<F: Write>(
+    pub(crate) fn write_internal<F: Write, LB: LanguageBackend>(
         &self,
+        language_backend: &LB,
         config: &Config,
         out: &mut SourceWriter<F>,
         with_default: bool,
-    ) {
+    ) where
+        GenericArgument: Source<LB>,
```
emilio, line 105: So all these generic arguments are really unfortunate... :/

src/bindgen/bindings.rs
```diff
@@ -204,151 +200,46 @@ impl Bindings {
...
-            write!(out, "#include \"{}\"", include);
-            out.new_line();
-        }
-
-        if self.config.language == Language::Cython {
-            for (module, names) in &self.config.cython.cimports {
-                write!(out, "from {} cimport {}", module, names.join(", "));
-                out.new_line();
+            Language::Cython => {
+                self.write_with_backend(file, &CythonLanguageBackend::new(self.config.clone()))
```
emilio, line 209: Cloning the config here seems kind of unfortunate? Ideally the backend is an empty struct and the config is either given by callers or part of the source writer?

emilio (changes requested):
This looks generally reasonable, but I think if instead of relying on `Source<LB> for T` being implemented we had stuff like:

```
trait LanguageBackend {
  fn write_enum<W: Write>(&self, enum: &Enum, out: &mut SourceWriter<W>);
  fn ...;
}
```

etc, then all the generic soup goes away, right?

emilio:
Sorry for the lag getting to this btw :/

fredszaq:
Thanks for the review ! I think I tried the `write_xxx` functions in the language backend trait first but was blocked by something.  I cannot remember what though. I'll take a look when I'm back from holidays in september

src/bindgen/bindings.rs
```diff
@@ -204,151 +200,46 @@ impl Bindings {
...
+        Type: Source<LB>,
+        Documentation: Source<LB>,
+        Literal: Source<LB>,
+    {
         if self.noop {
             return;
         }
 
-        let mut out = SourceWriter::new(file, self);
+        let writer = Box::new(file);
```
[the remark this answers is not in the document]
fredszaq: was indeed not not needed, removed

src/bindgen/bindings.rs
fredszaq: I managed to avoid the clone by only passing a reference (we only need the config read only and the language backend is dropped after this call)

src/bindgen/ir/generic_path.rs
fredszaq: removed all those

fredszaq:
Hi @emilio ! I just rebased the branch and did the changes you asked, this effectively removes the source trait.

src/bindgen/bindings.rs
```diff
@@ -367,19 +247,19 @@ impl Bindings {
...
-                ItemContainer::Struct(ref x) => x.write(&self.config, &mut out),
-                ItemContainer::Union(ref x) => x.write(&self.config, &mut out),
-                ItemContainer::OpaqueItem(ref x) => x.write(&self.config, &mut out),
-                ItemContainer::Typedef(ref x) => x.write(&self.config, &mut out),
+                ItemContainer::Enum(ref x) => language_backend.write_enum(&mut out, x),
+                ItemContainer::Struct(ref x) => language_backend.write_struct(&mut out, x),
+                ItemContainer::Union(ref x) => language_backend.write_union(&mut out, x),
+                ItemContainer::OpaqueItem(ref x) => language_backend.write_opaque_item(&mut out, x),
+                ItemContainer::Typedef(ref x) => language_backend.write_type_def(&mut out, x),
             }
```
Tacodiva, line 255: I'm not too sure about the way you've abstracted the language backend. The problem is that each item is passed one at a time to the language backend. This works fine for simple languages where you can write any of these items anywhere and they only need to be in one place, but will cause problems for more complicated setups.

Consider a hypothetical language where in a source file, all the functions have to be defined before the structs. This abstraction would make writing a backend for that language very annoying, because the structs are always written first. Or, consider the situation that is making me write this in the first place; in my C# bindings for performance and flexibility of the generated file, I want to have a struct which contains all the function pointers dynamically loaded from the shared library. The problem is that in addition to the struct with the raw function pointers, I want to have nice and pretty C# methods which call the lower level function pointers (consumers of a native library in C# won't want to be dealing with function pointers). The way this code stands, this is unnecessarily difficult to do because each function needs to be in two places in the generated file, once as a field in the function pointer struct and second the method consumers will be using, but the abstraction expects me to just be able to write any function right after all the type definitions.

In my opinion, basically this whole `write_with_backend` function should be a part of the backend itself. Ideally, the backend would get passed a source writer, the configs and a struct containing a bunch of lists of all the enums, structs, unions, opaque items, type definitions, constants, globals and functions. Then, it's up to the backend to loop through these lists as is it pleases and can reason about the bindings as a whole rather than being drip-fed them one item at a time.

fredszaq: Having this code implemented inside the backend would mean :
  - duplicated code for most languages
  - harder implementation of new languages: right now implement the trait and you've a function that is called for every item that probably needs your attention this won't be the case if we move this reponsability t the backend

For your use case, what I would do is save the items you want to use and actually do the write in either `open_close_namespace(Close)`  or `write_footer()`. I can make the trait functions to take a `&mut self` instead of a `&self` in order to prevent having to have a `Mutex` to do that. Would that be a solution for you ?

Tacodiva: Yes, that is a solution. I would still maintain that the order of things within the output, and things like the astetic newlines this method is writing between the different types of things should be a responsibility of the backend. I believe I'd be nicer to keep the backends stateless, avoid copying data unnecessarily and avoid the weirdness of writing function data in the function which is supposed to be writing the footer, especially considering the price for all that flexibility is duplicating code to loop over the data. But, if you don't share that opinion than I will happily implement your suggested solution.

fredszaq: I have found a solution that is in fact way better: I just pushed changes that indeed creates a `write_bindings`  function in the `LanguageBackend` trait, but provides a default implementation for it. This way most languages should still only have to implement the abstract functions in the trait and have something that works but we can tweak the default in case  we need it. This allowed me to remove a few of the remaning `if`s testing agaist the language in the generic impl \o/. 

All functions in the trait now also take `&mut self` instead of `&self` this should help impl for languages that need to remember stuff between calls

ZhaoXiangXML:
Hello, is this MR going to be merged?

fredszaq:
Ping @emilio, I would love to see this merged (and #857) as well. Please tell be what I can do to facilitate this. All the remarks that have been raised till now have been addressed and I'll be pleased to make any other change deemed necessary.

kang-sw:
Oh Wow, this was exactly what I was looking for and trying to implement on my own from this night. God thanks I tried searching 'dynamic' on issue list and found this. Great appreciation for your effort and wish it'd be merged onto mainstream as soon as possible, thanks!

src/bindgen/mod.rs
```diff
@@ -46,6 +46,7 @@ mod declarationtyperesolver;
 mod dependencies;
 mod error;
 mod ir;
+mod language_backend;
```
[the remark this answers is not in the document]
kang-sw: Oh, nevermind. It seems then we have to expose bunch of another depdendencies such as `SourceWriter` etc to achive this ... seems not desirable. This PR already exposing several required properties of `Bindings` struct; therefore the user is able to write customized source generator already even without backend.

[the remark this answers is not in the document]
fredszaq: I just pushed a change removing the `pub` on `Bindings::write_with_backend`, no need having it in the public api as it is not usable

PR #856, opened 2023-07-19
Generate bindings for non-public extern items

### LINKED ISSUE #822: Generate bindings for private `extern` functions by default
Rust compiler [will always export](https://github.com/rust-lang/rust/issues/18134) private `extern` functions. We can say that visibility rules don't apply for `extern` functions. Considering this, generating C bindings for private `extern` functions should be opt-out. In the meantime, is there some way to opt-in generating bindings for private functions?

mversic:
Generate bindings for non-public rust items. 

We shouldn't be mixing concepts of visibility and exporting. Rust dynamic libraries always export `#[no_mangle]` items regardless of their visibility. More on this can be found in the issue description

Closes #822

### DIFF
--- a/src/bindgen/ir/global.rs
+++ b/src/bindgen/ir/global.rs
@@ -26,7 +26,11 @@ pub struct Static {
 }
 
 impl Static {
-    pub fn load(item: &syn::ItemStatic, mod_cfg: Option<&Cfg>) -> Result<Static, String> {
+    pub fn load(
+        path: Path,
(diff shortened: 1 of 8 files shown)


emilio (approved):
Looks reasonable, though I worry a bit about people using the pub check as a escape hatch.

emilio:
Merged with a warning fix in 5f235ec199f6a0b934f620fec0d2c481ba1d457a

PR #857, opened 2023-07-21
Add java support (using JNA)

fredszaq:
This builds on #853 and adds a new language backend capable of generating java bindings. 
    
The generated bindings use [JNA](https://github.com/java-native-access/jna)  (I used the last version 5.13.0 while developing but this will probably work with earlier ones as well)
    
Most of the new code lives in `bindgen/language_backend/java_jna.rs` and is pretty straightforward generation as most on the complicated things were done in the previous refactoring.
    
The test suite has been updated to generate and compile the bindings for java when running. On nice thing gained with this is the possibility to put a configuration file specific to a language alongside the standard one and the test will take this one instead of the default one if it exists (quite a few test use the `header` config to add some more code inside the generated bindings, C or preprocessor code doesn't work well in a java source file :P)
    
All tests are green, but that doesn't mean all the generated bindings are correct, some cases may fail at runtime as I didn't test all cases in a real program. I did however generate and use the bindings on a non trivial lib I'm developing and they worked as expected.


coolbluewater:
@fredszaq very cool! Currently I hand-author a Kotlin bridge class, e.g.:

```kotlin
package com.foo.bar
import android.view.Surface
class RustBridge {
    init { System.loadLibrary("mylibrary") }
    external fun enterFrame(rustObj: Long)
}
```
and use the following Rust code with the 'jni_name' attribute to ensure the name is correct:

```Rust
#[no_mangle]
#[jni_fn("com.foo.bar.RustBridge")]
pub fn enterFrame(_env: *mut JNIEnv, _: JClass, obj: jlong) {
    let obj = unsafe { &mut *(obj as *mut WgpuCanvas) };
    obj.enter_frame();
}
```

Will your backend be able to generate the kotlin bindings for this type? Also, will it handle the rust name conversion without needing the jni_fn crate? 
Thanks!

fredszaq:
Hi @coolbluewater  ! The generated bindings use [JNA](https://github.com/java-native-access/jna/). JNA works a bit differently than JNI (which your example uses) even though JNA uses JNI under the hood. 

When using JNI, the native code needs to be aware that it is called by a JVM. This is done for rust via the `#[jni_fn()]` proc macro. 

When using JNA, the native code doesn't need to be aware it is called from a JVM, as JNA is able to work with plain C compatible symbols. This means you don't have to have specific symbols for your lib for it to be able to be loaded from the jvm. 

This means the rust code would probably look something like that 

```
#[no_mangle]
pub fn enterFrame(obj: *mut WgpuCanvas) {
   unsafe{ obj.enterFrame}
}
```

and the koltin code would look like that 

```
class RustBridge {
    fun enterFrame(rustObj: MyLib.WgpuCanvasByReference) {
         MyLib.INSTANCE.entrerFrame(rustObj)
    }
}
```
With the code for the class `MyLib` beeing generated.

You can configure the name and the package of the generated class (actually the public api is an interface) in cbingen's config

```
[java_jna]
package = "my.package"
interface_name = "MyLib"
```

coolbluewater:
@fredszaq thanks for your response!
I need the JNIEnv and also need to pass `jobject`s from Kotlin to Rust.
Is this possible in some way?

fredszaq:
Not sure you want to mix JNI and JNA as the approaches are different and I'm not sure you'll be able to get access to the `JNIEnv` via JNA. 

Writing a  `java-jni` language backend that expects the rust code to be JNI aware could be a solution if you really need to access the `JNIEnv`/`jobject`s from the rust code. The the `java-jna` language backend in this PR kinda expects you are using plain old C compatible symbols.

fredszaq:
@emilio I rebased this with the latest master, following the merge of #942

src/bindgen/config.rs
```diff
@@ -42,18 +44,48 @@ impl FromStr for Language {
             "C" => Ok(Language::C),
             "cython" => Ok(Language::Cython),
             "Cython" => Ok(Language::Cython),
+            "Java-JNA" => Ok(Language::JavaJna),
+            "Java-Jna" => Ok(Language::JavaJna),
+            "java-jna" => Ok(Language::JavaJna),
+            "JavaJNA" => Ok(Language::JavaJna),
```
emilio, line 50: Let's just keep the dashed spelling?

src/bindgen/config.rs
```diff
@@ -42,18 +44,48 @@ impl FromStr for Language {
...
     }
 }
 
+impl Display for Language {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                Language::Cxx => {
```
emilio, line 64: nit: no need for braces

src/bindgen/config.rs
```diff
@@ -42,18 +44,48 @@ impl FromStr for Language {
...
+}
+
 deserialize_enum_str!(Language);
 
 impl Language {
     pub(crate) fn typedef(self) -> &'static str {
         match self {
             Language::Cxx | Language::C => "typedef",
             Language::Cython => "ctypedef",
+            _ => unimplemented!(),
```
emilio, line 88: Let's list JNA explicitly. But why can't it hit it?

src/bindgen/ir/enumeration.rs
```diff
@@ -721,6 +721,7 @@ impl Enum {
                     write!(out, "{}enum {}", config.style.cython_def(), tag_name);
                 }
             }
+            _ => unimplemented!(),
```
emilio, line 724: Same, let's at least keep the language listed. But why can't it be reached? Worth a comment.

src/bindgen/language_backend/java_jna.rs
```diff
@@ -0,0 +1,860 @@
...
+
+impl LanguageBackend for JavaJnaLanguageBackend<'_> {
+    fn open_namespaces<W: Write>(&mut self, out: &mut SourceWriter<W>) {
+        out.new_line_if_not_start();
+        let name = &self
+            .config
+            .java_jna
+            .interface_name
+            .clone()
+            .unwrap_or("Bindings".to_string());
```
emilio, line 34: no need to clone / allocate here, right?

Doesn't: `self.config.java_jna.interface_name.as_deref().unwrap_or("Bindings")` work?

src/bindgen/language_backend/java_jna.rs
```diff
@@ -0,0 +1,860 @@
...
+        }
+
+        if let Some(ref package) = self.config.java_jna.package {
+            out.new_line_if_not_start();
+            write!(out, "package {package};");
+            out.new_line();
+            out.new_line();
+        }
+
+        out.write("import com.sun.jna.*;");
```
emilio, line 101: Probably worth an `if !config.no_includes` around these blocks.

src/bindgen/language_backend/java_jna.rs
```diff
@@ -0,0 +1,860 @@
...
+    ty: &'a Type,
+    name: &'a Option<String>,
+    index: usize,
+}
+
+impl JavaJnaLanguageBackend<'_> {
+    fn write_deprecated<F: Write>(
+        &mut self,
+        out: &mut SourceWriter<F>,
+        deprecated: &Option<String>,
```
emilio, line 571: Let's use `Option<&str>` (and deprecated.as_deref() in the callers)

src/bindgen/language_backend/java_jna.rs
```diff
@@ -0,0 +1,860 @@
...
+                }
+                Type::Array(_, _) => out.write("Pointer"),
+                Type::FuncPtr { .. } => out.write("CallbackReference"),
+            },
+
+            Type::Path(path) => {
+                write!(out, "{}", path.export_name())
+            }
+            Type::Primitive(primitive) => {
+                //https://github.com/java-native-access/jna/blob/master/www/Mappings.md
```
emilio, line 427: space.

src/bindgen/language_backend/java_jna.rs
```diff
@@ -0,0 +1,860 @@
...
+                let typ = match primitive {
+                    PrimitiveType::Void => "void",
+                    PrimitiveType::Bool => "boolean",
```
emilio, line 430: Not right, this is 1 byte in rust, 4 in java.

src/bindgen/language_backend/java_jna.rs
```diff
@@ -0,0 +1,860 @@
...
+                    PrimitiveType::Double => "double",
+                    PrimitiveType::VaList => "Pointer",
+                    PrimitiveType::PtrDiffT => "Pointer",
+                    PrimitiveType::Integer { kind, .. } => {
+                        match kind {
+                            IntKind::Short => "short",
+                            IntKind::Int => "int",
+                            IntKind::Long => "NativeLong",
+                            IntKind::LongLong => "long",
+                            IntKind::SizeT => "NativeLong", // TODO probably not right
```
emilio, line 445: Def. not right. Should be "pointer" most likely.

src/bindgen/language_backend/java_jna.rs
```diff
@@ -0,0 +1,860 @@
...
+                            IntKind::Size => "NativeLong",  // TODO probably not right
+                            IntKind::B8 => "byte",
+                            IntKind::B16 => "short",
+                            IntKind::B32 => "int",
+                            IntKind::B64 => "long",
```
emilio, line 450: Also not right, this doesn't seem guaranteed to be 64 bytes.

src/bindgen/language_backend/java_jna.rs
```diff
@@ -0,0 +1,860 @@
...
+            IntKind::Short => JnaIntegerType::Short,
+            IntKind::Int => JnaIntegerType::Int,
+            IntKind::Long => JnaIntegerType::NativeLong,
+            IntKind::LongLong => JnaIntegerType::Long,
+            IntKind::SizeT => JnaIntegerType::SizeT,
+            IntKind::Size => JnaIntegerType::SizeT,
+            IntKind::B8 => JnaIntegerType::Byte,
+            IntKind::B16 => JnaIntegerType::Short,
+            IntKind::B32 => JnaIntegerType::Int,
+            IntKind::B64 => JnaIntegerType::Long,
```
emilio, line 546: Also not right.

src/bindgen/language_backend/java_jna.rs
```diff
@@ -0,0 +1,860 @@
...
+            JnaIntegerType::Long => "getLong(0)",
+        }
+    }
+
+    pub fn from_kind(kind: &IntKind) -> Self {
+        match kind {
+            IntKind::Short => JnaIntegerType::Short,
+            IntKind::Int => JnaIntegerType::Int,
+            IntKind::Long => JnaIntegerType::NativeLong,
+            IntKind::LongLong => JnaIntegerType::Long,
```
emilio, line 540: LongLong is not Long, for sure.

emilio (changes requested):

src/bindgen/config.rs
fredszaq: :+1:

src/bindgen/config.rs
fredszaq: done

src/bindgen/config.rs
fredszaq: this typedef function is only used by the enum code

src/bindgen/language_backend/java_jna.rs
fredszaq: :+1:

src/bindgen/language_backend/java_jna.rs
fredszaq: :+1:

src/bindgen/language_backend/java_jna.rs
fredszaq: fixed, now using a dedicated JNA IntegerType with two constants TRUE and FALSE

src/bindgen/language_backend/java_jna.rs
fredszaq: Fixed, now using a dedicated JNA IntegerType using size SIZE_T_SIZE

src/bindgen/language_backend/java_jna.rs
fredszaq: long is 64 bytes in java

src/bindgen/language_backend/java_jna.rs
fredszaq: Long is 64 bytes in java (the C long is NativeLong)

ZhaoXiangXML:
I'm looking forward for this one to be merged. I'm working on a C# backend that has some common ground with Java.

fredszaq:
Hi @emilio, I just rebased this on the latest master as there was some conflicts. It would be great if you have time to review again (I know this is a big piece). Regading length of various java types, a good documentation can be found here https://docs.oracle.com/javase/tutorial/java/nutsandbolts/datatypes.html 
We've been using bindings generated by this for a while at work now and didn't have any problems with them

Larkooo:
It would be great if we could have typed enums support

Larkooo:
This is very rough: https://github.com/Larkooo/cbindgen/tree/java-jna-backend but this should handle typed enums with unions and decorates them with a tag which is the original generated enum

PR #859, opened 2023-07-22
Some fix for new contributers.

sevenc-nanashi:
(My mother language is not English, so I may say some inappropriate things.)
I tried to contribute project, but there were some errors. I fixed them.

- This project requires nightly rust (as test uses `-Z`). Added rust-toolchain.toml.
- If proc-macro2 is older than 1.0.60, the build will fail. (cf: https://github.com/rust-lang/rust/issues/113152#issuecomment-1612580132) Changed the dependency section.
- The tests fails with Cython 3. Added some notes in contributing.md.

<details>
<summary>Cython 3 detail</summary>

As cython 3 prints this warning, the snapshot doesn't match.
```
Error compiling Cython file:
------------------------------------------------------------
...
from libc.stdint cimport int8_t, int16_t,

int32_t, int64_t, intptr_t
^
------------------------------------------------------------

tests/expectations/workspace.tag.pyx:1:0: Dotted filenames ('workspace.tag.pyx') are deprecated. Please use the normal Python package directory layout.
```

</details>

PR #860, opened 2023-07-23
Add: Support #[deprecated] attribute

### LINKED ISSUE #408: Translate #[deprecated] on Rust functions to deprecated attribute
Hello!  Thanks for making and maintaining `cbindgen`, it's very useful!

We have a C API and would like to mark some of our older functions as deprecated, in Rust this can be done with the [deprecated](https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html) attribute.  There is also a [deprecated](https://gcc.gnu.org/onlinedocs/gcc-4.7.0/gcc/Function-Attributes.html) [attribute](https://clang.llvm.org/docs/LanguageExtensions.html#messages-on-deprecated-and-unavailable-attributes) [in](https://docs.microsoft.com/en-us/cpp/preprocessor/deprecated-c-cpp?view=vs-2019) C.  Unfortunately it's not standardized and it appears that some compilers don't support it.

It seems like for this to be more portable, we'd have to generate some preprocessor logic.

Another nice to have feature would be to be able to put all the deprecated functions in a separate header file so that the user has to opt-in to use them, too.  The progression for us would then be, mark them as deprecated, move them to a separate `deprecated.h`, and then remove them.

What are your thoughts on deprecation with cbindgen?  Maybe there's a better solution!

sevenc-nanashi:
(My mother language is not English, so I may say some inappropriate things.)
This PR adds support of `#[deprecated]` attribute like must_use.
It supports:
- `#[deprecate]`,
- `#[deprecate = "note"]`,
- `#[deprecate(note = "note")]`
for:
- function
- enum
- struct

This PR closes #408. Also, I'm new to Rust, so I should have written some ineffective code. I'd like some reviews. Thank you.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 serde_json = "1.0"
 tempfile = "3"
 toml = "0.5"
-proc-macro2 = "1"
+proc-macro2 = "1.0.60"
 quote = "1"
 heck = "0.4"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ keywords = ["bindings", "ffi", "code-generation"]
 categories = ["external-ffi-bindings", "development-tools::ffi"]
 repository = "https://github.com/eqrion/cbindgen"
 edition = "2018"
-rust-version = "1.57"
+rust-version = "1.64"
 exclude = [
   "tests/profile.rs", # Test relies in a sub-crate, see https://github.com/rust-lang/cargo/issues/9017
 ]
--- a/src/bindgen/ir/annotation.rs
+++ b/src/bindgen/ir/annotation.rs
@@ -55,12 +55,12 @@ impl AnnotationSet {
         self.must_use && config.language != Language::Cython
     }
 
-    pub(crate) fn deprecated_note(&self, config: &Config) -> Option<String> {
+    pub(crate) fn deprecated_note(&self, config: &Config) -> Option<&str> {
         if config.language == Language::Cython {
             return None;
         }
 
-        self.deprecated.clone()
+        self.deprecated.as_deref()
     }
 
     pub(crate) fn format_deprecated_note(
--- a/src/bindgen/ir/annotation.rs
+++ b/src/bindgen/ir/annotation.rs
@@ -55,12 +55,25 @@ impl AnnotationSet {
         self.must_use && config.language != Language::Cython
     }
 
-    pub(crate) fn deprecated_node(&self, config: &Config) -> Option<String> {
+    pub(crate) fn deprecated_note(&self, config: &Config) -> Option<String> {
         if config.language == Language::Cython {
             return None;
         }
 
         self.deprecated.clone()
+    }
+
+    pub(crate) fn format_deprecated_note(
+        &self,
+        format_without_note: Option<&str>,
+        format_with_note: Option<&str>,
+        note: &str,
+    ) -> Option<String> {
+        if note.is_empty() {
+            return format_without_note.map(|x| x.to_string());
+        }
+        format_with_note
+            .map(|format_with_note| format_with_note.replace("{}", format!("{:?}", note).as_str()))
     }
 
     pub fn load(attrs: &[syn::Attribute]) -> Result<AnnotationSet, String> {
--- a/src/bindgen/ir/annotation.rs
+++ b/src/bindgen/ir/annotation.rs
@@ -55,6 +55,10 @@ impl AnnotationSet {
         self.must_use && config.language != Language::Cython
     }
 
+    pub(crate) fn deprecated(&self, config: &Config) -> bool {
+        self.deprecated.is_some() && config.language != Language::Cython
+    }
+
     pub fn load(attrs: &[syn::Attribute]) -> Result<AnnotationSet, String> {
         let lines = attrs.get_comment_lines();
         let lines: Vec<&str> = lines
--- a/src/bindgen/ir/annotation.rs
+++ b/src/bindgen/ir/annotation.rs
@@ -70,10 +70,36 @@ impl AnnotationSet {
             .collect();
 
         let must_use = attrs.has_attr_word("must_use");
-        if attrs.has_attr_word("deprecated") {
-            dbg!(attrs);
-            todo!();
-        }
+        let deprecated = if let Some(note) = attrs.attr_name_value_lookup("deprecated") {
+            Some(note)
+        } else if attrs.has_attr_word("deprecated") {
+            Some("".to_string())
+        } else if let Some(attr) = attrs.iter().find(|attr| {
+            if let Ok(syn::Meta::List(list)) = attr.parse_meta() {
+                list.path.is_ident("deprecated")
+            } else {
+                false
+            }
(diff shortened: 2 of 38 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ serde = { version = "1.0.103", default-features = false, features = ["derive"] }
 serde_json = "1.0"
 tempfile = "3"
 toml = "0.5"
-proc-macro2 = "1"
+proc-macro2 = "1.0.60"
```
sevenc-nanashi, line 29: ref: #859

src/bindgen/utilities.rs
```diff
@@ -293,3 +293,12 @@ fn split_doc_attr(input: &str) -> Vec<String> {
         .map(|s| s.trim_end().to_string())
         .collect()
 }
+
+pub fn create_deprecate_attribute(deprecated: &String) -> String {
+    if deprecated.is_empty() {
+        "[[deprecated]]".to_string()
+    } else {
+        // FIXME: I used {:?} to escape the string, but there should be much better way
```
sevenc-nanashi, line 301: Are there some function to escape string already?

src/bindgen/ir/annotation.rs
```diff
@@ -53,6 +55,10 @@ impl AnnotationSet {
         self.must_use && config.language != Language::Cython
     }
 
+    pub(crate) fn deprecated(&self, config: &Config) -> bool {
```
emilio, line 58: let's make this `fn deprecated_node(&self, config: &Config) -> Option<&str>`. That allows the `unwrap()`s elsewhere.

src/bindgen/ir/annotation.rs
```diff
@@ -68,6 +74,36 @@ impl AnnotationSet {
             .collect();
 
         let must_use = attrs.has_attr_word("must_use");
+        let deprecated = if let Some(note) = attrs.attr_name_value_lookup("deprecated") {
```
emilio, line 77: can we factor this out into its own function? That way you can also early-return and tidy it up a little bit.

src/bindgen/ir/annotation.rs
```diff
@@ -68,6 +74,36 @@ impl AnnotationSet {
...
+            Some(note)
+        } else if attrs.has_attr_word("deprecated") {
+            Some("".to_string())
+        } else if let Some(attr) = attrs.iter().find(|attr| {
```
emilio, line 81: Is this code-path tested?

src/bindgen/ir/enumeration.rs
```diff
@@ -787,6 +816,21 @@ impl Enum {
                     }
                 }
 
+                if self.annotations.deprecated(config) {
```
emilio, line 819: This is copy-pasted in three different places, let's not?

src/bindgen/ir/enumeration.rs
```diff
@@ -865,6 +909,17 @@ impl Enum {
...
         }
 
+        if self.annotations.deprecated(config) {
+            let note = self.annotations.deprecated.as_ref().unwrap();
+            if note.is_empty() {
+                if let Some(ref anno) = config.structure.deprecated {
+                    write!(out, " {}", anno);
+                }
+            } else if let Some(ref anno) = config.structure.deprecated_with_note {
+                write!(out, " {}", anno);
```
emilio, line 919: This doesn't use the note?

emilio (changes requested):
Looks generally reasonable, thanks. Some comments below

src/bindgen/ir/annotation.rs
sevenc-nanashi: It should be.

src/bindgen/utilities.rs
```diff
@@ -116,6 +116,47 @@ pub trait SynAttributeHelpers {
             })
     }
 
+    fn find_deprecated_note(&self) -> Result<Option<String>, String> {
```
sevenc-nanashi, line 119: I have no idea around where to put these utilities.

src/bindgen/ir/annotation.rs
```diff
@@ -55,14 +55,27 @@ impl AnnotationSet {
...
-    pub(crate) fn deprecated_node(&self, config: &Config) -> Option<String> {
+    pub(crate) fn deprecated_note(&self, config: &Config) -> Option<String> {
         if config.language == Language::Cython {
             return None;
         }
 
         self.deprecated.clone()
     }
 
+    pub(crate) fn format_deprecated_note(
```
sevenc-nanashi, line 66: I have no idea around where to put these utilities.

sevenc-nanashi:
Review applied. Thank you for your review

src/bindgen/ir/annotation.rs
```diff
@@ -53,6 +55,27 @@ impl AnnotationSet {
         self.must_use && config.language != Language::Cython
     }
 
+    pub(crate) fn deprecated_note(&self, config: &Config) -> Option<String> {
```
emilio, line 58: Why cloning? Can't you return an `Option<&str>`?

src/bindgen/ir/annotation.rs
```diff
@@ -53,6 +55,27 @@ impl AnnotationSet {
...
+        if config.language == Language::Cython {
+            return None;
+        }
+
+        self.deprecated.clone()
```
emilio, line 63: So `self.deprecated.as_deref()` here.

src/bindgen/utilities.rs
emilio: This one looks great, thanks! But rather than aborting parsing the item when we fail to parse `#[deprecated]`, I'd rather make this return an `Option<String>`, and return `None` for the error cases (maybe `warn!` ing if you want).

src/bindgen/ir/function.rs
```diff
@@ -284,6 +293,21 @@ impl Source for Function {
                         out.new_line();
                     }
                 }
+                if let Some(note) = func.annotations.deprecated_note(config) {
+                    if note.is_empty() {
+                        if let Some(ref anno) = config.function.deprecated {
+                            write!(out, "{}", anno);
+                            out.new_line();
+                        }
+                    } else if let Some(ref anno) = config.function.deprecated_with_note {
```
emilio, line 302: Any reason this is not using the helper?

emilio (changes requested):

src/bindgen/ir/function.rs
sevenc-nanashi: No, made it use the helper.

src/bindgen/ir/annotation.rs
sevenc-nanashi: I can, I made it return Option<&str> instead.

sevenc-nanashi:
> This one looks great, thanks! But rather than aborting parsing the item when we fail to parse #[deprecated], I'd rather make this return an Option<String>, and return None for the error cases (maybe warn! ing if you want).

(I couldn't reply somehow, so replying here): Thanks, the function returns `Option<String>` and `warn!` for error cases..

emilio:
I rebased your changes since we had some test renaming recently, and cleaned-up a bit. Thanks for the patch! I realized I left a pending comment below, feel free to ignore.

PR #863, opened 2023-07-31
Allow `=` and multiline prefix/postfix in annotations

novafacing:
This adds two things:

1. `=` is allowed in annotations. Previously, it was an error to have an `=` in the annotation but I'm doing hacky stuff (see below) and it didn't seem like there was a good reason not to permit it.
2. Multiline annotations are now supported by using a `\` at the end of the annotated line and any following lines that should be concatenated together with newlines.

What this allows is basically:

```rust
#[cfg(target_arch = "i386")]
pub mod foo {
  #[no_mangle]
  /// cbindgen:prefix= \
  /// #define FUNC_THAT_NEEDS_TO_BE_IFDEFED(x) \ \
  ///     do { \ \
  ///         do_something(x); \ \
  ///    } while (0) \
  ///
  pub extern "C" fn bar() {}
}
```

`after_include` is useful, but it has trouble when you need to conditionally compile based on the `ifdef` generated for a `cfg` attribute. This fixes that and makes it possible to somewhat hackily embed arbitrary C code wherever you want (at the cost of having to declare *something*). Eventually an annotation that can occur on a free comment might be nice to eliminate this need, but the two changes added by this PR, the `=` allowing and multiline allowing would still be needed.

### DIFF
--- a/src/bindgen/ir/annotation.rs
+++ b/src/bindgen/ir/annotation.rs
@@ -54,16 +54,17 @@ impl AnnotationSet {
     }
 
     pub fn load(attrs: &[syn::Attribute]) -> Result<AnnotationSet, String> {
-        let lines = attrs.get_comment_lines();
-        let lines: Vec<&str> = lines
+        let attrs_lines = attrs.get_comment_lines();
+        let lines: Vec<(usize, &str)> = attrs_lines
             .iter()
-            .filter_map(|line| {
+            .enumerate()
+            .filter_map(|(i, line)| {
                 let line = line.trim_start();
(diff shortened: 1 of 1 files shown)


novafacing:
Whoops, looks like I failed the MSRV check. Will fix that.

novafacing:
Hmm, the clippy error is a false positive...I'll see what I can do.

emilio:
At the very least this needs a test. Not opposed to this in principle tho?

novafacing:
Yeah it's breaking a lot of existing comments, I want to rework it and use some less silly parsing!

PR #867, opened 2023-08-11
Support for custom languages

LucaCiucci:
I wanted to generate bindings for Fortran, but I think this is not possible with the current version of `cbindgen`.
I added the `CustomLanguage` trait and the corresponding `Language::Custom` variant, also exposed some necessary types.  
An example usage can be found [here](https://github.com/LucaCiucci/call-rust-from-fortran/blob/3f38ba25445dca1a0ffe3ddbad27f5263b61679c/rust/build/build.rs#L38).


fredszaq:
Hi @LucaCiucci you may wand to take a look at https://github.com/mozilla/cbindgen/pull/853 where I've reworked quite a few things to make the addition of new languages easier.

An example of addtion of a new language is available in https://github.com/mozilla/cbindgen/pull/857 where I add support for java (using the jna lib), the interesting part is here https://github.com/mozilla/cbindgen/blob/d8b3cfae7a197109b4fa1bcf97372db8e7d49f4e/src/bindgen/language_backend/java_jna.rs

PR #868, opened 2023-08-25
Fix CI

jschwe:
Fix the following things that broke CI.

- Commit 667de09279c3d5f0524216fefa949e64cbd3bc1a broke CI by adding rust-toolchain.toml, which changed the default rust version to nightly. Explicitly specify the rust version in CI to use the correct toolchain.
- Revert upgrade to clap 4, since it causes the "Test package" test to fail.
- Fix a new clippy warning

### DIFF
--- a/.github/workflows/cbindgen.yml
+++ b/.github/workflows/cbindgen.yml
@@ -23,11 +23,11 @@ jobs:
 
     - name: Run rustfmt
       run: |
-        cargo fmt --check
+        cargo +stable fmt --check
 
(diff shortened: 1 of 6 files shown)


jschwe:
@emilio This PR fixes CI failures introduced by the recent merges. In the case of clap, I propose reverting for now until the author has time to investigate why the tests are failing after upgrading clap.

emilio:
Yeah, I agree. I was fixing the failures locally, but this is better to get 0.25 out.

PR #884, opened 2023-09-10
Add support for out-of-line bitfields declarations

glandium:


src/bindgen/parser.rs
```diff
@@ -560,6 +563,25 @@ impl Parse {
...
+                        .for_items_mut(&Path::new(type_name.unraw().to_string()), |item| {
+                            item.annotations.add_default(
+                                "internal-derive-bitflags",
+                                AnnotationValue::Bool(true),
+                            );
+                        });
+                }
+            }
+            self.load_syn_assoc_consts_from_impl(crate_name, mod_cfg, &impl_)
+        }
```
emilio, line 583: Why do you need to handle these at the end rather than inside `load_builtin_macro`? Is this to handle potentially out of order declaration vs. `impl` block?

I'd rather leave all the bitflags logic inside `load_builtin_macro`. Tests still pass with:

```
diff --git a/src/bindgen/parser.rs b/src/bindgen/parser.rs
index 924c846..3e32ec5 100644
--- a/src/bindgen/parser.rs
+++ b/src/bindgen/parser.rs
@@ -483,7 +483,6 @@ impl Parse {
     ) -> Vec<&'a syn::ItemMod> {
         let mut impls_with_assoc_consts = Vec::new();
         let mut nested_modules = Vec::new();
-        let mut bitflags_items = Vec::new();
 
         for item in items {
             if item.should_skip_parsing() {
@@ -548,9 +547,7 @@ impl Parse {
                     }
                 }
                 syn::Item::Macro(ref item) => {
-                    if let Some(bitflags) = self.load_builtin_macro(config, item) {
-                        bitflags_items.push(bitflags);
-                    }
+                    self.load_builtin_macro(config, crate_name, mod_cfg, item);
                 }
                 syn::Item::Mod(ref item) => {
                     nested_modules.push(item);
@@ -563,25 +560,6 @@ impl Parse {
             self.load_syn_assoc_consts_from_impl(crate_name, mod_cfg, item_impl)
         }
 
-        for bitflags in bitflags_items {
-            let (struct_, impl_) = bitflags.expand();
-            if let Some(struct_) = struct_ {
-                self.load_syn_struct(config, crate_name, mod_cfg, &struct_);
-            }
-            if let syn::Type::Path(ref path) = *impl_.self_ty {
-                if let Some(type_name) = path.path.get_ident() {
-                    self.structs
-                        .for_items_mut(&Path::new(type_name.unraw().to_string()), |item| {
-                            item.annotations.add_default(
-                                "internal-derive-bitflags",
-                                AnnotationValue::Bool(true),
-                            );
-                        });
-                }
-            }
-            self.load_syn_assoc_consts_from_impl(crate_name, mod_cfg, &impl_)
-        }
-
         nested_modules
     }
 
@@ -994,21 +972,41 @@ impl Parse {
     fn load_builtin_macro(
         &mut self,
         config: &Config,
+        crate_name: &str,
+        mod_cfg: Option<&Cfg>,
         item: &syn::ItemMacro,
-    ) -> Option<bitflags::Bitflags> {
+    ) {
         let name = match item.mac.path.segments.last() {
             Some(n) => n.ident.unraw().to_string(),
-            None => return None,
+            None => return,
         };
 
         if name != "bitflags" || !config.macro_expansion.bitflags {
-            return None;
+            return;
         }
 
-        bitflags::parse(item.mac.tokens.clone())
-            .map_err(|e| {
+        let bitflags = match bitflags::parse(item.mac.tokens.clone()) {
+            Ok(bf) => bf,
+            Err(e) => {
                 warn!("Failed to parse bitflags invocation: {:?}", e);
-            })
-            .ok()
+                return;
+            },
+        };
+
+        let (struct_, impl_) = bitflags.expand();
+        if let Some(struct_) = struct_ {
+            self.load_syn_struct(config, crate_name, mod_cfg, &struct_);
+        }
+        if let syn::Type::Path(ref path) = *impl_.self_ty {
+            if let Some(type_name) = path.path.get_ident() {
+                self.structs.for_items_mut(&Path::new(type_name.unraw().to_string()), |item| {
+                    item.annotations.add_default(
+                        "internal-derive-bitflags",
+                        AnnotationValue::Bool(true),
+                    );
+                });
+            }
+        }
+        self.load_syn_assoc_consts_from_impl(crate_name, mod_cfg, &impl_)
     }
 }
```

If you found this is important to handle, let's add a test for it. But I think it's slightly preferable just leaving the `load_builtin_macro` call doing all the work.

emilio (approved):
r=me, but we need either a test that needs the load_builtin_macro stuff, or just that simplification I pasted below. Mild preference for the later.

src/bindgen/parser.rs
glandium: > Is this to handle potentially out of order declaration vs. impl block?

Yes. Although I guess it's unlikely enough to not care too much about it.

PR #894, opened 2023-10-29
Support converting Rust `<name>: ...` to  C `...` (varargs)

4lDO2:
Fixes https://github.com/mozilla/cbindgen/issues/891.

I made this configurable, but I'm not sure if that's correct. Should `...` be the default, or only accepted translation, given Rust supports both `va: ...` and `va: VaList` for those two purposes?


emilio (changes requested):
If you could simplify this to always generate the same code I think that'd be preferable. Otherwise let's update the docs to reflect the new options.

emilio:
Superseded by #970

Issue #900, opened 2023-11-14
cbindgen ignores version/don't parse depedency flag when selecting crates to expand which can make it unusable in certain situations


IceTDrinker:
Hello,

We are currently exploring using the https://github.com/dtolnay/semver-trick to be able to give old crates the ability to convert some serialized data to more modern formats.

The basics are as follow (taking our https://github.com/zama-ai/tfhe-rs project as example):

- we have released 0.4.0 with a certain serialization format
- we plan on releasing 0.5.0 with breaking changes to the serialization
- we want to provide a way for users to update their serialization format for old data

The plan for the semver trick is:
- release 0.5.0
- update the 0.4 code base to be able to convert the old format to the new format and release 0.4.1

We have C APIs we need to be able to expose so we want to run cbindgen.

Currently cbindgen parses the cargo metadata ignoring the version of the package for which the binding is being generated

https://github.com/mozilla/cbindgen/blob/6bfc2176187a6fc6fba6315323b0296112330294/src/bindgen/cargo/cargo.rs#L185

In the current prototype we randomly expand the 0.4.1 crate (which we want, it is the current workspace crate being worked, it's the old one that will expose the forward compatible conversion functions for serialization) but we also sometimes expand the 0.5.0 (as it appears as a dependency and cbindgen does not discriminate against it) which causes issues with a CargoExpand error indicating we can't specify features for dependencies. We have disabled parsing for dependencies but the flag is not considered either (it would be fine for us if it took it into account)

I believe it's possible to support the semver-trick edge case (it is very much an edge case I get that perfectly), I'm willing to contribute the patch if necessary or collaborate/answer questions on that particular thing.

I'm sure we can hack up something on our end but I think it would be better if it's a thing cbindgen supports, I don't know if cargo metadata track dependencies, but if it does I would say that honoring the "do not expand dependencies" during the metadata parsing would be an acceptable solution and maybe would avoid issues in situations where two crates may accidentaly share a name in local dev environments (though in that case the easy fix is not naming different stuff with the same name)

IceTDrinker:
Could be easier and may "just" need to check the manifest_path

mversic:
> We have disabled parsing for dependencies but the flag is not considered either (it would be fine for us if it took it into account)

what happens in this case? what flag isn't considered?

IceTDrinker:
We set parse_deps to false, but as the bug is happening in the way candidate crates are derived from the cargo metadata I believe it has no discernable effect

[...]

PR #901, opened 2023-11-15
feat: allow users to specify a crate version for bindings generation

IceTDrinker:
- this is useful in a setting where several crates may share the same name in e.g. a semver trick setting https://github.com/dtolnay/semver-trick to allow adding compatibility code to older versions of a crate to allow users to migrate towards newer version of a crate
- this is a pure addition and should keep all existing behaviors exactly the same

this is a PR that attemps to solve https://github.com/mozilla/cbindgen/issues/900 in a way that does not break anything that already works with cbindgen as this is a pure addition to the code base leaving all existing behaviors untouched (other solutions may involve canonicalizing paths which may panic in places where there were no panics and may cause issues for any build systems with symlinks and network disks).

This is akin to cargo's package spec e.g. `cargo build -p my_crate@0.1.0` to allow disambiguating which crate the command is referring to in cases where several crates either share the same name or several version of the same crates happen to be in the dependency/cargo metadata for a given project. The good news is that the package version is populated by Cargo in the environment when in a build system so it is easy to use for build script users. For binary users a quick parsing of their Cargo.toml with theri favorite Unix utils should be enough

To avoid changing flags a separate flag is added to the cbindgen binary and a with_crate_version is added to the cbindgen Builder to be able to use that in a build script.

@emilio as indicated in the contributing.md I am requesting a code review for this PR, I'm not quite sure how to write a test case for this right now, if you have some pointers? Maybe a separate crate_version file akin to what the profile test file might be doing ?

Cheers

### DIFF
--- a/src/bindgen/cargo/cargo.rs
+++ b/src/bindgen/cargo/cargo.rs
@@ -87,7 +87,7 @@ impl Cargo {
     }
 
     pub(crate) fn binding_crate_ref(&self) -> PackageRef {
-        match self.find_pkg_ref(&self.binding_crate_name) {
+        match self.find_pkg_to_generate_bindings_ref(&self.binding_crate_name) {
             Some(pkg_ref) => pkg_ref,
             None => panic!(
                 "Unable to find {} for {:?}",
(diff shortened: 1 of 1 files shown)


IceTDrinker:
hello any news on that @emilio I guess (?) or @mversic sorry if I'm not pinging the right persons, but if this PR does not get a chance to progress we likely will have to publish a custom version of cbindgen and I'd prefer avoiding that

I'm afraid the deadline is too close anyways for us to hold out longer on this :|

mversic (approved):
LGTM. But I don't have maintainer privileges so I can't do anything

mversic:
but I also think this could be resolved without introducing another flag

IceTDrinker:
> but I also think this could be resolved without introducing another flag

Definitely agree, the thing is I have no idea how cbindgen might be used in the wild, not being a maintainer myself I chose a path of least resistance and safety with respect to the current behavior, adding a new flag and keeping the old default behavior was a way to guarantee things would work as before for people not aware/needing the flag.

Let me know what should be done here (not sure there is a « simple » way to derive a crate’s version apart from using the cargo env var about the version but that would make cbindgen depend on the cargo environment being there and set-up which I believe is not the case today)

IceTDrinker:
@mversic I think I have something which looks to be working with manifest paths, it probably can break things with people using relative paths so I added a fallback that has the old behavior of returning the first package that has been found, otherwise it means cbindgen needs to canonicalize some paths and this can result in panics that could never occur before

IceTDrinker:
Gentle ping @emilio it was suggested on Zulip I pinged you here on this PR for issue #900.

Recap:

cbindgen parses the cargo metadata based on name only, which fails randomly to select the proper crate for cbindgen if a dependency has the same name as the main crate (as the cargo metadata is in a hash table). 

I had a first implementation that was a pure addition, old code can still be found here: https://github.com/IceTDrinker/cbindgen/tree/saved/am/feat/specify-crate-version-v0

The first approach to a fix linked above is 100% backwards compatible by adding a new setting for the bindgen.

The current PR has a different approach validating the Cargo manifest path, issue here is I'm not sure it manages all edge cases with e.g. unresolved paths like relative paths. Canonicalizing paths could introduce panics that did not exist before, so I went for a middle ground which tries to validate the manifest path, and if it does not succeed falls back to the old behavior.

Would be happy to help or chat on the channel that is more convenient for you.

I'm unfortunately a bit time pressed and if the patch does not land in a release before January 19th we'll publish a crate which is cbindgen + that patch under a name that is distinct enough that it won't interfere with cbindgen on crates.io and will allow us to release

emilio (approved):
Would be good to get a test for this...

IceTDrinker:
> Would be good to get a test for this...

Fully agree, if you have pointers on how to write them I'll take it, I was not quite sure how the test system worked

PR #911, opened 2023-12-27
Fix `..Default::default()` for struct `Config`

### LINKED ISSUE #908: Config (still) has private fields in `0.26`
See #841 .

This is still broken in `0.26`. This makes `Builder::with_config()` a PITA to use.

```rust
    cbindgen::Builder::new()
        .with_crate(std::env::var("CARGO_MANIFEST_DIR").unwrap())
        .with_config(cbindgen::Config {
            language: cbindgen::Language::C,
            braces: cbindgen::Braces::SameLine,
            cpp_compat: true,
            style: cbindgen::Style::Both,
            ..Default::default()
        })
        .generate()?
        .write_to_file("binding.h");
```

```
error[E0451]: field `config_path` of struct `Config` is private
   --> foo/build.rs:104:15
    |
104 |             ..Default::default()
    |               ^^^^^^^^^^^^^^^^^^ field `config_path` is private
```

jschwe:
Previously the `depfile` option added a new private member to the `Config` struct - a breaking change which many users complained about. Making the field public, but hiding it from the documentation and excluding it from `serde` reverts the breakage, while keeping the depfile functionality.

I also considered adding an `InternalConfig` struct, but this would require breaking changes to at least `Config::from_file()` and `Config::from_root_or_default()`, which is a lot more invasive than my PR.

Closes #908

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,7 @@
+# unreleased
+
+      * Revert: The `Config` struct now has a private member.
+
 # 0.26.0
 
       * Fix swapping of `>>=` and `<<=` in constants.
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -1010,8 +1010,15 @@ pub struct Config {
     pub only_target_dependencies: bool,
     /// Configuration options specific to Cython.
     pub cython: CythonConfig,
+    #[doc(hidden)]
     #[serde(skip)]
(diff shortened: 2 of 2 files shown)


jschwe:
@emilio Any chance you could find some time to review this patch?

emilio (approved):
Thanks! Yeah, thanks for the ping, and sorry, I haven't been able to dedicate a lot of time to cbindgen recently.

PR #916, opened 2024-01-08
Add a way to generate a list of symbols for dynamic linkage

### LINKED ISSUE #907: Allow access to bindings symbols names
What I'm trying to do:
1. write a Rust library with a C-compatible interface (extern functions, etc)
2. use `cbindgen` to generate a C header, to use the library from C
3. compile a Rust program that depends on the library and links with `-C link-args=-Wl,--export-dynamic-symbol=<regex of symbols to export>` (or another [ldd option](https://man.archlinux.org/man/extra/lld/ld.lld.1.en)), in order to load plugins (written in C) that use the library symbols

For step 3, I need the list of symbols in the C header. I can parse the header or fill this information by hand, but `cbindgen` already has the list! What is missing is a way to access it.

I can see two ways of solving this problem (both may be useful):
1. Modify the `Bindings` impl: add a public API to read the `items` of the bindings (with methods to get the exported name of an item of any type), for instance:
```rs
let bindings = cbindgen::Builder::new().with_crate(crate_dir).with_language(C).generate().unwrap(); // usual generation
let symbols_names: Vec<String> = bindings.items.iter().map(|i| i.export_name()).collect(); // would be awesome if possible
```
2. Add a way to export the symbols in the [linker format](https://stackoverflow.com/a/70555660), which looks like
```
{
   symbol_1;
   symbol_2;
};
```
I will then be able to use `--export-dynamic-symbol-list=<exported file>` in order to export the symbols properly.

If someone guides me a little bit, I could try to implement one of the aforementioned features (or both, who knows?) :slightly_smiling_face:.

TheElectronWill:
This resolves #907 by adding the ability to generate the list of exportable symbols (i.e. functions and globals), in a format that the linker (ld) can understand. Like `depfile`, this new feature is available both as a CLI option (`--symfile`) and as a method in `Bindings`. Therefore, it's easy to use it in a build script.

Most of the new files are tests on the output of the new option `--symfile`.


src/bindgen/bindings.rs
```diff
@@ -131,6 +131,28 @@ impl Bindings {
...
+        res
+    }
+    
+    pub fn generate_symfile<P: AsRef<path::Path>>(&self, symfile_path: P) {
+        let symbols = self.dynamic_symbols_names();
+        if let Some(dir) = symfile_path.as_ref().parent() {
+            std::fs::create_dir_all(dir).unwrap();
+        }
+        let mut symfile = File::create(symfile_path).unwrap();
+        write!(&mut symfile, "{{\n{};\n}};", symbols.join(";\n")).expect("Writing symbol file failed");
```
jschwe, line 153:
```suggestion
        write!(&mut symfile, "{{\n{};\n}};\n", symbols.join(";\n")).expect("Writing symbol file failed");
```

Is a final newline possible?

jschwe (commented):
Nice - I might need exactly something like this in a project of mine in the next few months.

> in a format that the linker (ld) can understand.

Which linkers specifically support this option?  Is the `--dynamic-list` option of the gold linker compatible with the format, or did you only test with `ldd`? Is there any documentation on the exact format?

src/bindgen/bindings.rs
TheElectronWill: yes it is :)

TheElectronWill:
> Which linkers specifically support this option? Is the `--dynamic-list` option of the gold linker compatible with the format, or did you only test with `ldd`? Is there any documentation on the exact format?

The manual doesn't clearly explain the format. Gnu linker's manual reads ["The format of the dynamic list is the same as the version node without scope and node name. See VERSION"](https://linux.die.net/man/1/ld), and that "version" thing is introduced [here](https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html).

I would expect at least GNU ld, lld and gold to work with that (I have only tested ld for the moment, but I'm confident that [it works with lld](https://reviews.llvm.org/D18959)).

TheElectronWill:
rebased

emilio (commented):
Thanks, this looks good. Probably needs a rebase (can get to it if you can't) and I have one suggestion to avoid a couple intermediary allocations. But lmk if it's unreasonable.

src/bindgen/bindings.rs
```diff
@@ -131,6 +131,28 @@ impl Bindings {
         fields
     }
 
+    /// Lists the exported symbols that can be dynamically linked, i.e. globals and functions.
+    pub fn dynamic_symbols_names(&self) -> Vec<String> {
```
emilio, line 135: Probably:

```
fn dynamic_symbols_names(&self) -> impl Iterator<Item = &str>
```

And just chain the two things:

```
self.functions.iter().map(|f| f.path().name())
    .chain(self.globals.iter().map(|g| g.export_name()))
```

src/bindgen/bindings.rs
```diff
@@ -131,6 +131,28 @@ impl Bindings {
...
+        res
+    }
+    
+    pub fn generate_symfile<P: AsRef<path::Path>>(&self, symfile_path: P) {
+        let symbols = self.dynamic_symbols_names();
+        if let Some(dir) = symfile_path.as_ref().parent() {
+            std::fs::create_dir_all(dir).unwrap();
+        }
+        let mut symfile = File::create(symfile_path).unwrap();
+        write!(&mut symfile, "{{\n{};\n}};\n", symbols.join(";\n")).expect("Writing symbol file failed");
```
emilio, line 153: I'd probably write this iteratively using the above function rather than storing all the file contents in memory, but it's probably not a huge deal.

TheElectronWill: good idea, I've done that :)

TheElectronWill:
Sorry for the (very) late reply, it's been a busy year! I've rebased the branch and applied your suggestions.

TheElectronWill:
@emilio Is there something else that should be modified? This feature would be great for alumet-dev/alumet#58. Thanks for the review :)

emilio (approved):
No, looks good!

emilio:
Ugh, the merge queue is... not blocking on things passing, that really sucks. I'll fix up the CI tho, I don't think I'll need to revert this.

PR #924, opened 2024-02-01
Add no-export in cbindgen.toml config

tisonkun:
@aidanhs @emilio 

I need it for my c binding of [OpenDAL](https://github.com/apache/opendal) and local tests pass with this change.

Let me know if this patch needs some improvement :D

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -267,8 +267,6 @@ pub mod my_uninteresting_mod; // This won't be scanned by cbindgen.
 
 cbindgen will usually emit all items it finds, as instructed by the parse and export config sections. This annotation will make cbindgen skip this item from the output, while still being aware of it. This is useful for a) suppressing "Can't find" errors and b) emitting `struct my_struct` for types in a different header (rather than a bare `my_struct`).
 
-There is no equivalent config for this annotation - by comparison, the export exclude config will cause cbindgen to not be aware of the item at all.
-
 Note that cbindgen will still traverse `no-export` structs that are `repr(C)` to emit types present in the fields. You will need to manually exclude those types in your config if desired.
 
 ```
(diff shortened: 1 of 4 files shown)


src/bindgen/bindings.rs
```diff
@@ -363,6 +363,16 @@ impl Bindings {
                 continue;
             }
 
+            if self
+                .config
+                .export
+                .no_export
+                .iter()
+                .any(|i| i == item.deref().export_name())
```
tisonkun, line 371: I'm not very sure if `export_name` is desired here but it works for me locally.

src/bindgen/builder.rs
```diff
@@ -179,6 +179,15 @@ impl Builder {
         self
     }
 
+    #[allow(unused)]
```
tisonkun, line 182: I don't know how it works but follow the "include"/"exclude" pattern above.

emilio (changes requested):
Thanks, this is probably fine with better docs and a test.

docs.md
```diff
@@ -579,6 +577,10 @@ include = ["MyOrphanStruct", "MyGreatTypeRename"]
 # default: []
 exclude = ["Bad"]
 
+# A list of items to not export in the generated bindings
+# default: []
```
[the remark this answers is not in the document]
tisonkun: Updated.

tests/expectations/inner_no_export.compat.c
```diff
@@ -0,0 +1,20 @@
+#include <stdint.h>
+#include <stddef.h>
+#include <stdbool.h>
+
+/**
+ * Outer docs should be exported.
+ */
+typedef struct {
+  Inner *inner;
```
tisonkun, line 9: @emilio weird. When I use the same code for opendal c binding, it generates something like `struct Inner *inner` for such field. But in cbindgen tests it cannot generate the correct result.

emilio:
> It seems that for Style::Tag and Language::C, my work case can pass. But the test coverage also check for other style and language like pyx. The previous `no-export` attr should share the same issue.
> 
> @emilio what would you suggest to write a test? I cannot write a proper test for it but it's the same as when we previously merge "no-export` attr.

For tests, you can include relevant boilerplate using config.header or so to make the test files compile, right? We do that in a bunch of tests.

PR #926, opened 2024-02-09
Added support for integrating the package_version

hitbear:
The package version information appears in a comment of the header file. The new command line argument for this is --package-version. The TOML config file flag is package_version=true||false.

This new feature is helpful when you are working on a growing Rust library with new functions and structs. The package version entry in the header file appears as
**"/*Package version: {}*/",**

and gives a better overview which header file fits to which version of the library.


emilio (changes requested):
Looks fine, but needs a test, and probably undoing the rust-toolchain change. Let me know if you need any help with that.

rust-toolchain.toml
```diff
@@ -1,2 +0,0 @@
-[toolchain]
-channel = "nightly"
```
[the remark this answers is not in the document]
hitbear: Thank you for your response. Yes I fixed both requests but I think that I need some help to write an appropriate test.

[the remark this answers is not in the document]
emilio: You can add files to `tests/rust` where your toml file has the new option.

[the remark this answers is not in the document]
hitbear: Ok, in my newest commit I added a test crate package_version that checks whether the file is generated with the correct package version. 
Since some other tests failed with cython, I modified cbindgen. Now Comments such as header, trailer, etc. are written to the file after a check if the comment should be in C/C++ syntax (/*) or Python syntax ('''). 

`cargo test` now tells `ok. 145 passed;`

src/bindgen/bindings.rs
```diff
@@ -209,9 +212,30 @@ impl Bindings {
             return;
         }
 
+        if self.config.package_version {
+            out.new_line_if_not_start();
+            match self.config.language {
+                Language::C | Language::Cxx => {
+                    write!(out, "/* Package version: {} */", self.package_version,);
```
emilio, line 219: I'd remove the trailing comma here and below.

src/bindgen/bindings.rs
```diff
@@ -228,16 +252,35 @@ impl Bindings {
...
+                    );
+                }
+            }
+
             out.new_line();
         }
         if let Some(ref f) = self.config.autogen_warning {
             out.new_line_if_not_start();
-            write!(out, "{}", f);
+            match self.config.language {
```
emilio, line 276: I think we should keep these as it was, this contents are controllable by the user so we shouldn't try to mangle them.

src/bindgen/bindings.rs
```diff
@@ -209,9 +212,30 @@ impl Bindings {
...
+                }
+                Language::Cython => {
+                    write!(out, "''' Package version: {} '''", self.package_version,);
+                }
+            }
+
+            out.new_line();
+        }
+
         if let Some(ref f) = self.config.header {
```
emilio, line 229: Same here, this should probably remain as it was.

src/bindgen/bindings.rs
```diff
@@ -228,16 +252,35 @@ impl Bindings {
...
+                    write!(
+                        out,
+                        "/* Generated with cbindgen:{} */",
+                        crate::bindgen::config::VERSION
+                    );
+                }
+                Language::Cython => {
+                    write!(
+                        out,
+                        "''' Generated with cbindgen:{} '''",
```
emilio, line 266: This fix is obviously good.

emilio (changes requested):
Thanks! I think we should keep user inputs as-is and not mangle them, but the rest looks fine, thank you!

src/bindgen/bindings.rs
hitbear: So with these two commits, I removed the user input modification - and adjusted also the package_version test. Yes I see, it could be dangerous to change input.

tests/rust/package_version/cbindgen.toml
```diff
@@ -0,0 +1,7 @@
+package_version = true
+
+[parse.expand]
```
emilio, line 3: Do you need to use expand for this to work?

tests/rust/package_version/cbindgen.toml
```diff
@@ -0,0 +1,7 @@
...
+crates = ["package_version"]
+features = ["cbindgen"]
+
```
emilio, line 6: remove redundant newlines?

tests/rust/package_version/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#[repr(C)]
+pub struct Foo {
+    #[cfg(not(feature = "cbindgen"))]
```
emilio, line 3: This seems orthogonal to the feature you're testing?

hitbear: Yes, actually I could remove the whole stub code but in my latest commit, i removed just this line.

emilio (approved):

PR #928, opened 2024-02-15
Support specifying version for macro expansion.

alilleybrinker:
See: [#900].

Previously, cbindgen might sometimes match the wrong version of a crate
if the crate occurs multiple times in the dependency list produced by
`cargo metadata`. This meant that you'd observe transient errors where
_sometimes_ the right output would be produced (when the intended
version is macro-expanded), and sometimes it would not (when the wrong
version is macro-expanded).

This commit modifies the configuration to permit optionally specifying
name and version separately instead of solely specifying version.

This is an initial draft, as I have not yet been able to test it.

[#900]: https://github.com/mozilla/cbindgen/issues/900

Signed-off-by: Andrew Lilley Brinker <alilleybrinker@gmail.com>

### DIFF
--- a/src/bindgen/builder.rs
+++ b/src/bindgen/builder.rs
@@ -6,7 +6,7 @@ use std::path;
 
 use crate::bindgen::bindings::Bindings;
 use crate::bindgen::cargo::Cargo;
-use crate::bindgen::config::{Braces, Config, Language, Profile, Style};
+use crate::bindgen::config::{Braces, Config, CrateMatcher, Language, Profile, Style};
 use crate::bindgen::error::Error;
 use crate::bindgen::library::Library;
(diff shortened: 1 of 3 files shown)


alilleybrinker:
Some quick testing of mine showed that on initial run this patch did appear to cause just the correct version to be selected, but for some reason if run again it resulted in failure. I haven't yet been able to debug why that is. Suffice to say, this patch is not yet a sufficient solution.

alilleybrinker:
I'm actually going to close this out. I figured out a way to work around this issue for my own use case, and it doesn't make sense to further pursue work on this patch. The patch doesn't currently work.

That said, if someone else wants to take this and do something with it, feel free!

PR #950, opened 2024-05-07
Rename duplicate crate in workspace

gferon:
Otherwise, `cargo` prints a warning when using `cbindgen` as a `build-dependency` in my workspace:

```
warning: skipping duplicate package `single_crate` found at `/Users/gferon/.cargo/git/checkouts/cbindgen-4f7d248bb26fc43d/10f32b0/tests/depfile/single_crate`
warning: skipping duplicate package `single_crate` found at `/Users/gferon/.cargo/git/checkouts/cbindgen-4f7d248bb26fc43d/10f32b0/tests/depfile/single_crate_default_config`
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,3 +50,10 @@ required-features = ["clap"]
 [lib]
 name = "cbindgen"
 path = "src/lib.rs"
+
+[workspace]
+exclude = [
+  "tests/depfile/single_crate",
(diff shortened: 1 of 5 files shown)


emilio (approved):
hmm, ok, but we should try to exclude those tests crates somehow else...

gferon:
I added them to `workspace.exclude` and this seems to do the trick as well.

PR #956, opened 2024-05-17
Apply defines to struct literal

### LINKED ISSUE #955: `defines` is not applied for impl const
source (x.rs):
```rust
#[repr(C)]
pub struct Foo {
    a: i32,
    #[cfg(feature = "feat")]
    b: i32,
}

impl Foo {
    pub const ZEROED: Self = Self {
        a: 0,
        #[cfg(feature = "feat")]
        b: 0,
    };
}
```

cbindgen.toml:
```toml
[export]
include=["Foo"]

[defines]
"feature = feat" = "FEAT"
```

command:
```sh
cbindgen --config cbindgen.toml x.rs
```

actual result:
```c++
#include <cstdarg>
#include <cstdint>
#include <cstdlib>
#include <ostream>
#include <new>

struct Foo {
  int32_t a;
#if defined(FEAT)
  int32_t b
#endif
  ;
};
constexpr static const Foo Foo_ZEROED = Foo{ /* .a = */ 0, /* .b = */ 0 };
```

expected result:
```c++
#include <cstdarg>
#include <cstdint>
#include <cstdlib>
#include <ostream>
#include <new>

struct Foo {
  int32_t a;
#if defined(FEAT)
  int32_t b
#endif
  ;
};
constexpr static const Foo Foo_ZEROED = Foo{
  /* .a = */ 0,
#if defined(FEAT)
  /* .b = */ 0
#endif
};
```

youknowone:
Fix #955


tests/rust/cfg_field.rs
```diff
@@ -9,3 +9,18 @@ pub fn foo() {
...
+    pub y: u32,
+}
+
+impl Bar {
+    pub const BAR: Self = Self {
+        x: 0,
+        #[cfg(windows)]
+        y: 0,
+    };
+}
```
youknowone, line 26: How can I properly add a test? This change doesn't add diff

youknowone:
I didn't push newline/indent fix not to break test yet. Please let me know if the approach is okay. Then I will finish the whitespace work.

emilio (changes requested):
Thanks, looks good with a question.

Regarding the test, let's put it in tests/rust/cfg.rs.

Feel free to remove tests/rust/cfg_field.rs, it was meant as a crashtest at some point, looking at the git history, but right now it's covered by tests/rust/cfg.

src/bindgen/language_backend/clike.rs
```diff
@@ -913,23 +913,34 @@ impl LanguageBackend for CLikeLanguageBackend<'_> {
                 }
 
                 write!(out, "{{ ");
-                let mut is_first_field = true;
                 // In C++, same order as defined is required.
                 let ordered_fields = out.bindings().struct_field_names(path);
-                for ordered_key in ordered_fields.iter() {
+                let last_i = ordered_fields.len() - 1;
```
emilio, line 918: Ah, the question didn't get saved. It's basically... can't this underflow? It'd be better to do `if ordered_fields.is_empty() { 0 } else { ordered_fields.len() - 1 };`

tests/expectations/cfg.compat.c
```diff
@@ -95,6 +95,11 @@ typedef struct {
 #endif
   ;
 } ConditionalField;
+#define ConditionalField_ZERO (ConditionalField){ 
+#if defined(X11)
```
youknowone, line 99: @emilio Bad news here. This implementation does not cover C. Any advice to only enable the feature for C++ with sufficient features?

emilio (changes requested):
Ah, it might be harder than that... Recording with a bit whether we have conditional fields, and either skip generating the constant or generate it without the conditions for C seems feasible.

Alternatively doing some macro soup for C might work but it is annoying. Something like gathering the conditional fields first, and generating something like:

```
#ifdef X11
#define zero_(v) .zero = v,
#else
#define zero_(v)
#endif
#define ConditionalField_ZERO (ConditionalField){ \
  zero_(0) \
};
```

Or so, but that gets clunky.

src/bindgen/language_backend/clike.rs
```diff
@@ -918,26 +918,59 @@ impl LanguageBackend for CLikeLanguageBackend<'_> {
...
+                                condition.write_before(self.config, out);
+                                let define = format!("#define __{export_name}_{ordered_key}(v)");
+                                write!(out, "{define} (v)");
+                                write!(out, "\n#else\n");
+                                write!(out, "{define}");
+                                condition.write_after(self.config, out);
+                            }
+                        }
+                    }
+                }
```
youknowone, line 973: Where is the good place to put this?
From this place, it will write duplicated definitions for each literal definition

youknowone (commented):
Thank you for the advice. It had progress, but not yet finished.

I left another comment for duplicated field generation.
Another problem is constexpr context. `write_literal` needs context about `allow_constexpr` to choose macro or inline field. Do you have a suggestion for that?

src/bindgen/language_backend/clike.rs
```diff
@@ -911,30 +911,42 @@ impl LanguageBackend for CLikeLanguageBackend<'_> {
                 fields,
                 path,
             } => {
+                let is_constexpr = self.config.language == Language::Cxx
+                    && (self.config.constant.allow_constexpr || l.can_be_constexpr());
```
emilio, line 915: This seems wrong? Shouldn't it be `allow_constexpr && can_be_constexpr`?

src/bindgen/language_backend/clike.rs
```diff
@@ -911,30 +911,42 @@ impl LanguageBackend for CLikeLanguageBackend<'_> {
...
                     if let Some(lit) = fields.get(ordered_key) {
-                        if !is_first_field {
-                            write!(out, ", ");
-                        }
-                        is_first_field = false;
-                        if self.config.language == Language::Cxx {
+                        if is_constexpr {
+                            if i > 0 {
+                                write!(out, ", ");
+                            }
```
emilio: Factor this out of the if/else?

src/bindgen/language_backend/clike.rs
```diff
@@ -911,30 +911,42 @@ impl LanguageBackend for CLikeLanguageBackend<'_> {
...
+
+                            if self.config.language == Language::Cxx {
+                                // TODO: Some C++ versions (c++20?) now support designated
+                                // initializers, consider generating them.
+                                write!(out, "/* .{} = */ ", ordered_key);
+                            } else {
+                                write!(out, ".{} = ", ordered_key);
+                            }
+                            self.write_literal(out, lit);
```
emilio: Same here.

tests/expectations/cfg_both.c
```diff
@@ -77,6 +77,18 @@ typedef struct ConditionalField {
 #endif
   ;
 } ConditionalField;
+#define ConditionalField_ZERO (ConditionalField){ .field = __ConditionalField_field(0) }
+#if defined(X11)
+#define __ConditionalField_field(v) (v)
+#else
+#define __ConditionalField_field(v)
+#endif
+#define ConditionalField_ONE (ConditionalField){ .field = __ConditionalField_field(1) }
```
emilio, line 86: Does this even build when it's not defined? Doesn't it expand to `.field =`?

src/bindgen/language_backend/clike.rs
youknowone: Oh.. I was confused by allow_static_const. Thank you

tests/expectations/cfg_both.c
youknowone: Thanks, fixed it

src/bindgen/language_backend/clike.rs
youknowone: If you looked in first commit, this is changed in second commit

youknowone: If you don't mind, I separated the former 2 commits to another PR in #988, which is not directly related to cfg handling.

src/bindgen/language_backend/clike.rs
youknowone: @emilio This part is also not resolved yet

youknowone:
@emilio could you take another look?

youknowone:
Though #989 fixed the similar case to #955 for modern C++ using constexpr, this patch using `#define` is not finished yet.

I am not interested in finishing this PR anymore because #989 is enough for my case. Please feel free to take it if anyone is interested in it.

PR #957, opened 2024-05-17
Update tests.rs to use pretty_assertions for readable diffs

scovich:
The `pretty_assert` package makes it much (much) easier to compare expected vs. actual output when a test fails. Example:
image


emilio (approved):
Looks fine!

PR #958, opened 2024-05-17
Handle constants of nested transparent types

scovich:
When defining a constant of multiply-nested transparent struct type, the struct is reduced to typedefs and so the extra layers of constant initializer need to be removed. The existing code only removed one layer. Fixed by changing the check to a loop.

Without the fix, a declaration like this:
```rust
#[repr(transparent)]
struct TransparentStruct { field: u8 }

#[repr(transparent)]
struct Foo<T> { field: T }

#[repr(transparent)]
struct Bar<T> {
    field: Foo<T>,
}

pub const BAR: Bar<TransparentStruct> =
    Bar { field: Foo { field: TransparentStruct { field: 7 } } };
```

exports as 
```cpp
constexpr static const Bar<TransparentStruct> BAR =
    Foo { /* .field = */ TransparentStruct{ /* .field = */ 7 } };
```
instead of the desired
```cpp
constexpr static const Bar<TransparentStruct> BAR = 7;
```

### DIFF
--- a/src/bindgen/ir/constant.rs
+++ b/src/bindgen/ir/constant.rs
@@ -663,14 +663,13 @@ impl Constant {
             Cow::Owned(format!("{}_{}", associated_name, self.export_name()))
         };
 
-        let value = match self.value {
-            Literal::Struct {
-                ref fields,
-                ref path,
-                ..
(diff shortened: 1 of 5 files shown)


emilio (approved):
Thanks!

PR #961, opened 2024-05-20
deps: Update syn to 2.0

GrayJack:
The tests failed to me locally. Both without and with this modifications

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -76,7 +76,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serial_test",
- "syn 1.0.109",
+ "syn",
  "tempfile",
  "toml",
 ]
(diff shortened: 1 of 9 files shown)


GrayJack:
I took the changes of #957 to inspect better what the failures are.

The first one I'm looking at is the `layout.rs`, and I believe that the expected file is wrong, the comments say that `PackedStruct` and `PackedUnion` aren't translated, but they are in the expected file. And with this PR changes, they are not being translated, like the comment in the rust code says it should happen

emilio:
> The tests failed to me locally. Both without and with this modifications

Can you elaborate on how? CI looks green here tho, so that's always something :)

emilio:
I fixed / simplified some of the va-list handling (which otherwise breaks with #970).

emilio:
Thanks!

emilio:
Also if you could file an issue about the tests failing on your machine, it'd be great to get it sorted out.

GrayJack:
> Can you elaborate on how? CI looks green here tho, so that's always something :)

It was the first comment, I fixed the failures in the other commits and forgot to update the comment 😅

PR #963, opened 2024-05-22
Generalize Item to expose documentation and generic params

scovich:
Small PR to make `Item` more fully capture commonalities of the underlying types.

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -318,7 +318,7 @@ impl Enum {
     }
 
     pub fn add_monomorphs(&self, library: &Library, out: &mut Monomorphs) {
-        if self.generic_params.len() > 0 {
+        if self.is_generic() {
             return;
         }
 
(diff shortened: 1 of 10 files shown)


src/bindgen/ir/enumeration.rs
```diff
@@ -492,6 +496,10 @@ impl Item for Enum {
         }
     }
 
+    fn generic_params(&self) -> Option<&GenericParams> {
```
emilio, line 499: We could have a `const EMPTY: Self = Self(Vec::new());` in `GenericParams` to make this a bit simpler perhaps (remove the `Option`, make callers return `&GenericParams::EMPTY` instead)?

How does that sound?

src/bindgen/ir/item.rs
```diff
@@ -36,6 +37,13 @@ pub trait Item {
     fn resolve_declaration_types(&mut self, _resolver: &DeclarationTypeResolver) {
         unimplemented!()
     }
+    fn generic_params(&self) -> Option<&GenericParams>;
+
+    fn is_generic(&self) -> bool {
+        self.generic_params()
+            .is_some_and(|params| !params.is_empty())
```
emilio, line 44: Then this is just `self.generic_params().is_empty()`, for example.

emilio (approved):
Seems fine but I think it can be slightly simpler with that tweak. Mind applying it if you agree? If not, please reply and I can merge as-is.

emilio:
Made that suggested tweak and merged this, thanks!

PR #966, opened 2024-05-24
Add support for transparent typedefs

scovich:
Following https://github.com/mozilla/cbindgen/issues/967, introduce `/// cbindgen:transparent-typedef` annotation that causes cbindgen to replace transparent structs and typedefs with their underlying type.

This allows the following:
```rust
/// cbindgen:transparent-typedef
#[repr(transparent)]
struct Foo(NonNull<i32>);

/// cbindgen:transparent-typedef
type Function = extern "C" fn(i: i32) -> i64;

type NullableFoo = Option<Foo>;
type NullableFunction = Option<Function>;
```
to export as
```c
typedef i32 *NullableFoo;
typedef int64_t (*NullableFunction)(int32_t);
```
instead of today's output:
```c++
template<typename T = void>
struct Option;

using Foo = int32_t*;
using Function = int64_t(*)(int32_t i);

using NullableFoo = Option<Foo>;
using NullableFunction = Option<Function>;
```

### DIFF
--- a/src/bindgen/ir/ty.rs
+++ b/src/bindgen/ir/ty.rs
@@ -600,30 +600,24 @@ impl Type {
                 is_nullable,
                 never_return,
             } => {
-                // Predict that ret+args will all be erased, but decrement the count whenever we're
-                // wrong. If the count drops to 0, then type erasure was a no-op after all.
-                let mut num_erased = 1 + args.len();
-                let erased_ret = eraser
-                    .erase_transparent_types(library, ret, mappings)
-                    .unwrap_or_else(|| {
-                        num_erased -= 1;
-                        ret.as_ref().clone()
-                    });
-
+                // Attempt to erase ret and all args; if any of them were actually erased, then
+                // assemble and return the simplified function signature that results.
+                let mut erased_any = false;
+                let mut try_erase = |ty| {
+                    if let Some(erased) = eraser.erase_transparent_types(library, ty, mappings) {
+                        erased_any = true;
+                        erased
+                    } else {
+                        ty.clone()
+                    }
+                };
+                let erased_ret = try_erase(ret);
                 let erased_args = args
                     .iter()
-                    .map(|(name, ty)| {
-                        let erased_ty = eraser
-                            .erase_transparent_types(library, ty, mappings)
-                            .unwrap_or_else(|| {
-                                num_erased -= 1;
-                                ty.clone()
-                            });
-                        (name.clone(), erased_ty)
-                    })
+                    .map(|(name, ty)| (name.clone(), try_erase(ty)))
                     .collect();
 
-                if num_erased > 0 {
+                if erased_any {
                     return Some(Type::FuncPtr {
                         ret: Box::new(erased_ret),
                         args: erased_args,


docs.md
```diff
@@ -316,27 +385,25 @@ The rest are just local overrides for the same options found in the cbindgen.tom
...
 
-```
-#define MY_ATTRIBUTES [[nodiscard]]
-```
+  Combined with something like:
 
-for example.
+  ```
+  #define MY_ATTRIBUTES [[nodiscard]]
+  ```
```
emilio, line 406: These changes seem unrelated? Seems ok, but maybe worth splitting into its own commit / PR? Or is it some sort of autoformatting i'm not aware of?

src/bindgen/ir/ty.rs
```diff
@@ -519,59 +519,121 @@ impl Type {
...
+                }
+            }
+            Type::FuncPtr {
+                ref ret,
+                ref args,
+                is_nullable,
+                never_return,
+            } => {
+                // Predict that ret+args will all be erased, but decrement the count whenever we're
+                // wrong. If the count drops to 0, then type erasure was a no-op after all.
```
emilio, line 604: Could you elaborate on what this is trying to handle? An example would be great.

src/bindgen/ir/ty.rs
```diff
@@ -519,59 +519,121 @@ impl Type {
...
+                let mut num_erased = 1 + args.len();
```
emilio, line 605: I think it'd be a bit clearer to do it the other way around. In fact, wouldn't it be a bit nicer to just:

```
let mut erased_any = false;
macro_rules! try_erase {
    ($item:tt) => {
        if let Some(erased) = eraser.erase_transparent_types(library, $item, mappings) {
            erased
        } else {
            $item.clone()
        }
    }
```

Then:

```
let ret = try_erase!(ret);
...

if erased_any {
    return Some(..);
}
```

emilio (commented):
Need to take a closer look but over all looks reasonable.

src/bindgen/ir/ty.rs
emilio: Disregard this, this was about me not groking that the logic was basically ("try and find if we erased any")

docs.md
scovich: I was just fixing broken indentation in the existing .md, so that bullets would render correctly. If you hide whitespace changes the diff disappears:
image

Happy to split it out if it's too annoying, tho.

src/bindgen/ir/ty.rs
scovich: Ah, nice suggestion. I used a lambda instead of a macro tho.

scovich:
I just noticed that this PR overlaps significantly with https://github.com/mozilla/cbindgen/pull/699, and probably fixes both https://github.com/mozilla/cbindgen/issues/690 (`NonZero`) and https://github.com/mozilla/cbindgen/issues/326 (`Option`)?

src/bindgen/ir/generic_path.rs
```diff
@@ -234,6 +255,48 @@ impl GenericArgument {
     }
 }
 
+/// Helper for erasing transparent types, which memoizes already-seen types to avoid repeated work.
+#[derive(Default)]
+pub struct TransparentTypeEraser {
+    // Remember paths we've already visited, so we don't repeat unnecessary work.
+    // TODO: how to handle recursive types such as `struct Foo { next: Box<Foo> }`?
+    known_types: HashMap<Type, Option<Type>>,
```
emilio, line 263: Do we really expect extremely long chains of transparent typedefs? Memoizing this seems over-kill without performance numbers, IMHO...

src/bindgen/ir/ty.rs
```diff
@@ -519,59 +519,115 @@ impl Type {
...
+                let mut erased_any = false;
+                let mut try_erase = |ty| {
+                    if let Some(erased) = eraser.erase_transparent_types(library, ty, mappings) {
+                        erased_any = true;
+                        erased
+                    } else {
+                        ty.clone()
+                    }
+                };
+                let erased_ret = try_erase(ret);
```
emilio, line 614: A bit unfortunate to do all this cloning unconditionally just to ~almost always throw it away, fwiw...

src/bindgen/library.rs
```diff
@@ -172,6 +175,39 @@ impl Library {
...
 
+    fn erase_transparent_types_for_items<T: Item + Clone>(
+        &self,
+        eraser: &mut TransparentTypeEraser,
+        items: &ItemMap<T>,
+    ) -> ItemMap<T> {
+        // NOTE: Because `items` is actually a shared reference to `self`, we cannot take it as
+        // mutable. We also cannot `drain` or `take` it first, because then the items would be
+        // unavailable for lookup during the type erasure process. So we mutate a clone, and let the
+        // caller assign the result back after these shared references have died.
```
emilio, line 186: This also looks really expensive for big projects... I'd much rather find a way to avoid this unconditional clone of everything... That may mean rejiggering how we store the items, or at the very least changing how we do the pass.

It seems the only reason we need the library around is to do the `transparent_alias` stuff, correct?

If so, would it make sense to first do a pass over `self.opaque_items`, `self.typedefs`, and `self.structs`, in order to collect a `Path -> Type` map, and only pass _that_ to `erase_transparent_types`?

emilio (changes requested):
Sorry for the lag getting to this. I think this is a nice improvement, though I'm a bit concerned about performance tbh.

See some thoughts above.

src/bindgen/ir/generic_path.rs
scovich: I don't know about long chains, but I do expect frequent use (typedefs exist to eliminate repetition in typing out complex type names). Is frequent use not sufficient reason to memoize?

scovich:
@emilio -- thanks for pointing out the performance/memory hazards. It turns out this PR was also incomplete/incorrect in its handling of type parameters (specialization was not always correct). I made a second attempt from scratch in https://github.com/mozilla/cbindgen/pull/1029, which should address all of these issues. PTAL?

PR #978, opened 2024-06-24
Implement whitelist in ExportConfig

AnthonyYoManz:
Adds a field `whitelist` to `ExportConfig`, allowing users to specify an exclusive list of items they wish to include in the generated header file. If the field is left empty it is ignored. 

If excludes or includes are specified, they override the whitelist. For example in the case where we have functions `foo`, `bar`, and `baz` and our whitelist specifies `foo` and `bar`, then only `foo` and `bar` will be included in the output header file. If they add an include `baz`, all three functions will be included in the output. If they then add an exclude `foo`, only `bar` and `baz` will be included in the output.

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -596,6 +596,11 @@ include = ["MyOrphanStruct", "MyGreatTypeRename"]
 # default: []
 exclude = ["Bad"]
 
+# A list of items to include in the generated bindings, while excluding other unlisted
+# items from being exported. Has lower precedence than `include` and `exclude`, so they
+# can still add or remove items regardless of the whitelist's value.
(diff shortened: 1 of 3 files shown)


AnthonyYoManz:
Maybe I've missed it but I could not find where tests for other ExportConfig fields are, so I took the path of least resistance and simply wrote none for this - although I did test 'manually', writing a separate crate and making sure behaviour was as described in my PR description. With some guidance on where exactly to put such a test, I am happy to write it myself, although it might be a little awkward as I am on Windows and the entire test-suite very much does not function on this platform.

Issue #55, opened 2017-09-27
Support non-C calling conventions


Michael-F-Bryan:
I'm trying to export some Rust functions as a DLL to be used in a larger application, but instead of the usual `extern "C"` calling convention, I'm using `stdcall`. With `stdcall`, `cbindgen` doesn't emit declarations.

Would it be possible to tweak the rules so `cbindgen` will make bindings for anything marked `extern`  and `#[no_mangle]`, no matter the calling convention? I don't think there's any time you *wouldn't* want bindings for a function marked `extern`.

This looks like it's related to #49, but affects all forms of `extern`, not just where the `"C"` is omitted.

eqrion:
This seems reasonable to me.

What needs to be output for `fn extern "stdcall"`? Just `__stdcall` after the return type?

Michael-F-Bryan:
According to [msdn] the syntax looks like this: `return-type __stdcall function-name[(argument-list)]`. I just tried this out on my Arch Linux and while clang compiles a dummy program just fine, gcc spits out a syntax error. Apparently gcc wants you to write `__attribute__((cdecl))` instead. 

I haven't used the C pre-processor for much more than conditional compilation and `#include`, is there a non-hacky way to make this work across different compilers and platforms?

Alternatively, you could add an extra flag to the config struct which takes a `Option<CallingConvention>` (where `CallingConvention` is an enum of `GCC` or `MSVC`) with the default being `None`. This would mean you need to explicitly choose how the calling convention should be rendered in a header file. Then you can add a pass before rendering to emit warnings if a non-C calling convention is detected and the user hasn't specified a `CallingConvention`.

---

Awesome project by the way. I originally made something similar because `rusty-binder` looked inactive, it's nowhere near as complete as `cbindgen` though!

[msdn]: https://msdn.microsoft.com/en-us/library/zxk0tw93.aspx

eqrion:
Hmm, so I think we can generate the following to handle this:

```c++
#ifndef _MSC_VER
// might need some fudging depending on where the attribute needs to go in the function declaration.
#define __stdcall __attribute__((stdcall))
#endif
```

Seeing as `stdcall` is a supported function attribute for gcc (and possibly clang). [(1)](https://gcc.gnu.org/onlinedocs/gcc/x86-Function-Attributes.html#x86-Function-Attributes)

Michael-F-Bryan:
[...]

PR #979, opened 2024-06-28
Other calling conventions

novafacing:
This PR adds support for other calling conventions as discussed in #55. It adds an ABI field (instead of just an extern field) to `Function`s and adds a concept of `Predefines` which right now only contains the compiler/platform-specific `#defines` needed to make the calling conventions work nicely but may be useful in the future for other cases where something needs to look at all the code, then conditionally emit some lines at the top of the file.

### DIFF
--- a/src/bindgen/bindings.rs
+++ b/src/bindgen/bindings.rs
@@ -18,6 +18,7 @@ use crate::bindgen::ir::{
 use crate::bindgen::language_backend::{
     CLikeLanguageBackend, CythonLanguageBackend, LanguageBackend,
 };
+use crate::bindgen::predefines::Predefines;
 use crate::bindgen::writer::SourceWriter;
 
 /// A bindings header that can be written.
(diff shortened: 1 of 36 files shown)


novafacing:
Here's a godbolt with the generated headers (test cconv.rs): https://godbolt.org/z/MqPhhzqcf

There are a few compile errors which I intentionally left in (like requesting sysv64 on clang/gcc non-x86_64). I can change that behavior so a compile error is never emitted and a warning is instead if that's preferred.

novafacing:
Realized I had a misunderstanding about what the extern_decl flag was doing (toggling `extern` keyword in output). I put it back to how it was, so now that plays nice with the ABI detection and dispatching.

novafacing:
Documented the configuration option added for this feature. Should be ready to merge :)

novafacing:
Calling convention predefines are not emitted in a defined order.

PR #988, opened 2024-08-10
Newline for each field for constexpr field constants

youknowone:
generate newline for each constexpr literal structs' fields

A shared patch to branch #956 and #989


emilio (approved):
Ok, I reviewed this on the other PR. Only issue is this adds trailing whitespace for constexpr expressions which should be removed, but with that fixed, looks good.

emilio (approved):

emilio:
Thanks!

PR #991, opened 2024-08-12
Deduplicate the code that turns transparent structs into typedefs

scovich:
This is a "prefactor" for https://github.com/mozilla/cbindgen/pull/966, which centralizes the way cbindgen handles struct definitions that should be emitted as typedefs instead. For example, all backends emit `NonZero<T>` as a simple typedef, and the C backend emits `Option<T>` as a typedef as well. With this change, the code is generalized so that struct typedefs can be handled cleanly and centrally, instead of replicating code in each backend that needs it. It also opens a future path to handling transparent enums in the same way.

NOTE: With this change, cbindgen also handles zero-sized structs gracefully instead of crashing.

### DIFF
--- a/src/bindgen/ir/structure.rs
+++ b/src/bindgen/ir/structure.rs
@@ -122,11 +122,18 @@ impl Struct {
         has_tag_field: bool,
         is_enum_variant_body: bool,
         alignment: Option<ReprAlign>,
-        is_transparent: bool,
+        mut is_transparent: bool,
         cfg: Option<Cfg>,
         annotations: AnnotationSet,
         documentation: Documentation,
     ) -> Self {
+        // WARNING: The rust compiler rejects a `#[repr(transparent)]` struct with 2+ NZST fields, but
+        // accepts an empty struct (see https://github.com/rust-lang/rust/issues/129029). We must
+        // not emit a typedef in that case, because there is no underlying type.
+        if is_transparent && fields.len() != 1 {
+            error!("Illegal empty transparent struct {}", &path);
+            is_transparent = false;
+        }
         let export_name = path.name().to_owned();
         Self {
             path,
@@ -152,19 +159,10 @@ impl Struct {
 
     /// Attempts to convert this struct to a typedef (only works for transparent structs).
     pub fn as_typedef(&self) -> Option<Typedef> {
-        if self.is_transparent {
-            // NOTE: The rust compiler rejects a `#[repr(transparent)]` struct with 2+ NZST fields,
-            // but accepts an empty struct (see https://github.com/rust-lang/rust/issues/129029).
-            // Don't emit the typedef in that case.
-            if let Some(field) = self.fields.first() {
-                return Some(Typedef::new_from_struct_field(self, field));
-            }
-            error!(
-                "Cannot convert empty transparent struct {} to typedef",
-                self.name()
-            );
+        match self.fields.first() {
+            Some(field) if self.is_transparent => Some(Typedef::new_from_struct_field(self, field)),
+            _ => None,
         }
-        None
     }
 
     pub fn add_monomorphs(&self, library: &Library, out: &mut Monomorphs) {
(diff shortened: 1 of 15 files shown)


src/bindgen/ir/structure.rs
```diff
@@ -150,6 +150,23 @@ impl Struct {
...
     }
 
+    /// Attempts to convert this struct to a typedef (only works for transparent structs).
+    pub fn as_typedef(&self) -> Option<Typedef> {
+        if self.is_transparent {
+            // NOTE: A `#[repr(transparent)]` struct with 2+ NZT fields fails to compile, but 0
+            // fields is allowed for some strange reason. Don't emit the typedef in that case.
+            if let Some(field) = self.fields.first() {
+                return Some(Typedef::new_from_struct_field(self, field));
+            } else {
```
emilio, line 160: nit: no else after return, just de-indent the error!

src/bindgen/ir/structure.rs
```diff
@@ -150,6 +150,23 @@ impl Struct {
         }
     }
 
+    /// Attempts to convert this struct to a typedef (only works for transparent structs).
+    pub fn as_typedef(&self) -> Option<Typedef> {
+        if self.is_transparent {
+            // NOTE: A `#[repr(transparent)]` struct with 2+ NZT fields fails to compile, but 0
+            // fields is allowed for some strange reason. Don't emit the typedef in that case.
+            if let Some(field) = self.fields.first() {
```
emilio, line 158: Can you add a test for `#[repr(transparent)] struct Foo;` or so, so that we have test-coverage for the `None` branch?

emilio (approved):
Looks great, thanks! Just some minor questions

src/bindgen/ir/structure.rs
scovich: nice catch, thanks.

src/bindgen/ir/structure.rs
scovich: I tried, but it exposed a latent cbindgen bug for test cases that enable struct tagging:
```
Output failed to compile: Output {
  status: ExitStatus(unix_wait_status(256)),
  stdout: "",
  stderr: "
    /path/to/cbindgen/tests/expectations/transparent_tag.compat.c:43:11: 
    error: must use 'struct' tag to refer to type 'TransparentEmptyStructure'
    TransparentEmptyStructure h,
    ^
    struct
    1 error generated.
  "
}
```

If I run the same test against master, cbindgen itself fails:
```
cbindgen failed: Some("/path/to/cbindgen/tests/expectations/transparent.compat.c") with error: 
thread 'main panicked at src/bindgen/language_backend/clike.rs:548:34:
index out of bounds: the len is 0 but the index is 0
```
My newly added check avoided the index out of bounds panic, but I don't know where the tagging issue would lurk?

scovich: Also filed https://github.com/rust-lang/rust/issues/129029, because it seems like the rust compiler shouldn't allow an empty transparent struct in the first place.

scovich: Found it: [structure.rs:274](https://github.com/mozilla/cbindgen/blob/58c6156b0d91e82abb03c26187b8d18fa4345ce0/src/bindgen/ir/structure.rs#L274):
```diff
- if self.is_transparent {
+ if self.is_transparent && self.fields.len() == 1 {
```

scovich: The above fix is surgical, but I'm not sure it's ideal? Should we forbid `is_transparent` for an empty struct, rather than tolerate it like this code currently does? I worry the checks are proliferating...

scovich: I went ahead and moved the check into the constructor, so that transparent implies single field everywhere else.

scovich: Hmm. It's apparently intentional behavior to allow an empty transparent struct: https://github.com/rust-lang/rust/issues/77841#issuecomment-716575747. But I don't know what the underlying should be in that case, if we were to emit a typedef. Should we continue suppressing the typedef for empty transparent structs? Or is there some canonical type we should use as underlying for the typedef?

scovich: Update: All zero-sized structs now trigger a warning about undefined behavior, and an empty struct definition is emitted regardless of whether the user asked for repr(C) or repr(transparent).

emilio (approved):

emilio:
Squashed the patches, closed by 3ed9434f097b3a05b4ed6eedf80877d5d4b9ace9

PR #992, opened 2024-08-12
Add support for transparent enums

scovich:
The code restructuring work done for https://github.com/mozilla/cbindgen/pull/966 makes it straightforward to also support rust's transparent enums if we want to do so. Add the support, including and accompanying tests and documentation.


tests/rust/transparent.rs
```diff
@@ -8,66 +8,271 @@ struct TransparentComplexWrappingStructTuple(DummyStruct);
 #[repr(transparent)]
 struct TransparentPrimitiveWrappingStructTuple(u32);
 
-// Transparent structure wrapping a struct.
+// Transparent struct wrapping a struct.
 #[repr(transparent)]
-struct TransparentComplexWrappingStructure { only_field: DummyStruct }
+struct TransparentComplexWrappingStruct { only_field: DummyStruct }
```
youknowone, line 13: I like this patch in general, but these changes which are not related to the changed behavior makes a bit harder to read the patch.

PR #994, opened 2024-08-15
Generate flexible array members the c99 way instead of the GNU way

serge-sans-paille:
There a re three ways to express flexible array members:

1. char fam[0];
2. char fam[1];
3. char fam[];

3. is the only standard way (in c99, but supported in c++ as an extension), the other two are GNU syntax (still supported in clang though).

Let's generate the standard syntax by default, while leaving it possible to use 1. through struct.gnu_flexible_array_members

Cython only supports the 1. mode.


tests/rust/struct.rs
```diff
@@ -28,11 +28,18 @@ struct TupleRenamed(i32, f32);
 #[repr(C)]
 struct TupleNamed(i32, f32);
 
+#[repr(C)]
+struct WithFlexibleArrayMember {
+    x: i32,
+    y: [i8; 0],
```
emilio, line 34: Can you test with some other field afterwards too? Not sure what the behavior there should be, maybe it's fine.

serge-sans-paille: I had to rework the whole approach. I've added a specific type to represent flexible array members, and that makes everything much more consistent.
I'm unsure about the idiomaticity of my rust code, don't hesitate to bash me :-)

serge-sans-paille:
@emilio even if that change is correct from a compiler point of view, it would actually break mozilla code if gnu behavior is of by default because the firefox codebase is making an optimistic use of flexible array members. I guess it's still good to merge this, but with gnu behavior on by default. What do you think?

src/bindgen/cdecl.rs
```diff
@@ -286,10 +291,26 @@ impl CDecl {
                     if last_was_pointer {
                         out.write(")");
                     }
+
                     write!(out, "[{}]", constant);
 
                     last_was_pointer = false;
                 }
+                CDeclarator::FlexibleArray() => {
```
emilio, line 299: There's no need for `()`, just remove it from the enum declaration.

emilio (approved):
Yeah, seems fine to do this. Want to apply that nit? Otherwise I can.

emilio:
> @emilio even if that change is correct from a compiler point of view, it would actually break mozilla code if gnu behavior is of by default because the firefox codebase is making an optimistic use of flexible array members.

Can you elaborate? Do you have an example?

PR #1003, opened 2024-09-22
ir: add support for UnsafeCell and SyncUnsafeCell

### LINKED ISSUE #1002: Add support for UnsafeCell and SyncUnsafeCell
Support for `RefCell` has been added in #489, but `UnsafeCell` and `SyncUnsafeCell` are not supported.
Both are substantially equal to a `Cell`, so it could be implemented in the same way.

alekitto:
fixes #1002


emilio (approved):

emilio:
Thanks for the fix, and specially thanks for including a test!

PR #1006, opened 2024-10-04
Fix local override of enum prefix-with-name

### LINKED ISSUE #808: `cbindgen:prefix-with-name=false` comment does not override toml value `enum.prefix_with_name=true`
Example:

```rust
/// cbindgen:prefix-with-name=false
#[repr(C)]
pub enum Result<T> {
    Ok(T),
    Err(*mut Error), // Error isn't defined, doesn't matter for this case.
}

#[no_mangle]
pub extern "C" fn hello() -> Result<i32> {
    todo!()
}
```

cbindgen.toml:
```toml
language = "C"

[enum]
prefix_with_name = true
```

Output of `cbindgen .`:

```
WARN: Can't find Error. This usually means that this type was incompatible or not found.
#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>

typedef enum Result_i32_Tag {
  Result_i32_Ok_i32,
  Result_i32_Err_i32,
} Result_i32_Tag;

typedef struct Result_i32 {
  Result_i32_Tag tag;
  union {
    struct {
      int32_t ok;
    };
    struct {
      Error *err;
    };
  };
} Result_i32;

typedef struct Result_i32 I32Result;

I32Result hello(void);
```

Note that the `Result_i32_Tag` cases have been prefixed with `Result_i32_` despite the inline comment of `/// cbindgen:prefix-with-name=false`.

jsgf:
Fixes #808 using @emilio's suggestion in https://github.com/mozilla/cbindgen/issues/808#issuecomment-1460383144

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -534,8 +534,10 @@ impl Item for Enum {
             }
         }
 
-        if config.enumeration.prefix_with_name
-            || self.annotations.bool("prefix-with-name").unwrap_or(false)
+        if self
+            .annotations
+            .bool("prefix-with-name")
+            .unwrap_or(config.enumeration.prefix_with_name)
         {
             let separator = if config.export.mangle.remove_underscores {
                 ""


emilio (approved):

emilio:
Looks good tho a test would be nice. Given it's trivial and I don't have the cycles to do it myself right now, and I've already taken too long to get to this, this is fine.

PR #1007, opened 2024-10-04
Add rename-all=prefix:

jsgf:
This adds a new rename-all annotation which allows setting a specific prefix to all fields. For example:

```rust
/// cbindgen:rename-all=prefix:ERR_
#[repr(i32)]
pub enum Error {
   BAD = 0,
   VERY_BAD = 1
}
```

While in principle this can be added to the config file, it's primarily useful for overrides on a case-by-case basis.

### DIFF
--- a/src/bindgen/ir/union.rs
+++ b/src/bindgen/ir/union.rs
@@ -171,7 +171,7 @@ impl Item for Union {
         let rules = self
             .annotations
             .parse_atom::<RenameRule>("rename-all")
-            .unwrap_or(config.structure.rename_fields);
+            .unwrap_or(config.structure.rename_fields.clone());
 
         if let Some(o) = self.annotations.list("field-names") {
             let mut overriden_fields = Vec::new();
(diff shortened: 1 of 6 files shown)


jsgf:
Alternatively (and perhaps preferably) this could be a separate option which is applied after the normal rename rule so you can both rename and add a prefix (which is exclusive with prefix_with_name for enums).

src/bindgen/ir/union.rs
```diff
@@ -171,7 +171,7 @@ impl Item for Union {
         let rules = self
             .annotations
             .parse_atom::<RenameRule>("rename-all")
-            .unwrap_or(config.structure.rename_fields);
+            .unwrap_or(config.structure.rename_fields.clone());
```
emilio, line 174: Maybe avoid these clones by:

```rust
let rules = self.annotations.parse_atom::<>();
let rules = rules.as_ref().unwrap_or(&config.structure.rename_fields);
```

But fine either way.

emilio (approved):
This is fine for now :)

PR #1011, opened 2024-10-11
Add `unstable_ir` feature flag that makes the ir pub

### LINKED ISSUE #1009: Make `ir` module public
This would allow users to easily customize the bindings by inspecting the bindings items.

Is there any particular reason why most of the modules are private?  
If the concern is exposing too much complexity, we could just make `ir` and other modules public with a feature flag, such as `"internals"`.

AFAIK the only walk-around at the moment is to use a modified version of this crate.

heesooy:
Fixes #1009

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ pretty_assertions = "1.4.0"
 
 [features]
 default =  ["clap"]
+unstable_ir = []
 
 [[bin]]
 name = "cbindgen"
--- a/docs.md
+++ b/docs.md
@@ -94,9 +94,18 @@ cbindgen = "0.24.0"
 
 If you'd like to use a `build.rs` script with a `cbindgen.toml`, consider using [`cbindgen::generate()`](https://docs.rs/cbindgen/*/cbindgen/fn.generate.html) instead.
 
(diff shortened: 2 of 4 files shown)


zhp46:
This would be a fantastic addition to have

emilio (changes requested):
I guess this is reasonable to do, but I don't want to make any change to the ir a breaking change.

Let's call this `unstable_ir` and document it with no stability guarantees in docs.md (that is, if you use this, you need to pin the library, and are on your own).

With that, this looks reasonable to me, if it simplifies life downstream compared to a fork...

heesooy:
> I guess this is reasonable to do, but I don't want to make any change to the ir a breaking change.
> 
> Let's call this `unstable_ir` and document it with no stability guarantees in docs.md (that is, if you use this, you need to pin the library, and are on your own).
> 
> With that, this looks reasonable to me, if it simplifies life downstream compared to a fork...

@emilio -- done; renamed to `unstable_ir` and added a section in the docs. Let me know if you want me to move the docs to a different section, wasn't sure where would be best. Thanks for taking a look!

emilio (approved):

PR #1019, opened 2024-10-27
Parse `unsafe` attributes

### LINKED ISSUE #1013: Parse `unsafe` in attributes
See https://blog.rust-lang.org/2024/10/17/Rust-1.82.0.html#unsafe-attributes

Now attributes can contain `unsafe`, but cbindgen gives the following error:

```txt
ParseSyntaxError { crate_name: "...", src_path: "...", error: Error("expected identifier, found keyword `unsafe`") }
```

I think this issue could be urgent (should be done before edition 2024), because programmers, like me, may write:

```rs
#[unsafe(export_name = "xxx")]
pub fn f() {

}
```

Current rust stable (1.82.x) introduces this syntax and recommends us to add `unsafe`, while the old form (without `unsafe`) is still acceptable. But the old form will be a **hard error** in edition 2024.

---

This issue seems actually from upstream: https://github.com/dtolnay/syn/issues/1710

Coekjan:
Closes #1013 . The linked issue is urgent, so @emilio (hope we can resolve the issue as soon as possible).

The dependency `syn` is updated in order to handle the `unsafe` attributes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -405,9 +405,9 @@ checksum = "7da8b5736845d9f2fcb837ea5d9e2628564b3b043a70948a3f0b778838c5fb4f"
 
 [[package]]
 name = "syn"
-version = "2.0.72"
+version = "2.0.85"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


emilio (approved):
Thanks, this looks good to me. There's some minor nits and an extra toml change that needs to happen in order for CI to be happy. I sent #1020 with those.

PR #1024, opened 2024-11-01
Explicitly instantiate function return values that are template classes

scovich:
Following the ideas discussed in https://github.com/mozilla/cbindgen/issues/402, search the set of referenced types for templates that are function return values and emit them all as fields of a dummy struct. This silences compiler warnings (clang/gcc) and errors (MSVC) about potential ABI compatibility issues caused by returning template classes by value from `extern "C"` functions. We take the dummy struct approach because explicit instantiation changes semantics of C++ templates and can cause linker errors if multiple compilation units do it -- not good for a general header file.

This behavior is implemented as a new pass (similar to the existing `instantiate_monomorphs` pass) that is controlled by a new export config, `instantiate_return_value_monomorphs`. The name of the dummy struct defaults to `__cbindgen_monomorph_struct` but can be overridden by a second config.

### DIFF
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -341,12 +341,14 @@ pub struct ExportConfig {
     pub renaming_overrides_prefixing: bool,
     /// Mangling configuration.
     pub mangle: MangleConfig,
-    /// Whether to instantiate the struct templates that were emitted, as members of a giant struct
-    /// with the name optionally overridden by [`instantiate_monomorphs_struct_name`]. This silences
-    /// warnings and errors for several compiles, notably MSVC.
-    pub instantiate_monomorphs: bool,
+    /// Whether to instantiate the monomorphs of template types used as function return values. This
+    /// is needed for C compatibility, because otherwise compilers warn (gcc/clang) or even reject
+    /// (MSVC) those function definitions. The compensation is made by emitting a single struct with
+    /// one field for each monomorphized type. The emitted wrapper struct's name can optionally be
+    /// overridden by [`return_value_monomorphs_struct_name`].
+    pub instantiate_return_value_monomorphs: bool,
     /// The struct name to use when [`instantiate_monomorphs`] is enabled, ignored otherwise.
-    pub instantiate_monomorphs_struct_name: Option<String>,
+    pub return_value_monomorphs_struct_name: Option<String>,
 }
 
 /// Mangling-specific configuration.
--- a/src/bindgen/config.rs
+++ b/src/bindgen/config.rs
@@ -341,6 +341,12 @@ pub struct ExportConfig {
     pub renaming_overrides_prefixing: bool,
     /// Mangling configuration.
     pub mangle: MangleConfig,
+    /// Whether to instantiate the struct templates that were emitted, as members of a giant struct
+    /// with the name optionally overridden by [`instantiate_monomorphs_struct_name`]. This silences
+    /// warnings and errors for several compiles, notably MSVC.
+    pub instantiate_monomorphs: bool,
+    /// The struct name to use when [`instantiate_monomorphs`] is enabled, ignored otherwise.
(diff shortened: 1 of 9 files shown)


tests/tests.rs
```diff
@@ -144,9 +144,6 @@ fn compile(
             command.arg("-Wno-attributes");
             // clang warns about unused const variables.
             command.arg("-Wno-unused-const-variable");
-            // clang also warns about returning non-instantiated templates (they could
-            // be specialized, but they're not so it's fine).
-            command.arg("-Wno-return-type-c-linkage");
```
scovich, line 149: No longer needed -- All tests that used to trigger this warning have been updated to use the new feature.

tests/rust/swift_name.rs
```diff
@@ -1,3 +1,9 @@
+// A stand-in for Box<T>, since Box<T> is undefined behavior in C++
+#[repr(transparent)]
+pub struct Foo<T> {
```
scovich, line 3: There's no way to compile C++ code that uses (opaque) `Box<T>`. At least, not without additional hacks that would trigger undefined behavior (such as manually providing a definition for the class). 

This new feature ignores opaque items, which causes the test to fail compilation with `-Wreturn-type-c-linkage -Werror` warnings. If I change the feature to emit opaque items, compilation fails because the type lacks a definition.

I don't believe this test was specifically trying to test `Box` handling in the first place (and probably not even C++), so I updated it to use a transparent struct that will have the same behavior as `Box` in C code.

Or, we could just add the `.skip_warning_as_error` suffix to this test, if that's preferred.

tests/tests.rs
```diff
@@ -189,7 +186,21 @@ fn compile(
...
+        };
+        let stderr = match str::from_utf8(&out.stderr) {
+            Ok(s) => s.to_string(),
+            Err(_) => format!("{:?}", out.stderr),
+        };
+        println!("Output failed to compile: {:?}", out.status);
+        println!("=== STDOUT ===\n{}", stdout);
+        println!("=== STDERR ===\n{}", stderr);
+        println!("==============");
+        assert!(out.status.success());
```
scovich, line 202: This change wasn't strictly necessary, but it sure made debugging broken tests easier

scovich:
@emilio -- would it be possible to get a review of this PR?

src/bindgen/config.rs
```diff
@@ -341,6 +341,15 @@ pub struct ExportConfig {
     pub renaming_overrides_prefixing: bool,
     /// Mangling configuration.
     pub mangle: MangleConfig,
+    /// Whether to instantiate the monomorphs of template types used as function return values. This
+    /// is needed for C compatibility, because otherwise compilers warn (`-Wreturn-type-c-linkage`
+    /// on gcc/clang) or even reject (MSVC) those function definitions. The compensation is made by
+    /// emitting a single struct with one field for each monomorphized type. The emitted wrapper
+    /// struct's name can optionally be overridden by [`return_value_monomorphs_struct_name`].
+    pub instantiate_return_value_monomorphs: bool,
```
emilio, line 349: FWIW I think this is kind of a workaround... But yeah it's not pretty. I don't think you need a struct or so tho, you can just explicitly instantiate them, right?

See [here](https://searchfox.org/mozilla-central/rev/f571db8014431de31d245017e2f5457046aec4ea/layout/style/ServoStyleConstsInlines.h#33) for example.

emilio (commented):
Thanks, I think this is worth fixing, but mind clarifying on why you went for a wrapper struct rather than just explicit instantiation?

That doesn't hurt I guess, but it is a bit weird when you can just instantiate it. You could even output it at the function level without an extra global pass (though that'd be a bit noisy perhaps? Seems fine if you track the paths you've declared).

src/bindgen/config.rs
scovich: (also updated the PR description with some of this info, since it wasn't clear before)

PR #1028, opened 2024-11-06
Support forward declarations for C-like languages

### LINKED ISSUE #43: Generate forward declarations for structs with cyclical pointer references

masnagam:
Fixes #43


src/bindgen/ir/ty.rs
```diff
@@ -687,15 +688,54 @@ impl Type {
         generic_params: &GenericParams,
         library: &Library,
         out: &mut Dependencies,
+        ptr_types: &mut IndexSet<GenericPath>,
+
```
emilio, line 692: nit: space. But also rather than having an `is_ptr` param, it seems this should be handled in `Type::Ptr` itself?

emilio (changes requested):
So... I'm not sure why we need a separate set around, it seems generally this should be tracked as part of `Dependencies`?

Also, I don't think we should unconditionally create forward declarations for non-problematic cases as you're doing here (anything that is a pointer gets forward-declared).

masnagam:
> So... I'm not sure why we need a separate set around, it seems generally this should be tracked as part of `Dependencies`?
> 
> Also, I don't think we should unconditionally create forward declarations for non-problematic cases as you're doing here (anything that is a pointer gets forward-declared).

i'll check your points and modify this PR when i have time.

PR #1029, opened 2024-11-07
Add support for transparent typedefs (take 2)

scovich:
Supersedes https://github.com/mozilla/cbindgen/pull/966

Following https://github.com/mozilla/cbindgen/issues/967, introduce `/// cbindgen:transparent-typedef` annotation that causes cbindgen to replace transparent structs and typedefs with their underlying type.

This allows the following:
```rust
/// cbindgen:transparent-typedef
#[repr(transparent)]
struct Foo(NonNull<i32>);

/// cbindgen:transparent-typedef
type Function = extern "C" fn(i: i32) -> i64;

type NullableFoo = Option<Foo>;
type NullableFunction = Option<Function>;
```
to export as
```c
typedef i32 *NullableFoo;
typedef int64_t (*NullableFunction)(int32_t);
```
instead of today's output:
```c++
template<typename T = void>
struct Option;

using Foo = int32_t*;
using Function = int64_t(*)(int32_t i);

using NullableFoo = Option<Foo>;
using NullableFunction = Option<Function>;
```

### DIFF
--- a/docs.md
+++ b/docs.md
@@ -299,7 +299,76 @@ fn bar() -> Foo { .. } // Will be emitted as `struct foo bar();`
 
 ### Struct Annotations
 
-* field-names=\[field1, field2, ...\] -- sets the names of all the fields in the output struct. These names will be output verbatim, and are not eligible for renaming.
+* field-names=\[field1, field2, ...\] -- sets the names of all the fields in the output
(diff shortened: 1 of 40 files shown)


scovich:
@emilio -- would it be possible to get a review of this PR?

PR #1034, opened 2024-12-12
Add rename rule for generated associated constant's base name

danielocfb:
In FFI Rust code it is often necessary to use constants instead of enums for compatibility reasons. The most natural thing to do to somewhat preserve grouping is to use associated constants. E.g.:

  ```rust
  #[repr(C)]
  struct Foo {}

  impl Foo {
      pub const FLAG1: u32 = 10;
      pub const FLAG2: u32 = 11;
      ...
  ```

For the above, the generated constants would be called `Foo_FLAG1` & `Foo_FLAG2`. Many Linux core C libraries adhere to a slightly different casing, however, where the type prefix is fully upper cased: `FOO_FLAG1`. Currently, it does not seem possible to generate such constants.

This change introduces a new config option to structured types that enables this use case: the `struct.associated_constants_rename_base_name` option expects a rename rule (as other rename related config options) and applies it to the `struct` base name as used in conjunction with associated constants to form the final name of the constant.


docs.md
```diff
@@ -828,6 +828,12 @@ deprecated_with_notes = "DEPRECATED_STRUCT_WITH_NOTE"
 # default: false
 # associated_constants_in_body: false
 
+# The rename rule to apply to the struct name used for prefixing associated
+# constants.
+#
+# default: "None"
+associated_constants_rename_base_name: "None"
```
emilio, line 835: This should use `= ` right?

emilio (changes requested):
Sorry for the massive lag reviewing this. This looks good modulo a typo in the docs, and the fact that the name is not _great_...

The name implies that it renames the base name only but it also renames the constant name doesn't it?

Also, all rename rules currently start with `rename-*`, so would be good to make that work.

Maybe just `rename-associated-constant`?

danielocfb:
Thanks for the review. I renamed it as per your suggestion and fixed the typo (thanks for catching that).

emilio (approved):

PR #1036, opened 2025-01-03
Add aarch64/arm64 to CI

### LINKED ISSUE #1032: Adding arm64 builds in releases?
Hi!

Would you mind adding arm64 ubuntu builds in releases?

Regards

NickeZ:
fixes #1032

This is a proof of concept to build and test arm64 executables. If this concept is acceptable then I would like to refactor to a "strategy matrix" build.

edit: The last build in my repo is successful: https://github.com/NickeZ/cbindgen/actions/runs/12596962803/job/35109043466, so hopefully this rebased version also works.

### DIFF
--- a/.github/workflows/cbindgen.yml
+++ b/.github/workflows/cbindgen.yml
@@ -88,3 +88,72 @@ jobs:
       run: |
         cargo +nightly update -Zminimal-versions
         cargo +nightly test
+
+  build-arm64:
+
+    runs-on: ubuntu-latest
+    env:
+      CARGO_BUILD_TARGET: aarch64-unknown-linux-gnu
+      CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_RUNNER: qemu-aarch64-static
+      CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER: aarch64-linux-gnu-gcc-11
+
+    steps:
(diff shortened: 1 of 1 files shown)


emilio (commented):
Looks reasonable at a glance, yeah, but using a matrix would be better indeed.

NickeZ:
> Looks reasonable at a glance, yeah, but using a matrix would be better indeed.

Would you mind approving the ci job to run?

emilio:
Gah, for some reason I was enrolled on some github experiment that didn't show the "approve and run" button.

emilio:
Seems somewhat known given the discussion in https://github.com/orgs/community/discussions/143787.

NickeZ:
Do you prefer that I squash the commits or not?

emilio (approved):
Thanks, this looks great. I'm a bit unsure about how reliable the cross-arch sources.list stuff is, it seems ideally it shouldn't be needed? But alas, let's try this.

NickeZ:
The official repos don't have arm64 and the Ubuntu docker images uses ports.ubuntu.com, so I hope it is maintained.

I tried using "cross" package as well for libc. But then I had issues running the tests that call executables. I guess std::process doesn't work with binfmt or something.

One option would be to call all binary tests with "CARGO_..._RUNNER" if available.

PR #1050, opened 2025-01-25
Resolve the order of dependencies in the clike language backend.

flaviojs:
It can reorder dependencies and add forward declarations.
Warns if there is stuff that is cannot resolve.

Includes inner-working traces that can be seen with `cbindgen -vv` (should be left there to facilitate future changes).

Includes tests for assumptions that `ResolverOrder` is making.

Includes a clippy MSVR change to 1.74 (same as `Cargo.toml`) to fix a warning.

----

Before this I had to add extra stuff in `after_includes` for about 10% of symbols that ended up in `items`. All those are handled, leaving only the cases where reordering items and adding forward declarations cannot fix the problem (like constants used in structs appearing after the items are written).

I'm sure there are C/C++ cases that are not being handled, but the documentation and tests should facilitate implementing whatever is left.

I reeeally wanted to do some refactoring to facilitate changes and add documentation before attempting this, but the "single person reviewing and approving" bottleneck prevents that solution. I reached a point where it's impossible to continue developing without adding a hack or implementing something like this.
@emilio if there is a proper way for PRs to signal they are "ready to review" after your initial review is done and more changes are applied, please document it somewhere. =~~


flaviojs:
Had to remove the auto-declaration code, which caused unexpected problems.
This means more declarations, which is fine as long as the header compiles.

The resolver should be fuzzed to make sure different the initial orders are always compilable.

tests/clike_language_backend.rs
```diff
@@ -0,0 +1,197 @@
...
+    enum e {
+        V
+    };
+    "#;
+
+    use super::*;
+
+    #[test]
+    fn test_c() {
+        compile(Language::C, CODE);
```
emilio, line 115: We already compile our tests on automation. So please add regular tests rather than this?

src/bindgen/language_backend/clike.rs
```diff
@@ -1009,4 +1036,611 @@ impl LanguageBackend for CLikeLanguageBackend<'_> {
...
+
+    /// Resolve dependencies by computing a new order of items.
+    ///
+    /// Can add item declarations.
+    ///
+    /// Returns `Ok(())` when done.
+    /// Returns `Err(n)` when it gives up with `n` pending items.
+    pub fn resolve(&mut self) -> Result<(), usize> {
+        loop {
+            if self.pending.is_empty() {
```
emilio, line 1170: `while !self.pending.is_empty() {`?

src/bindgen/language_backend/clike.rs
```diff
@@ -1009,4 +1036,611 @@ impl LanguageBackend for CLikeLanguageBackend<'_> {
...
+    pub order: Vec<usize>,
+}
+
+impl<'a> ResolveOrder<'a> {
+    /// Create a new resolver capable of changing the order of the dependencies.
+    ///
+    /// Does not try to preserve the original order.
+    pub fn new(config: &'a Config, dependencies: &'a mut Dependencies) -> Self {
+        let num_items = dependencies.order.len();
+        let mut paths2indexes: MultiMap<Path, usize> = MultiMap::new();
```
emilio, line 1126: Rather than pulling a new dependency for this, I think we can just use a `HashMap<Path, Vec<usize>>` right?

emilio (commented):
So it's been a while I haven't thought about this problem because in general it's very easy to just add forward-declarations are needed.

But it seems requiring another global pass seems rather unfortunate? Rather than that, shouldn't `add_dependencies` be able to track cycles (and break them by inserting a forward declaration if possible)?

E.g., when `add_dependencies` finds a pointer, it seems it shouldn't add a "hard" dependency, but a "soft" one, which can be dealt with by a forward declaration.

Then that information can be used to sort the items and add any forward declaration as needed.

PR #1051, opened 2025-01-29
Improved generation of methods for traits implemented for objects

PuffyWithEyes:
Development challenges arise when working with traits because if I want to ensure that my struct implements all possible methods from a trait, I encounter the problem described in the following code snippets:

```rust
pub trait DummyTrait {
    extern "C" fn dummy_method(self);
}

#[repr(C)]
struct DummyStruct {
    a: isize,
}

impl DummyTrait for DummyStruct {
    #[no_mangle]
    extern "C" fn dummy_method(mut self) {
        self.a = 1;
    }   
}

#[repr(C)]
struct DummyOtherStruct {
    b: isize,
}

impl DummyTrait for DummyOtherStruct {
    #[no_mangle]
    extern "C" fn dummy_method(mut self) {
        self.b = 1;
    }
}
```

```c
#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>

typedef struct MyStruct {
  intptr_t a;
} DummyStruct;

typedef struct OtherStruct {
  intptr_t b;
} DummyOtherStruct;

void dummy_method(struct DummyStruct self);

void dummy_method(struct DummyOtherStruct self);
```

As we can see, we have two functions with the same name. Currently, this limitation can be worked around with somewhat inconvenient solutions, which is why I would like to propose my own solution that generates functions with more distinct names. An example of the generated code after applying my changes:

```c
#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>

typedef struct MyStruct {
  intptr_t a;
} DummyStruct;

typedef struct OtherStruct {
  intptr_t b;
} DummyOtherStruct;

void dummy_method_DummyTraits_DummyStruct(struct DummyStruct self);

void dummy_method_DummyTraits_DummyOtherStruct(struct DummyOtherStruct self);
```

This code will compile. I have also created a separate branch named `traits-method-gen-feature`, where this functionality can be used as a feature.

P.S.: Apologies for any mistakes, as this is my first fork.

### DIFF
--- a/src/bindgen/parser.rs
+++ b/src/bindgen/parser.rs
@@ -543,6 +543,12 @@ impl Parse {
 
                     if let syn::Type::Path(ref path) = *item_impl.self_ty {
                         if let Some(type_name) = path.path.get_ident() {
+                            let trait_name = if let Some(trait_) = &item_impl.trait_ {
+                                Some(&trait_.1.segments[0].ident)
+                            } else {
+                                None
+                            };
+
                             for method in item_impl.items.iter().filter_map(|item| match item {
                                 syn::ImplItem::Fn(method) if !method.should_skip_parsing() => {
                                     Some(method)
@@ -556,6 +562,7 @@ impl Parse {
                                     mod_cfg,
                                     &Path::new(type_name.unraw().to_string()),
                                     method,
+                                    trait_name,
                                 )
                             }
                         }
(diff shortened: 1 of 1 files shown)


PuffyWithEyes (commented):
Although my solution also does not look so beautiful, maybe this can be done more beautifully...

PuffyWithEyes (commented):
This cfg! was superfluous. Because of him was not going

PR #1054, opened 2025-02-17
The ability to generate trait methods using types

PuffyWithEyes:
When writing similar code now:

```rust
pub trait DummyTrait {
    type DummyIn;
    type DummyOut;

    extern "C" fn dummy(self, in_: Self::DummyIn) -> Self::DummyOut;
}

#[repr(C)]
pub struct Dummy0 {
    dummy: usize,
}

impl DummyTrait for Dummy0 {
    type DummyIn = usize;
    type DummyOut = Self;

    #[unsafe(export_name = "dummy_Dummy0")]
    extern "C" fn dummy(self, in_: Self::DummyIn) -> Self::DummyOut {
        Self {
            dummy: in_,
        }
    }
}

#[repr(C)]
pub struct Dummy1 {
    dummy: usize
}

impl DummyTrait for Dummy1 {
    type DummyIn = ();
    type DummyOut = i32;

    #[unsafe(export_name = "dummy_Dummy1")]
    extern "C" fn dummy(self, in_: Self::DummyIn) -> Self::DummyOut {
        0
    }
}
```

Incorrect code will be generated, looking something like this:

```c
#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>

typedef struct {
  uintptr_t dummy;
} Dummy0;

typedef struct {
  uintptr_t dummy;
} Dummy1;

DummyOut dummy_Dummy0(Dummy0 self, DummyIn in_);

DummyOut dummy_Dummy1(Dummy1 self, DummyIn in_);
```

My fix allows the code to be generated correctly:

```c
#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>

typedef struct {
  uintptr_t dummy;
} Dummy0;

typedef struct {
  uintptr_t dummy;
} Dummy1;

Dummy0 dummy_Dummy0(Dummy0 self, uintptr_t in_);

int32_t dummy_Dummy1(Dummy1 self);
```

### DIFF
--- a/src/bindgen/parser.rs
+++ b/src/bindgen/parser.rs
@@ -702,6 +702,7 @@ impl Parse {
     }
 
     /// Loads a `fn` declaration inside an `impl` block, if the type is a simple identifier
+    #[allow(clippy::too_many_arguments)]
     fn load_syn_assoc_methods(
         &mut self,
         config: &Config,
@@ -766,32 +767,25 @@ impl Parse {
                                 // Search and replacement of the type of method with associative
                                 // types
                                 items_ty.iter().for_each(|&ty| {
-                                    if ret_path.name() == &ty.ident.to_string() {
-                                        match method.ret.try_set_assoc_name(format!("{}_{}_{}", self_path.name(), trait_ident, ty.ident)) {
-                                            Err(e) => {
-                                                warn!("It is not possible to chage return type for method {}::{} - ({})", crate_name, method.path.name(), e);
-                                                return;
-                                            }
-                                            Ok(..) => {}
+                                    if ty.ident == ret_path.name() {
+                                        if let Err(e) = method.ret.try_set_assoc_name(format!("{}_{}_{}", self_path.name(), trait_ident, ty.ident)) {
+                                            warn!("It is not possible to chage return type for method {}::{} - ({})", crate_name, method.path.name(), e);
+                                            return;
                                         }
                                     }
                                     // Removing all arguments with `void` type
                                     method.args.retain(|arg| if let (Some(p), syn::Type::Tuple(syn::TypeTuple { ref elems, .. })) = (arg.ty.get_root_path(), &ty.ty) {
-                                        !(elems.is_empty() && p.name() == &ty.ident.to_string())
+                                        !(elems.is_empty() && ty.ident == p.name())
                                     } else {
                                         true
                                     });
                                     if let Some(arg) = method.args.iter_mut().find(|arg| if let Some(p) = arg.ty.get_root_path() {
-                                        p.name() == &ty.ident.to_string()
+                                        ty.ident == p.name()
                                     } else {
                                         false
                                     }) {
-                                        match arg.ty.try_set_assoc_name(format!("{}_{}_{}", self_path.name(), trait_ident, ty.ident)) {
-                                            Err(e) => {
-                                                warn!("It is not possible to chage method's argument {}::{} - ({})", crate_name, method.path.name(), e);
-                                                return;
-                                            }
-                                            Ok(..) => {}
+                                        if let Err(e) = arg.ty.try_set_assoc_name(format!("{}_{}_{}", self_path.name(), trait_ident, ty.ident)) {
+                                            warn!("It is not possible to chage method's argument {}::{} - ({})", crate_name, method.path.name(), e);
                                         }
                                     }
                                 });


src/bindgen/parser.rs
```diff
@@ -533,13 +533,15 @@ impl Parse {
...
-                    if has_assoc_const {
-                        impls_with_assoc_consts.push(item_impl);
-                    }
+                    let mut impls_with_assoc_ty = Vec::new();
+
+                    item_impl.items.iter().for_each(|item| {
+                        if matches!(item, syn::ImplItem::Const(_)) {
+                            impls_with_assoc_consts.push(item_impl);
+                        } else if let syn::ImplItem::Type(t) = item {
+                            impls_with_assoc_ty.push((&t.ident, &t.ty));
```
emilio, line 544: This array seems wrong, it's not storing `impls`, but ident and types, so maybe just call the vector `associated_tys` or so?

src/bindgen/parser.rs
```diff
@@ -549,13 +551,46 @@ impl Parse {
                                 }
                                 _ => None,
                             }) {
+                                let mut method = method.clone();
+                                let out = match Type::load_from_output(&method.sig.output) {
+                                    Ok((out, ..)) => out,
+                                    Err(..) => continue,
+                                };
+
+                                for (ident, ty) in &impls_with_assoc_ty {
```
emilio, line 562: So I'm a bit confused as to what are you trying to do here. Can you add some comments to clarify?

Let's see if I get this right... When you see a method, you go through the return and argument types, and try to fix them up by matching with the associated types.

I don't think this is quite the right thing to do. I guess this does work for very simple cases, but:

 * At the very least you should do this recursively.
 * I think ideally, we emit the right thing which would be something like:

```
typedef void Dummy0_DummyTrait_DummyIn;
// etc
```

And then `load_syn_method` just knows about associated paths with `Self` and deals with that. Doing this mapping on the syn types seems a bit funky / sketchy at least.

PuffyWithEyes: @emilio in my first trial version, the implementation was approximately the way you described, but I thought that it would be excessive. If you think that it will be better - I will do so

PuffyWithEyes:
Perhaps this execution will be acceptable...

src/bindgen/parser.rs
```diff
@@ -533,33 +534,23 @@ impl Parse {
...
+                        } else if matches!(item, syn::ImplItem::Fn(_))
+                            && !assoc_method_find
+                            && matches!(*item_impl.self_ty, syn::Type::Path(_))
+                        {
+                            impls_with_assoc_fn.push(item_impl);
+                            assoc_method_find = true;
                         }
-                    }
+
+                        assoc_const_find && assoc_method_find
```
emilio, line 552: I'm confused, doesn't this stop once we've found both an associated constant and an associated method? Why? Shouldn't we iterate over all the items? This needs comments.

src/bindgen/parser.rs
```diff
@@ -653,25 +702,113 @@ impl Parse {
...
+        for item in items_fn {
+            if !config
+                .parse
+                .should_generate_top_level_item(crate_name, binding_crate_name)
+            {
+                info!(
+                    "Skip {}::{} - (fn's outside of the binding crate are not used).",
+                    crate_name, &item.sig.ident
+                );
+                return;
```
emilio, line 741: This should be outside of the `for` loop, or continue, right? Why is it here?

src/bindgen/parser.rs
```diff
@@ -718,40 +853,32 @@ impl Parse {
...
-        let loggable_item_name = || {
-            let mut items = Vec::with_capacity(3);
-            items.push(crate_name.to_owned());
-            if let Some(ref self_type) = self_type {
-                items.push(self_type.to_string());
-            }
-            items.push(sig.ident.unraw().to_string());
-            items.join("::")
-        };
+        let loggable_item_name = format!("{}::{}", crate_name, sig.ident);
```
emilio, line 856: The reason this was done was to avoid having to build this string unconditionally...

src/bindgen/utilities.rs
```diff
@@ -50,6 +50,7 @@ impl SynItemHelpers for syn::ImplItemFn {
     fn exported_name(&self) -> Option<String> {
         self.attrs
             .attr_name_value_lookup("export_name")
+            .or_else(|| self.unsafe_attr_name_value_lookup("export_name"))
```
emilio, line 53: This fix seems like could be sent independently.

src/bindgen/parser.rs
```diff
@@ -653,25 +702,113 @@ impl Parse {
...
+                return;
+            }
+
+            let loggable_item_name = format!("{}::{}::{}", crate_name, self_path, item.sig.ident);
+
+            let is_extern_c = item.sig.abi.is_omitted() || item.sig.abi.is_c();
+            let exported_name = item.exported_name();
+
+            match (is_extern_c, exported_name) {
+                (true, Some(exported_name)) => {
```
emilio, line 750: Let's do instead:

```
if !is_extern_c {
    if exported_name.is_some() {
        warn!("...");
    }
    continue;
}
let Some(exported_name) = exported_name else {
    warn!("...");
    continue;
};
```

Then de-indent the whole thing.

src/bindgen/ir/ty.rs
```diff
@@ -627,6 +627,38 @@ impl Type {
...
+        T: Into<String>,
+    {
+        let mut current = self;
+        loop {
+            match current {
+                Type::Ptr { ref mut ty, .. } => current = ty,
+                Type::Path(ref mut generic) => {
+                    generic.set_assoc_ty(assoc_name);
+
+                    break Ok(());
```
emilio, line 641: Just `return Ok(())` right?

src/bindgen/ir/ty.rs
```diff
@@ -627,6 +627,38 @@ impl Type {
...
     }
 
+    pub fn try_set_assoc_name<T>(&mut self, assoc_name: T) -> Result<(), String>
+    where
+        T: Into<String>,
+    {
+        let mut current = self;
+        loop {
+            match current {
+                Type::Ptr { ref mut ty, .. } => current = ty,
```
emilio, line 637: Is this actually right? What happens if I return different levels of pointer-ness?

PR #1055, opened 2025-02-20
Fixed error of generating structures with arrays inside and keyword as

PuffyWithEyes:
Created a correction associated with this [issue](https://github.com/mozilla/cbindgen/issues/1031)


PuffyWithEyes:
Perhaps you should add converting to one or another type (C-style)

src/bindgen/ir/ty.rs
```diff
@@ -270,6 +270,14 @@ impl ConstExpr {
                 let generic_path = GenericPath::load(&path.path)?;
                 Ok(ConstExpr::Name(generic_path.export_name().to_owned()))
             }
+            syn::Expr::Cast(ref cast) => {
```
emilio, line 273: Just `return ConstExpr::load(&cast.expr)`, that way it also deals with `4 as usize` or what not.

emilio (changes requested):

PuffyWithEyes:
@emilio thank you for the remark!

src/bindgen/ir/ty.rs
```diff
@@ -270,6 +270,7 @@ impl ConstExpr {
                 let generic_path = GenericPath::load(&path.path)?;
                 Ok(ConstExpr::Name(generic_path.export_name().to_owned()))
             }
+            syn::Expr::Cast(ref cast) => Ok(ConstExpr::load(&cast.expr)?),
```
emilio, line 273: This could be simpler by just using `=> Self::load(&cast.expr)` rather than that. But no big deal.

emilio (approved):

PR #1067, opened 2025-04-11
re-export bindgen::Library for proc macros

onlycs:
I'm writing my own proc macro which wraps around `cbindgen`, and I'd like to have access to `Library` to generate bindings without writing to a file and pointing `cbindgen` to that file.

Thanks

### DIFF
--- a/src/bindgen/mod.rs
+++ b/src/bindgen/mod.rs
@@ -67,4 +67,5 @@ pub use self::builder::Builder;
 pub use self::config::Profile; // disambiguate with cargo::Profile
 pub use self::config::*;
 pub use self::error::Error;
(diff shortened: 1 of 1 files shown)


src/bindgen/mod.rs
```diff
@@ -67,3 +67,4 @@ pub use self::builder::Builder;
 pub use self::config::Profile; // disambiguate with cargo::Profile
 pub use self::config::*;
 pub use self::error::Error;
+pub use self::library::Library;
```
emilio, line 70: I'm confused, why does this help? Don't you need `unstable_ir` as well to make this useful? Should this be part of the `unstable_ir` feature?

Issue #1069, opened 2025-04-16
Incompatibility with non-mod.rs module layout


BiagioFesta:
Standard [suggests to prefer package layout like][1]:
```
src/
├─ lib.rs
├─ util.rs
└─ util/
   └─ config.rs
```

instead of
```
src/
├─ lib.rs
└─ util/
   ├─ mod.rs
   └─ config.rs
```

> [...] It is encouraged to use the new naming convention as it is more consistent, [...]

However, this does not seems compatible with current `cbindgen` source parsing: https://github.com/mozilla/cbindgen/blob/b105f550ed0d62e9207cb7da694ca8b988f32f18/src/bindgen/parser.rs#L266-L266

where only `mod.rs` seems to be detected as module.

[1]: https://doc.rust-lang.org/reference/items/modules.html#r-items.mod.outlined.search-mod

PR #1070, opened 2025-04-23
Fix parsing submodule dir for 2018 convention

BiagioFesta:
Fix for: https://github.com/mozilla/cbindgen/issues/1069


emilio:
Ok, so I think in the example in #1069, the `depth == 0` is confusing us. I think we need a test, but also, I'm not convinced hardcoding `lib.rs` is the right fix? IIRC you can totally customize the main path, right? https://doc.rust-lang.org/cargo/reference/cargo-targets.html#configuring-a-target

PR #1083, opened 2025-06-11
feat: enum.merge_generic_tags

morrisonlevi:
This is a continuation of #785 that has merged in the latest master.

There appears to be an issue with combining this with `export.prefix`: https://github.com/mozilla/cbindgen/pull/785#issuecomment-2960777472.


morrisonlevi:
It seems like I've fixed the compatibility with prefixing, though there should be tests added for it.

When running it on my real codebase, I get a duplicate struct. I'm trying to fix that. Making a contained reproducer has been fruitless so far.

morrisonlevi:
I've squashed Ben's commits, and my own, so this can go cleanly into master once reviewed.

Issue #1085, opened 2025-07-01
Incorrect detection of duplicated assosiated constants


xavi-oc:
Consider the following Rust code:
```Rust
pub struct MyStruct {
    pub field_a: u8,
}

impl MyStruct  {
    pub const FIELD_A_RELATED_CONSTANT : u8 = 0x1;
}

pub struct MyOtherStruct {
    pub field_a: u8,
}

impl MyOtherStruct  {
    pub const FIELD_A_RELATED_CONSTANT: u8 = 0x1;
}
```
If we run `cbindgen --crate issue --lang c` the following output will be generated:
```C
#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>

#define MyStruct_FIELD_A_RELATED_CONSTANT 1
```
with an error message that states the following:
* **ERROR: Conflicting name for constant issue::MyOtherStruct::FIELD_A_RELATED_CONSTANT.**

As we can see, cbindgen considers that MyOtherStruct::FIELD_A_RELATED_CONSTANT is duplicated while it is another struct with the same associated constant name which is completly valid.

PR #1086, opened 2025-07-02
Incorrect detection of duplicated assosiated constants

xavi-oc:
This small modification solve the issue #1085

The idea behind this change is to provide context (impl_path) to the path variable. This way the undesired duplication detection is solved.

In addition, since I don't want to mess with the naming I modified how the `export_name` for the constant item is generated.


xavi-oc:
Added tests as @emilio requested.

When I execute `cargo test` the output is clean and the test is being executed. However, the c file is removed. Maybe I did something wrong could you double check?

Thanks,

xavi-oc:
@emilio  Added tests as requested

emilio (approved):
Thanks!

PR #1093, opened 2025-07-16
Check for CMSE ABI's as well

defermelowie:
Rust has nightly support for Cortex-M Security Extension through the [`cmse-nonsecure-entry`](https://doc.rust-lang.org/beta/unstable-book/language-features/cmse-nonsecure-entry.html) and [`cmse-nonsecure-call`](https://doc.rust-lang.org/beta/unstable-book/language-features/abi-c-cmse-nonsecure-call.html) ABIs. These are compatible with C but cause the compiler to insert extra instructions such as clearing sensitive registers before jumps. 

Therefore it would be convenient if cbindgen was able to generate declarations for functions annotated with these ABIs as well. This PR adds basic support for these ABIs by slightly relaxing the filter.

### DIFF
--- a/src/bindgen/utilities.rs
+++ b/src/bindgen/utilities.rs
@@ -371,24 +371,19 @@ impl_syn_item_helper!(syn::ItemTraitAlias);
 /// Helper function for accessing Abi information
 pub trait SynAbiHelpers {
     fn is_c(&self) -> bool;
+    fn is_cmse(&self) -> bool;
     fn is_omitted(&self) -> bool;
 }
 
 impl SynAbiHelpers for Option<syn::Abi> {
     fn is_c(&self) -> bool {
-        if let Some(ref abi) = *self {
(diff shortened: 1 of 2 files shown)


src/bindgen/utilities.rs
```diff
@@ -400,6 +395,16 @@ impl SynAbiHelpers for syn::Abi {
             false
         }
     }
+    fn is_cmse(&self) -> bool {
+        if let Some(ref lit_string) = self.name {
+            matches!(
+                lit_string.value().as_str(),
+                "cmse-nonsecure-entry" | "cmse-nonsecure-call"
```
emilio, line 402: Hmm, I'm not sure why these function shouldn't be tagged with `__attribute__((cmse_nonsecure_call))` and `__attribute__((cmse_nonsecure_entry))`, respectively. Mind elaborating on that?

emilio (approved):

PR #1100, opened 2025-08-23
Inline variables are only available in C++17 and above 

johnDeSilencio:
## The Problem

Rust code like this:

```rust
#[repr(C)]
pub struct Foo {
  foo: u32,
}

impl Foo {
  pub const SPECIAL_NUM: u32 = 153;
}
```

generates code like this:

```cpp
struct Foo {
  uint32_t foo;
  static const uint32_t SPECIAL_NUM;
};
inline const int32_t Foo::SPECIAL_NUM = 153;
```

However, `inline` variables are only available in C++17 and above: [Godbolt Link](https://godbolt.org/z/3MYKsb9dj). At my company in particular, we're stuck with C++11 for the foreseeable future (😭)

## Solution

Make generating associated constants with the `inline` keyword configurable.

In this PR, I add the new `[const]` config setting `allow_inline` which defaults to `true`. I also added a test when `allow_inline` is set to false.

### DIFF
--- a/tests/expectations/no_inline_constant_constexpr.compat.c
+++ b/tests/expectations/no_inline_constant_constexpr.compat.c
@@ -1,0 +1,19 @@
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <stdlib.h>
+
+#define CONSTANT_I64 216
+
+#define CONSTANT_FLOAT32 312.292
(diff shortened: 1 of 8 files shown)


emilio:
Ah indeed this had been pointed out above, sorry, I jumped directly to the code.

PR #1109, opened 2025-09-24
Allow must_use if a reason is specified ([must_use = "reason"])

ATantalizedLion:
Fixes https://github.com/mozilla/cbindgen/issues/1108

Adds a `has_attr_namevalue` that checks `syn::Meta::NameValue` for the attribute (original only checks ` syn::Meta::Path`). 
must_use now checks for either the `has_attr_word` or the new `has_attr_namevalue`.

I could not get the depfile.rs tests running locally (for both master and my own fork) so I am unsure if they pass.

### DIFF
--- a/src/bindgen/ir/annotation.rs
+++ b/src/bindgen/ir/annotation.rs
@@ -113,7 +113,7 @@ impl AnnotationSet {
             })
             .collect();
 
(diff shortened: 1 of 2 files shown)


emilio (approved):

emilio:
Would be good to add a test for this.

PR #1118, opened 2025-10-28
Expose the line_endings config option to use with the builder

alloncm:
I have been using this awesome tool lately with the builder interface and after getting line endings diffs all the time I have noticed that the line_endings option is not accessible using the builder.

I have manually tested this against my own project, let me know if further testing is required.


emilio (approved):
Looks good, thanks!

PR #1156, opened 2026-03-30
Use C++ fixed-type enumeration syntax under C23 (or higher) as well

garrison:
C23 [has introduced support for enumerations with fixed underlying type](https://open-std.org/JTC1/SC22/WG14/www/docs/n3030.htm), matching the existing C++ syntax for this functionality.

This PR modifies cbindgen to output a header that enables this syntax if either `__cplusplus` is defined (as before) or if the compilation is according to C23 or later (added in this PR).

My motivation for this change is that my toolchain complained that some names were defined twice, both as an enum and a typedef.  I [fixed the error](https://github.com/Qiskit/Qiskit.jl/blob/73bfbb9440168dc76685632f9fb3d00d3a17bd79/gen/generator.jl#L16-L18) by defining `__cplusplus` even thought the compiler is compiling in C mode.  This change will allow me to remove that workaround, as long as C23 or later is set as the C standard.

### DIFF
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -760,7 +760,7 @@ impl Enum {
 
         // Emit typedef specifying the tag enum's size if necessary.
         // In C++ or C23 enums can "inherit" from numeric types (`enum E: uint8_t { ... }`),
-        // but in olders versions of C `typedef uint8_t E` is the only way to give a fixed size to `E`.
+        // but in older versions of C `typedef uint8_t E` is the only way to give a fixed size to `E`.
         if let Some(prim) = size {
             if config.cpp_compatible_c() {
                 out.new_line_if_not_start();
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -680,7 +680,7 @@ impl Enum {
                     }
                     write!(out, " {tag_name}");
 
-                    // Write typed enum syntax (valid in C23 or later or C++)
+                    // Emit typed enum syntax (valid in C23 or later or C++)
                     let cond = if config.cpp_compatible_c() {
                         "defined(__cplusplus) || __STDC_VERSION__ >= 202311L"
                     } else {
--- a/src/bindgen/ir/enumeration.rs
+++ b/src/bindgen/ir/enumeration.rs
@@ -680,15 +680,20 @@ impl Enum {
                     }
                     write!(out, " {tag_name}");
 
-                    if config.cpp_compatible_c() {
-                        out.new_line();
-                        out.write("#if defined(__cplusplus) || __STDC_VERSION__ >= 202311L");
-                        out.new_line();
-                        write!(out, "  : {prim}");
-                        out.new_line();
-                        out.write("#endif // defined(__cplusplus) || __STDC_VERSION__ >= 202311L");
-                        out.new_line();
-                    }
+                    // Write typed enum syntax (valid in C23 or later or C++)
+                    let cond = if config.cpp_compatible_c() {
+                        "defined(__cplusplus) || __STDC_VERSION__ >= 202311L"
+                    } else {
+                        "__STDC_VERSION__ >= 202311L"
+                    };
+
+                    out.new_line();
+                    write!(out, "#if {cond}");
+                    out.new_line();
+                    write!(out, "  : {prim}");
+                    out.new_line();
+                    write!(out, "#endif // {cond}");
+                    out.new_line();
                 } else {
                     if config.style.generate_typedef() {
                         out.write("typedef ");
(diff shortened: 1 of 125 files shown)


emilio (commented):
You probably need to update tests right?

garrison:
> You probably need to update tests right?

Indeed -- and now everything passes locally.

src/bindgen/ir/enumeration.rs
```diff
@@ -680,15 +680,20 @@ impl Enum {
...
-                    if config.cpp_compatible_c() {
-                        out.new_line();
-                        out.write("#ifdef __cplusplus");
-                        out.new_line();
-                        write!(out, "  : {prim}");
-                        out.new_line();
-                        out.write("#endif // __cplusplus");
-                        out.new_line();
-                    }
+                    // Write typed enum syntax (valid in C23 or later or C++)
```
garrison:
```suggestion
                    // Emit typed enum syntax (valid in C23 or later or C++)
```

emilio (approved):
LGTM, I'll probably manually squash it once CI is green because there's a couple merge commits. Thanks!

src/bindgen/ir/enumeration.rs
```diff
@@ -680,15 +680,20 @@ impl Enum {
...
+                        "__STDC_VERSION__ >= 202311L"
+                    };
+
+                    out.new_line();
+                    write!(out, "#if {cond}");
+                    out.new_line();
+                    write!(out, "  : {prim}");
+                    out.new_line();
+                    write!(out, "#endif // {cond}");
+                    out.new_line();
```
cyphar, line 696: Unfortunately, unconditionally outputting this causes issues with some tools that parse header files.

In particular, Python's `cffi` has very limited support for preprocessor macros and **does not support `#if` conditionals at all**. For some conditionals this is trivial to work around (such as header guards -- you can remove them before getting `cffi` to parse them) but conditionals like this do not have a nice workaround.

If you try to use them, you get this error:

```py3
>>> import cffi
>>> ffi = cffi.FFI()
>>> ffi.cdef("#if FOO >= 100\n#endif\n")
cffi.CDefError: cannot parse "#if FOO >= 100"
<cdef source string>:1:1: Directives not supported yet
```

cyphar/libpathrs#382 is an example of how this can cause build failures in a downstream project -- [our build scripts strip out any non-`#define` macros](https://github.com/cyphar/libpathrs/blob/v0.2.4/contrib/bindings/python/pathrs/pathrs_build.py#L26) (which is the only thing `cffi` supports, and even then it's quite limited) so [the error you actually get](https://github.com/cyphar/libpathrs/actions/runs/26678106907/job/78633589980?pr=382) looks more like:

```
cffi.CDefError: cannot parse ": uint64_t"
<cdef source string>:22:3: before: :
```

But the core issue is the same. It would be nice if this output could be made configurable so that we can continue to use `cbindgen`-generated headers with `cffi`. Otherwise we will have to pin an older version...

src/bindgen/ir/enumeration.rs
```diff
@@ -759,17 +764,39 @@ impl Enum {
...
+                    out.write("#else");
+                }
+
                 out.new_line();
                 write!(out, "{} {} {};", config.language.typedef(), prim, tag_name);
+
+                if config.language == Language::C {
+                    out.new_line();
+                    out.write("#endif // __STDC_VERSION__ >= 202311L");
+                }
```
cyphar, line 799: And this change has a very similar issue -- I believe this will cause multiple-defines errors if our build script strips out the `#if` directives.

src/bindgen/ir/enumeration.rs
emilio: I would be ok with reviewing a patch to make this opt-in / opt-out.

Though, have you considered `unifdef` perhaps? I think that'd work?

cyphar: That's not a bad idea (and I must admit I hadn't considered it) -- unfortunately shelling out to `unifdef` from the setup script is a little dodgy, it might not work for some packaging setups (distros would probably be okay with the workaround but I think it would be problematic for Python packaging and people doing local from-source installs).

It would be easier to implement that in Python, at which point it might actually end up being less effort to just add support for basic `#if`/`#ifdef` conditionals in `cffi`...

> I would be ok with reviewing a patch to make this opt-in / opt-out.

What would you think about a patch which let you specify a C version to generate (to not make this switch too special-purpose) and so if you pick C99 mode (for instance) you don't get these `#if` conditionals but if you pick C23 mode you do? (To be fair, reading it back this is a little funky -- if you are picking C23 mode you wouldn't need the conditional. Maybe a "max" and "min" C version? :shrug:.)

Or would you prefer something more like an option for "limited C" and "full-on C"? (That would make things easier for me in the future, I must admit! :sweat_smile:)

emilio: I think something like `allow_fixed_size_c_enums` or something, defaulting to `true` and avoiding the ifdefs would be a bit more consistent with how we deal with other features like `constexpr` in c++.

It'd be a breaking change but it'd remove the ifdefs altogether from the generated code? I guess the ifdef route was taken here mostly because we already had a `__cplusplus` if def for C++-compatible-C but...

tzemanovic: I'm hitting the same issue elsewhere too, it would be nice to restore pre C23 version as default

PR #1167, opened 2026-07-04
Value-initialize the result local in generated C++ tagged-enum constructors

sylvestre:
Emit `T result{};` instead of `T result;` in the generated static variant constructors so the payload union is zero-initialized rather than left indeterminate. For no-payload variants only the tag is set, so the previously-uninitialized union produced false "uninitialized scalar" reports from static-analysis tools (e.g. Coverity). Payload variants are unaffected: the union is value-initialized and the payload is then constructed in place via placement new as before.

Add a regression test (tagged_enum_value_init) covering no-payload and payload variants, and regenerate the affected C++ expectations.

This will remove about 110 CID in coverity for firefox


emilio (changes requested):
@sylvestre I'm confused, why do we want this?

We're methodically initializing the fields right below... So it seems like at worst we're doing more work, and if we're lucky we're only giving more work to the compiler to optimize it out?

PR #1169, opened 2026-07-10
bitflags: allow parsing unnamed ("_") bitflags

nullstalgia:
To avoid potentially truncating unknown bits within the derived methods, [the documentation recommends](https://docs.rs/bitflags/latest/bitflags/#externally-defined-flags) adding an unnamed flag with all bits set. This commit allows cbindgen to skip parsing those flags without throwing an error (like it did previously, then skipping the entire item!), and only emit the named flags with values.

Before:
```
WARN: Failed to parse bitflags invocation: Error("expected identifier, found keyword `_`")
...
typedef uint8_t Flags;
```

After:
```
INFO: Take my_crate::Flags.
DEBUG: Continuing past unnamed bitflag
INFO: Take my_crate::Flags::A.
INFO: Take my_crate::Flags::B.
INFO: Take my_crate::Flags::C.
INFO: Take my_crate::Flags::D.
...
typedef uint8_t Flags;
#define Flags_A (uint8_t)(1 << 0)
#define Flags_B (uint8_t)(1 << 1)
#define Flags_C (uint8_t)(1 << 2)
#define Flags_D (uint8_t)(1 << 3)
```

I'm unsure if `unraw()` would be needed due to my use of the `IdentExt::parse_any`, might be worth adding as a test case?

And in order to pass `cargo clippy` with yesterday's Stable release (1.97.0), I had to do another commit with some fixes. Let me know if you'd prefer a cleaner history.

### DIFF
--- a/src/bindgen/bitflags.rs
+++ b/src/bindgen/bitflags.rs
@@ -4,6 +4,7 @@
 
 use proc_macro2::TokenStream;
 use std::collections::HashSet;
+use syn::ext::IdentExt;
 use syn::fold::Fold;
 use syn::parse::{Parse, ParseStream, Parser, Result as ParseResult};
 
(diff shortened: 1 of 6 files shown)


emilio (approved):
Looks good to me, thanks!

PR #1171, opened 2026-07-10
Emit string and byte-string constants (#927, #546)

### LINKED ISSUE #546: Support for constant byte literal
The code contains a [TODO](https://github.com/eqrion/cbindgen/blob/bb00d1c4a8e53c56880586987f0e1512fb252b9a/src/bindgen/ir/constant.rs#L220) for adding byte string support.

Is it on the roadmap and/or what would a preferred implementation look like?

My use case looks something like this:
```rust
pub const GET_INFO_FN_NAME: &'static u8 = b"get_info\0";
```

### LINKED ISSUE #927: Support 'pub const BAR: &CStr = c"hello world";'
We have C string literals now in Rust stable since https://github.com/rust-lang/rust/issues/105723.

It seems reasonable to have const string literals like these:

```rs
use std::ffi::CStr;
pub const BAR: &CStr = c"hello world";
```

and convert them to `#define` or `constexpr` or whatever cbindgen usually does with constants

Current behavior is very much not ideal (probably need to update syn to v2?):

```sh
$ cbindgen --lang C
thread 'main' panicked at /home/user/.cargo/registry/src/index.crates.io-6f17d22bba15001f/syn-1.0.109/src/lit.rs:1020:13:
Unrecognized literal: `c"hello world"`
```

cf-rhett:
Closes #927 and closes #546.

`&str`, `&CStr`, and byte-string (`b"..."`) constants were previously dropped silently. This emits them:

- A scalar `&str`/`&CStr` becomes `const char NAME[]` in the C++/Cython typed forms — an array, not a pointer, so `sizeof` gives the length at compile time and the declaration mirrors its string-literal initializer. In C it stays a bare `#define`. An array of them (`[&CStr; N]`) instead maps to `const char *[N]`, since ragged element lengths can't form a `char[N][]`.
- Byte-string constants (`b"..."`) become `uint8_t[]` array initializers rather than string literals, so non-printable and non-NUL-terminated data survives (per the guidance in #546). The declared type may be the unsized `&[u8]`, which cbindgen can't load, so the array length is taken from the literal. An empty byte string yields the same zero-length array as any other empty array constant — it is not special-cased.

### Design notes (to save reviewer time)

- **Why an array, not `const char *`.** `&str`/`&CStr` are fat/thin pointers, so no C type is ABI-compatible with them. But a Rust `const` has no symbol or address — it is inlined at every use site — so there is nothing to link against and no ABI to preserve. What a C consumer wants is the string *value*, and a `char[]` is the faithful lowering: it carries the length (`sizeof`) and preserves interior NULs, which the pointer form couldn't.
- **Where the type rewrite lives.** The rewrite is confined to `Constant::load`, not the global type-simplification pass. Collapsing these types is only sound for a constant *definition*, where the runtime representation is discarded in favor of the literal; it must never apply to struct fields, function arguments, or returns.
- **Byte strings are handled one level up.** `Literal::load` sees only the expression, so it can produce a self-describing string literal but cannot synthesize the `uint8_t[N]` *type* a byte string needs — that requires the length. `load_byte_string` derives the type and initializer together from the literal bytes, which also makes the sized (`&[u8; N]`) and unsized (`&[u8]`) forms behave identically. Nested byte strings remain unsupported (same silent-skip as before).
- **Escaping.** The C string escaper renders non-ASCII UTF-8 bytes as `\xNN` (not Rust's `\u{...}`, which is invalid C), escapes `?` to prevent trigraph formation, and splits a literal when a `\xNN` escape is followed by a hex-digit character (`"...\xNN" "f..."`) so the escape stays a single byte.

### Tests

New fixtures cover string edge cases (empty, interior NUL, non-UTF-8 `&CStr`, hex-digit adjacency in both cases, control/DEL bytes, trigraphs, arrays), byte strings (sized, unsized slice, binary, empty), and the `allow_static_const`/`allow_constexpr` off fallback path. Several existing `constant`/`assoc_constant`/`namespace*` expectations gained the `&str` constants they previously dropped.


cf-rhett:
Gentle ping on this @emilio — I think it's waiting on a maintainer to approve the CI run (no checks have kicked off yet). This implements the `uint8_t[]` approach you outlined in #546 and also closes #927. Happy to rebase or adjust anything. Thanks!

src/bindgen/ir/constant.rs
```diff
@@ -562,16 +689,20 @@ impl Constant {
         attrs: &[syn::Attribute],
         associated_to: Option<Path>,
     ) -> Result<Constant, String> {
-        let ty = Type::load(ty)?;
-        let mut ty = match ty {
-            Some(ty) => ty,
+        let (mut ty, mut lit) = match byte_string_bytes(expr) {
```
emilio: A bit hacky, but fair enough I guess?

tests/expectations/assoc_constant.cpp
```diff
@@ -8,6 +8,7 @@ struct Foo {
 
 };
 constexpr static const int32_t Foo_GA = 10;
+constexpr static const char *Foo_BU = "hello world";
```
emilio: Is this intentional choice of char pointer intentional out of curiosity? `const char Foo_BU[] = `  might be more useful as it allows you to get the length at compile time more easily, and it's what the actual literal would produce.

emilio (commented):
Looks sensible. One question about the const char* version of this, and thanks for the ping!

src/bindgen/ir/constant.rs
cf-rhett: Yeah, it was. Pulled it into a single `load_byte_string(expr) -> Option<(Type, Literal)>` so it reads as "byte string, else the usual typed load." The type and value have to come out together: `&[u8; N]` loads as a pointer-to-array rather than a bare array, and `&[u8]` doesn't load at all, so I synthesize the `uint8_t[N]` from the literal length right next to the initializer. Nice side effect is the sized and unsized forms end up identical.

tests/expectations/assoc_constant.cpp
cf-rhett: Yeah, agreed — done. Scalar `&str`/`&CStr` now emit `const char NAME[]`. Compile-time `sizeof` is the obvious win, and it also stops a `&str` with an interior NUL from getting silently truncated the way the pointer form did. Arrays of them stay `const char *[N]` though — ragged lengths can't be `char[N][]`. C is unaffected, still a bare `#define`.

emilio (approved):

PR #1172, opened 2026-07-16
Update TOML to version 1 and MSRV to 1.76

oscargus:
Using the most recent TOML versions will require edition 2024, but just going into the 1-series will make a bit difference for upstream crates.

Edit: MSRV had to be updated to 1.76 due to serde_spanned 1.0.4 required for TOML 1.

Edit 2: I managed to figure out how to get the lock-file to update to newest version that still can be compiled with 1.76. Good thing is that it reduces in size.

### DIFF
--- a/.github/workflows/cbindgen.yml
+++ b/.github/workflows/cbindgen.yml
@@ -33,7 +33,7 @@ jobs:
 
     - name: Install minimum supported Rust version
       id: msrv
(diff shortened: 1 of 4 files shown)


emilio (approved):

oscargus:
Rebased and fixed issue (different syn version).

PR #1174, opened 2026-07-22
Overwrite `CARGO_BUILD_BUILD_DIR` as well to avoid deadlocks

SwishSwushPow:
Hi everyone,

I was playing around with `CARGO_BUILD_BUILD_DIR` (https://releases.rs/docs/1.91.0/#cargo) the last couple of days and noticed that my Rust workspace, which uses cbindgens `[pase.expand]` feature, would deadlock when building.

Since cbindgen is already handling this issue related to `CARGO_TARGET_DIR` I wanted to open this PR to apply the same treatment to `CARGO_BUILD_BUILD_DIR`. I think that it is probably ok to not introduce a new `CARGO_EXPAND_BUILD_DIR` env var because I feel like it would be ok for the expand-related artifacts to be located in the same place.

I hope that it is ok that I've opened this PR. I am open to discuss any changes or additions that are suggested. :)

### DIFF
--- a/src/bindgen/cargo/cargo_expand.rs
+++ b/src/bindgen/cargo/cargo_expand.rs
@@ -14,6 +14,11 @@ use std::str::{from_utf8, Utf8Error};
 extern crate tempfile;
 use self::tempfile::Builder;
 
+fn set_target_dir(cmd: &mut Command, path: &Path) {
+    cmd.env("CARGO_TARGET_DIR", path);
+    cmd.env("CARGO_BUILD_BUILD_DIR", path);
+}
+
 #[derive(Debug)]
 /// Possible errors that can occur during `rustc -Zunpretty=expanded`.
 pub enum Error {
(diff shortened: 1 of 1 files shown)


src/bindgen/cargo/cargo_expand.rs
```diff
@@ -14,6 +14,11 @@ use std::str::{from_utf8, Utf8Error};
 extern crate tempfile;
 use self::tempfile::Builder;
 
+fn set_target_dir(cmd: &mut Command, path: &Path) {
+    cmd.env("CARGO_TARGET_DIR", path);
+    cmd.env("CARGO_BUILD_BUILD_DIR", path);
```
emilio, line 19: `CARGO_BUILD_BUILD_DIR`?

emilio: Huh, sorry, I somehow for the whole commit description I read `CARGO_BUILD_DIR` o.O

emilio (approved):

emilio:
I think this is fine, thanks!

PR #1175, opened 2026-07-26
refactor: bump `syn` to v3

AudaciousAxiom:
Following the [recent release of `syn` v3](https://github.com/dtolnay/syn/releases/tag/3.0.0), this bumps the dependency and makes the necessary refactors, with the goal of reducing dependency duplication in the dependents. The [MSRV of `syn` v3 is now 1.71](https://crates.io/crates/syn/3.0.0/code/Cargo.toml#L14) but that shouldn't be a problem with `cbindgen`'s MSRV of 1.74.

I've been trough the changelog and didn't find any other relevant breaking changes (but there is quite a few of them so I could have missed some). The biggest change is how receivers and the constness of pointers is now encoded: compare [`Receiver@v3`](https://docs.rs/syn/3.0.3/syn/struct.Receiver.html) with [`Receiver@v2`](https://docs.rs/syn/2.0.119/syn/struct.Receiver.html) and [`TypePtr@v3`](https://docs.rs/syn/3.0.3/syn/struct.TypePtr.html) with [`TypePtr@v2`](https://docs.rs/syn/2.0.119/syn/struct.TypePtr.html).

I've run `cargo test` but didn't do any other kind of testing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -354,9 +354,9 @@ checksum = "94143f37725109f92c262ed2cf5e59bce7498c01bcc1502d7b9afe439a4e9f49"
 
 [[package]]
 name = "serde"
-version = "1.0.228"
+version = "1.0.229"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9a8e94ea7f378bd32cbbd37198a4a91436180c5bb472411e48b5ec2e2124ae9e"
+checksum = "4148590afebada386688f18773da617792bf2ef03ffc1e4cbd2b1d45b023e0ba"
 dependencies = [
  "serde_core",
  "serde_derive",
@@ -364,22 +364,22 @@ dependencies = [
 
 [[package]]
 name = "serde_core"
-version = "1.0.228"
+version = "1.0.229"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "41d385c7d4ca58e59fc732af25c3983b67ac852c1a25000afe1175de458b67ad"
+checksum = "67dca2c9c51e58a4791a4b1ed58308b39c64224d349a935ab5039aa360942a48"
 dependencies = [
  "serde_derive",
 ]
 
 [[package]]
 name = "serde_derive"
-version = "1.0.228"
+version = "1.0.229"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d540f220d3187173da220f885ab66608367b6574e925011a9353e4badda91d79"
+checksum = "e7a5d71263a5a7d47b41f6b3f06ba276f10cc18b0931f1799f710578e2309348"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.104",
+ "syn 3.0.3",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,7 +78,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serial_test",
- "syn",
+ "syn 3.0.3",
  "tempfile",
  "toml",
 ]
@@ -379,7 +379,7 @@ checksum = "d540f220d3187173da220f885ab66608367b6574e925011a9353e4badda91d79"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn",
+ "syn 2.0.104",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


src/bindgen/ir/function.rs
```diff
@@ -224,17 +224,26 @@ trait SynFnArgHelpers {
 
 fn gen_self_type(receiver: &syn::Receiver) -> Result<Type, String> {
     let mut self_ty = Type::Path(GenericPath::self_path());
+    let is_const;
 
-    // Custom self type
-    if receiver.colon_token.is_some() {
-        self_ty = Type::load(receiver.ty.as_ref())?.unwrap_or(self_ty);
-    }
+    match &receiver.kind {
```
AudaciousAxiom: This is the part that requires the most attention due to the change around receivers and pointers.

Cargo.lock
AudaciousAxiom, line 1: Versions of `serde_derive` ~~and of `serial_test_derive`~~ (edit: actually `serial_test@v4` requires Rust 1.93) that use `syn` v3 have been released, so it would be possible to reduce duplication in the lockfile by simply bumping these (and therefore completely remove the dependency on `syn` v2). I can do this in this PR if you want, or it can be left for another PR.

emilio: Yeah would be good to avoid depending on syn v2 altogether if we can.

src/bindgen/ir/function.rs
```diff
@@ -224,17 +224,26 @@ trait SynFnArgHelpers {
...
-    if receiver.reference.is_none() {
-        return Ok(self_ty);
+            match &**ty {
+                syn::Type::Reference(ref_ty) => {
+                    is_const = ref_ty.mutability.is_none();
+                }
+                _ => return Ok(self_ty),
+            }
+        }
+        syn::ReceiverKind::Value | _ => return Ok(self_ty),
```
emilio: Rather than assume that `_` is handled like `self`, let's just error on unknown values.

src/bindgen/ir/function.rs
emilio: `let is_const = match ... {` please

src/bindgen/ir/function.rs
```diff
@@ -224,17 +224,26 @@ trait SynFnArgHelpers {
...
+        syn::ReceiverKind::Typed(_, ty) => {
+            self_ty = Type::load(ty)?.unwrap_or(self_ty);
 
-    if receiver.reference.is_none() {
-        return Ok(self_ty);
+            match &**ty {
+                syn::Type::Reference(ref_ty) => {
+                    is_const = ref_ty.mutability.is_none();
+                }
+                _ => return Ok(self_ty),
```
emilio, line 236: This doesn't seem right? doesn't this break `self: Box<Foo>` or so to be `self`?

emilio (changes requested):
Thanks for doing this!

src/bindgen/ir/function.rs
AudaciousAxiom: I'm not sure but I think it's fine: this matches the case where both `if`'s conditions were true (`colon_token` present but the type is not a reference). I think this is tested by the `SelfTypeTestStruct_should_not_exist_box` test case.

src/bindgen/ir/function.rs
AudaciousAxiom: That indeed sounds more robust, I've added an error message.

Cargo.lock
AudaciousAxiom: I've added a commit that bumps `serde` (and therefore `serde_derive`). Bumping `serial_test` would require increasing the MSRV of `cbindgen` to 1.93; is this ok? If so I can also bump the MSRV and the dependency.

src/bindgen/ir/function.rs
AudaciousAxiom: Indeed! Done

Cargo.lock
AudaciousAxiom: I don't think bumping `cbindgen`'s MSRV to 1.93 just to bump `serial_test`, which is only a dev-dependency, is worth it: `syn` will only be compiled twice for tests and examples; I think preserving the compatibility for the build-script use case is more important.

emilio: Yeah that's fine, thanks.

src/bindgen/ir/function.rs
emilio: Yeah it's fine, we do `Type::load()` earlier so it should get the `Box<>` indeed.

emilio (approved):

PR #1178, opened 2026-08-22
Refactor/update various blocks

rytheo:
Minor style pass as I familiarize myself with the codebase.


src/bindgen/ir/function.rs
```diff
@@ -101,13 +101,11 @@ impl Function {
...
-            let mut items = Vec::with_capacity(self.args.len());
-            for arg in self.args.iter() {
-                items.push(format!("{}:", arg.name.as_ref()?.as_str()));
-            }
-            items.join("")
-        };
+        let item_args = self
+            .args
+            .iter()
+            .map(|arg| Some(format!("{}:", arg.name.as_ref()?.as_str())))
```
emilio, line 107: Not sure this is particularly easier to reason about tbh. But it's equivalent.

emilio (approved):
Thanks!

PR #1181, opened 2026-08-28
Emit cfg helper macros for struct-literal `#define` constants 

youknowone:
another trial of #955

### DIFF
--- a/src/bindgen/ir/cfg.rs
+++ b/src/bindgen/ir/cfg.rs
@@ -257,7 +257,7 @@ impl ToCondition for Cfg {
     }
 }
 
-#[derive(Debug, Clone)]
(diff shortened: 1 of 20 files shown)


src/bindgen/ir/cfg.rs
```diff
@@ -266,6 +266,63 @@ pub enum Condition {
...
+        self.append_macro_ident(&mut name);
+        name
+    }
+
+    fn append_macro_ident(&self, out: &mut String) {
+        match self {
+            Condition::Define(define) => {
+                out.push('D');
+                out.push_str(&define.len().to_string());
+                out.push('_');
```
youknowone, line 283: a little bit of mangling to avoid accidental name collision. please tell me a better approach if this is not the best

youknowone (commented):
@emilio This is a replacement of #956

PR #1182, opened 2026-08-31
Add export.include_all for public types

### LINKED ISSUE #605: Option to export all the types that are public and repr(C)
Currently, cbindgen only considers the stuff used by pub export functions. It would be nice to also export free-standing enums and structs.

youknowone:
resolve #605


src/bindgen/parser.rs
```diff
@@ -484,6 +489,13 @@ impl Parse {
         self.functions.extend_from_slice(&other.functions);
         self.source_files.extend_from_slice(&other.source_files);
         self.package_version.clone_from(&other.package_version);
+        self.public_types.extend_from_slice(&other.public_types);
+    }
+
+    fn record_public_type(&mut self, record: bool, vis: &syn::Visibility, path: &Path) {
+        if record && matches!(vis, syn::Visibility::Public(_)) {
```
youknowone, line 496: I feed awkward that `record_public_type` and its callers have a parameter like `record: bool`, but didn't find good way not to do yet.

